PR #2, opened 2021-05-11
Add metrics and logs integration testing

maxday:
This PR bootstraps the integration tests setup for metrics and logs with the followings scenarios : 

- enhanced metrics ON
- enhanced metrics OFF
- timeout
- logs


integration_tests/serverless.yml
```diff
@@ -0,0 +1,84 @@
...
+    package:
+      artifact: ../.layers/datadog_extension.zip
+
+functions:
+  enhancedMetricTest:
+    name: integration-tests-lambda-extension-${sls:stage}-enhancedMetricTest
+    runtime: nodejs12.x
+    handler: src/metric.enhancedMetricTest
+    layers:
+      - arn:aws:lambda:${self:provider.region}:601427279990:layer:Datadog-Node12-x:${env:LAYER_VERSION}
```
nhinsch, line 46: I think we should pull the Node layer from the prod account (464622532012). That way we are testing the development version of the extension with the production Lambda Library.

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Build the layer
+        run: ./scripts/build_binary_and_layer.sh
+
+
+      - name: Run tests
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          # Todo : retrieve this layer version automatically from AWS
+          LAYER_VERSION: 6
```
nhinsch, line 58: I would rename this to `NODE_LAYER_VERSION` to make it more clear which layer we are talking about here.

integration_tests/recorder-extension/src/agent_payload.proto
```diff
@@ -0,0 +1,185 @@
+syntax = "proto3";
+
+// file copied from https://github.com/DataDog/agent-payload/blob/master/proto/metrics/agent_payload.proto
+
```
nhinsch, line 4: I think I may have left a comment to this effect before but I still think we should remove the code in this file that is not used by our tests. I know that you copied this file from elsewhere, but to me it still seems unnecessary to have some unrelated, unused Kubernetes-related code in our integration tests. I think it would be better to just copy the code that we use in these tests.

integration_tests/src/metric.js
```diff
@@ -0,0 +1,31 @@
...
+  invocationCount += 1;
+  return {
+    statusCode: 200,
+    body: 'ok'
+  };
+}
+
+async function myTimeoutHandler(event, context) {
+  sendDistributionMetric("serverless.lambda-extension.integration-test.count", invocationCount);
+  await new Promise(r => setTimeout(r, 30*1000)); //30 sec to be sure
```
nhinsch, line 21: Nit: For testing timeouts, I prefer to create an infinite loop. That way, the function  will definitely time out no matter how I adjust the Lambda timeout setting.

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Build the layer
+        run: ./scripts/build_binary_and_layer.sh
+
+
+      - name: Run tests
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          # Todo : retrieve this layer version automatically from AWS
```
nhinsch, line 57: Are you planning to incorporate getting the latest layer version into this PR?

integration_tests/serverless.yml
maxday: ✅

.github/workflows/integration_tests.yml
maxday: ✅  as we'll add more functions in different language soon

integration_tests/recorder-extension/src/agent_payload.proto
maxday: ✅

integration_tests/src/metric.js
maxday: I've adjust it with the max AWS allowed timeout value ✅

.github/workflows/integration_tests.yml
maxday: I don't think so, it's already pretty big I'd rather to do it in a next PR

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,59 @@
+name: build
+
+on: push
```
nhinsch, line 3: Weren't we going to run this hourly? Where is that set up?

nhinsch: Also, if the tests fail, will we be notified somehow?

nhinsch (approved):
This looks great! 💯 

I left a few last-minute questions about the GitHub action.

PR #3, opened 2021-05-24
Fix release scripts

nhinsch:
Fix a few problems with the release scripts:
- There was an extra, unnecessary `serverless` directory around the `scripts` directory
- Fix the file paths in the `build_docker_image.sh` script
- Create the correct tag in the datadog-lambda-extension GitHub when publishing to prod (`vX` not `lambda-extension-X`)
- Add instructions at the end of the production release for manually creating a tag in the agent repository and adding release notes
- Add a `.gitignore` file
- Fix other miscellaneous typos

When this is merged, I will also update our internal docs.


maxday:
Could you add a step to remove the any outdated previous builds to be extra cautious ? (extension and .layer folders)

maxday (approved):

nhinsch:
> Could you add a step to remove the any outdated previous builds to be extra cautious ? (extension and .layer folders)

I decided not to add this. The "build layers" script already deletes and removes the old versions of the layer. Also, if you try to publish an old version of the layer, the signing step will fail because the same file cannot be signed twice. It would actually be pretty difficult to publish an outdated build.

PR #11, opened 2021-06-21
Add trace support to integration testing

maxday:
Add integration tests for traces (nodejs + python)


integration_tests/serverless.yml
```diff
@@ -125,7 +125,7 @@ functions:
     runtime: python3.8
     handler: datadog_lambda.handler.handler
     layers:
-      - arn:aws:lambda:${self:provider.region}:464622532012:layer:Datadog-Python38:${env:PYTHON_LAYER_VERSION}
```
DarcyRaynerDD, line 128: Is there a reason we swapped to the staging layers?

DarcyRaynerDD (approved):
LGTM, left a small question

integration_tests/serverless.yml
maxday: I could not find a way to dynamically get the latest versions of prod layers without adding prod credentials as secrets

PR #14, opened 2021-07-28
Remove compress option, inject version number, and other miscellaney

nhinsch:
Build script:
- **Remove COMPRESS option** - We no longer want to compress the extension binary because we found it causes a significant performance penalty on cold starts. The old option also removed the symbol tables from the binary with the `-ldflags="-s -w"` option. We will now do this for all builds (production and sandbox) for consistency.
- **Inject the current extension version number into the binary** - This is achieved using another `ldflags` option. I will open another PR in the datadog-agent repository that adds this version number to all telemetry from the Lambda extension.

Integration tests:
- Mask extension version numbers in the snapshots so that they will be consistent from execution to execution. The new snapshots with the version numbers are not present in this pull request, as the corresponding datadog-agent pull request has not been merged yet.
- Add a `BUILD_EXTENSION` option that lets us build a new version of the extension as part of running the integration tests.
- Add usage instructions as a comment.
- Check for AWS credentials before running the integration tests.
- A few other minor changes, such as clarifying some comments.

### DIFF
--- a/scripts/build_binary_and_layer.sh
+++ b/scripts/build_binary_and_layer.sh
@@ -8,6 +8,14 @@
 set -e
 
 
+# Ensure the target extension version is defined
+if [ -z "$VERSION" ]; then
+    echo "Extension version not specified"
+    echo ""
+    echo "EXITING SCRIPT."
+    exit 1
+fi
+
 if [ -z "$CI" ]; then
     SERVERLESS_CMD_PATH="../datadog-agent/cmd/serverless"
 else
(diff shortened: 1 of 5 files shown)


scripts/build_binary_and_layer.sh
```diff
@@ -35,19 +43,16 @@ TARGET_DIR=$(pwd)/$EXTENSION_DIR
...
 cd $SERVERLESS_CMD_PATH
 
-LD_FLAGS=""
+# Add the current version number to the tags package before compilation
+LD_FLAGS="-X github.com/DataDog/datadog-agent/pkg/serverless/tags.currentExtensionVersion=${VERSION}"
 
-if [ "$COMPRESS" = true ]; then
-    LD_FLAGS="-s -w"
+if [ "$PRODUCTION" = true ]; then
+    LD_FLAGS+=" -s -w"
```
maxday, line 50: I think we should always build with those flags in order to have prod and non-pro very similar

maxday (approved):

maxday:
LGTM ! 💯 
left a small comment

PR #18, opened 2021-08-18
Dockerize build

maxday:
Since zstd is not compatible with cross compiling, we need to update the build to make it containerized


scripts/Dockerfile.build
```diff
@@ -0,0 +1,43 @@
+ARG VERSION
+FROM golang:1.16.5 as builder
+RUN mkdir -p /tmp/dd/datadog-agent
+
+# cache dependencies
+COPY ./scripts/.cache/datadog-agent /tmp/dd/datadog-agent
+WORKDIR /tmp/dd/datadog-agent
+RUN go mod graph \
+| grep -v winutil \
```
maxday, line 9: some package are failing during `go get` 
this is the only way to bypass the error and make the cache usable

nhinsch: Would you mind explaining more about why the packages are failing during `go get`?

maxday: sure, here is an example of an error :
`go: downloading github.com/StackExchange/wmi v1.2.1`
results in 
`github.com/StackExchange/wmi: no Go source files`
this will stop the loop (error code != 0) and the cache will be incomplete

scripts/build_binary_and_layer_dockerized.sh
```diff
@@ -0,0 +1,39 @@
...
+AGENT_PATH="../datadog-agent"
+EXTENSION_DIR=".layers"
+TARGET_DIR=$(pwd)/../$EXTENSION_DIR
+
+# Move into the root directory, so this script can be called from any directory
+SCRIPTS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+cd $SCRIPTS_DIR/..
+
+# First prepare a folder with only *mod and *sum files to enable Docker caching capabilities
+mkdir -p ./scripts/.src ./scripts/.cache
```
DarcyRaynerDD, line 26: This might touch the go.mod and go.sum files for every build. I'm not sure, but that might force the Docker file to rebuild all it's layers from the point you important the file. I'm not sure if docker uses last touched time on files, or whether it does a diff.

scripts/Dockerfile.build
```diff
@@ -0,0 +1,43 @@
...
+# zip the extension
+FROM ubuntu:latest as compresser
+RUN apt-get update
+RUN apt-get install -y zip
+RUN mkdir /extensions
+WORKDIR /extensions
+COPY --from=builder /tmp/dd/datadog-agent/cmd/serverless/datadog-agent /extensions/datadog-agent
+RUN  zip -r datadog_extension.zip /extensions
+
+# keep the smallest possible docker image
```
DarcyRaynerDD, line 40: Is there much point in doing this? Don't we just immediately copy the zip from the file?

DarcyRaynerDD (approved):
Left some questions

scripts/Dockerfile.build
nhinsch: I think the goal here was to not include Go/Linux/the ZIP utility in the final image we produce. It might be worth clarifying this comment slightly.

nhinsch (approved):
Looks good to me!

Is there a reason we want to keep the non-Dockerized version of the script as well? Or should this just fully replace the old script?

maxday:
> Looks good to me!
> 
> Is there a reason we want to keep the non-Dockerized version of the script as well? Or should this just fully replace the old script?

yes for integration tests which build the extension on github runner (linux based)

scripts/Dockerfile.build
maxday: this allow us to create a super tiny container based on this image to run `docker create + docker cp` on it here : https://github.com/DataDog/datadog-lambda-extension/pull/18/files#diff-e3684a939c0316200e1a5d9554fd2700a4be757efd37109326b92183234115c4R38

PR #22, opened 2021-09-03
Fix and speed up publish_sandbox

nhinsch:
- Move into the correct directory to use the `publish_sandbox` script
- Do not sign the layer when publishing to the sandbox AWS account. Signing the layer is time-consuming, requires an additional MFA token, and it not needed when testing in the sandbox account.

### DIFF
--- a/scripts/publish_sandbox.sh
+++ b/scripts/publish_sandbox.sh
@@ -8,10 +8,9 @@
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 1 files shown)


maxday (approved):

nhinsch:
Merging despite integration test failures as these failures are expected given the current state of the main agent branch, and that this PR has no effect on the tests.

PR #34, opened 2021-10-28
[WIP] Add go integration tests

jcstorms1:
This PR adds 3 new integration tests for the Go runtime in preparation for upcoming changes to the Lambda Extension.

- a metric test with the datadog lambda library
- a metric test without the datadog lambda library
- a timeout test
- a log test
- a trace test


integration_tests/run.sh
```diff
@@ -36,47 +36,51 @@ cd "./integration_tests"
...
 
 function remove_stack() {
     echo "Removing stack for stage : ${stage}"
-    serverless remove --stage ${stage} 
+    serverless remove --stage ${stage}
 }
 
 # always remove the stacks before exiting, no matter what
-trap remove_stack EXIT
+# trap remove_stack EXIT
```
nhinsch, line 74: I think we still want this line uncommented

integration_tests/src/go-tests/without-ddlambda/main.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/aws/aws-lambda-go/lambda"
+)
+
+type testResponse struct {
+	StatusCode int    `json:"statusCode"`
+	Body       string `json:"body"`
+}
+
+func PingHandler(ctx context.Context, ev events.APIGatewayProxyRequest) (testResponse, error) {
```
nhinsch, line 15: Let's add a log line here to make sure logging works without the ddlambda library

nhinsch: Actually, I think you could remove the "logs test" and move the log lines into the with-dd-lambda and without-dd-lambda tests. That way we cover logs in both cases.

integration_tests/run.sh
jcstorms1: Ahh yeah forgot I did that. Thanks!

integration_tests/serverless.yml
```diff
@@ -107,6 +126,16 @@ functions:
...
 
+  timeout-go:
+    runtime: provided.al2
+    timeout: 3
+    handler: src/bin/timeout
+    layers:
+      - { Ref: RecorderExtensionLambdaLayer }
+      - { Ref: DatadogExtensionIntegrationTestLambdaLayer }
+    environment:
+      DD_LOG_LEVEL: DEBUG
```
nhinsch, line 137: Why do you want to set the log level to `DEBUG` for this test?

integration_tests/serverless.yml
```diff
@@ -86,6 +96,15 @@ functions:
       DD_ENHANCED_METRICS: false
       DD_LAMBDA_HANDLER: src/metricPython.metric
 
+  no-enhanced-metric-go:
```
nhinsch, line 99: This function is called "no-enhanced-metric-go" but the handler is testing without the Lambda Layer. I think what we are really testing here is without the Lambda layer, so let's name the function based on that.

jcstorms1: 👍

integration_tests/serverless.yml
jcstorms1: I just followed the other tests. Not sure that it matters.

integration_tests/run.sh
```diff
@@ -36,50 +36,58 @@ cd "./integration_tests"
...
 cd recorder-extension
-    GOOS=linux GOARCH=amd64 go build -o extensions/recorder-extension main.go
-    zip -rq ext.zip extensions/* -x ".*" -x "__MACOSX" -x "extensions/.*"
+GOOS=linux GOARCH=amd64 go build -o extensions/recorder-extension main.go
+zip -rq ext.zip extensions/* -x ".*" -x "__MACOSX" -x "extensions/.*"
 cd ..
 
+go_test_dirs=("with-ddlambda" "without-ddlambda" "timeout" "trace")
+
 # build Go Lambda function
```
nhinsch, line 45: Nit: function -> functions

integration_tests/snapshots/log-go
```diff
@@ -0,0 +1,10 @@
+[log] {"message":{"message":"START XXX"}}
```
nhinsch, line 1: What function is this snapshot (`log-go`) for? Is this from an older function that you deleted?

nhinsch (approved):
Looks good, great work! Just left a couple small notes.

integration_tests/serverless.yml
```diff
@@ -86,6 +100,21 @@ functions:
...
       DD_LAMBDA_HANDLER: src/metricPython.metric
 
+  without-ddlambda-go:
+    runtime: provided.al2
+    handler: src/bin/without-ddlambda
+    layers:
+      - { Ref: RecorderExtensionLambdaLayer }
+      - { Ref: DatadogExtensionIntegrationTestLambdaLayer }
+    environment:
+      DD_ENHANCED_METRICS: false
```
nhinsch, line 110: I would set this to true and regenerate the snapshots.

PR #38, opened 2021-11-09
Fix go tests to maintain order of the log output

jcstorms1:
Originally the handler contained logs and metrics leading to a single binary, but we did not sort the log output of the integration test. This led to constant changes in the integration test output. Trying to use sort when normalizing the logs resulted in unordered log lines. I went back to a separated metric and log binary which is also in line with the other tests.


integration_tests/serverless.yml
```diff
@@ -199,6 +199,22 @@ functions:
       DD_EXTRA_TAGS: tagC:valueC tagD:valueD
       DD_LAMBDA_HANDLER: src/logPython.log
 
+  log-go:
+    without-ddlambda-go:
```
nhinsch, line 203: This looks a bit off, what's going on here?

nhinsch: i.e. without-dd-lambda go nested under log-go, maybe I am missing something though

integration_tests/src/go-tests/with-ddlambda/main.go
```diff
@@ -2,7 +2,6 @@ package main
```
nhinsch, line 2: I think we want to test logs and metrics both with and without dd-lambda-go.

nhinsch:
Looks like the tests are failing in CI (could be related to this PR, might be a pre-existing issue). What is causing those failures and is it possible to fix them in this PR?

nhinsch (approved):
Per Slack message, I think this is now in a place where it's ready to be merged (we will then move the tests to the agent repo).

integration_tests/run.sh
```diff
@@ -113,7 +114,10 @@ for function_name in "${all_functions[@]}"; do
     fi
 done
 
+now=$(date +"%r")
 echo "Sleeping $LOGS_WAIT_SECONDS seconds to wait for logs to appear in CloudWatch..."
+echo "This should be done in 10 minutes from $now"
```
nhinsch, line 119: This message is a nice touch, I think it improves the user experience of running the tests.

PR #40, opened 2021-11-15
Allow builds from any directory

nhinsch:
We should be able to run `build_binary_and_layer_dockerized` from any directory. This will allow us to move the integration tests to the agent repository while leaving other scripts in this repository. The integration test script in the other repository will be able to use this build script.


DarcyRaynerDD (approved):
LGTM

PR #52, opened 2022-02-09
Add the option to use go -race 

jcstorms1:
Created a secondary docker file, `Docker.race.build`, that will be used when `RACE_DETECTION_ENABLED` is set as `true`. It can be enabled with integration tests from the agent repo or when running the binary and layer build script individually.


scripts/build_binary_and_layer_dockerized.sh
```diff
@@ -44,20 +44,25 @@ cd $AGENT_PATH/..
 tar --exclude=.git -czf $ROOT_DIR/scripts/.src/datadog-agent.tgz datadog-agent
 cd $ROOT_DIR
 
+# Make sure to use the right build file if we want race detection enabled
```
nhinsch, line 47: Nit: I don't think you need this comment (code speaks for itself)

nhinsch (approved):
Looks good 👍

PR #54, opened 2022-03-21
Include tiered compilation for Java in wrapper script

nhinsch:
- Rename the wrapper script from `boot.sh` to `datadog_wrapper.sh` (the wrapper script hasn't been used by customers yet so this is not a breaking change)
- By default, set the `_JAVA_OPTIONS` environment variable to `-XX:+TieredCompilation -XX:TieredStopAtLevel=1` in the wrapper script. This will [stop tiered compilation at level 1](https://aws.amazon.com/blogs/compute/increasing-performance-of-java-aws-lambda-functions-using-tiered-compilation/), which will cause the JVM to spend less time optimizing the customer code. As a result, cold starts will take less time, though some functions that are called an extremely high number of times might see a negative performance impact.
- Add an environment variable `DD_OPTIMIZE_COLD_START` that can be set to `false` to disable the tiered compilation optimization. I don't think we should document this but if any customer wants to disable the optimization they will be able to do.

**NOTE:** The Java runtime will log `Picked up _JAVA_OPTIONS: -XX:+TieredCompilation -XX:TieredStopAtLevel=1` on cold start when this optimization is applied.

### DIFF
--- a/scripts/Dockerfile.build
+++ b/scripts/Dockerfile.build
@@ -29,8 +29,8 @@ RUN apt-get install -y zip
 RUN mkdir /extensions
 WORKDIR /extensions
 COPY --from=builder /tmp/dd/datadog-agent/cmd/serverless/datadog-agent /extensions/datadog-agent
-COPY ./scripts/boot.sh /boot.sh
-RUN  zip -r datadog_extension.zip /extensions /boot.sh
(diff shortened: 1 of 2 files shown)


maxday:
Could you add before/after performance cold start duration numbers?

nhinsch:
We are going with an alternative solution that does not require adding a wrapper script.

PR #55, opened 2022-03-22
Added setting the agent version in building and publishing extension

thedavlee:


local_tests/build-docker-python.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 
+set -e
+
 # Save the current path
 CURRENT_PATH=$(pwd)
 
 # Build the extension
-ARCHITECTURE=amd64 VERSION=1 ./scripts/build_binary_and_layer_dockerized.sh
+ARCHITECTURE=amd64 VERSION=1 AGENT_VERSION=$AGENT_VERSION ./scripts/build_binary_and_layer_dockerized.sh
```
maxday, line 9: maybe we can set a dummy AGENT_VERSION here as we're dealing with local testing (and remove the set -e) ?

scripts/Dockerfile.build
```diff
@@ -17,10 +18,17 @@ WORKDIR /tmp/dd/datadog-agent/cmd/serverless
...
-  -tags serverless -o datadog-agent
+    --mount=type=cache,target=/root/.cache/go-build \ 
+    if [ -z "$AGENT_VERSION" ]; then \
+        go build -ldflags="-w -s \
+        -X github.com/DataDog/datadog-agent/pkg/serverless/tags.currentExtensionVersion=$EXTENSION_VERSION" \
+        -tags serverless -o datadog-agent; \
+    else \
+        go build -ldflags="-w -s \
+        -X github.com/DataDog/datadog-agent/pkg/serverless/tags.currentExtensionVersion=$EXTENSION_VERSION \
+        -X github.com/DataDog/datadog-agent/pkg/version.agentVersionDefault=$AGENT_VERSION" \
```
maxday, line 29: am I.bit afraid that, since this variable is not in serverless owned files, this could change and break our build, potentially silently?
what happen if the variable is renamed or deleted? does it fail explicitly or silently?

thedavlee: if the variable is renamed, our agent version override fails and the extension falls back on using the default agent version (which we are currently using) in the metrics intake API calls. Perhaps I can do a check that the variable exists first?

local_tests/build-docker-python.sh
thedavlee: if agent version isn't specified in the when running `build-docker-python.sh`, `build_binary_and_layer_dockerized.sh` will not override the default agent version and the default will be used. How come we want to remove set -e?

scripts/build_binary_and_layer_dockerized.sh
```diff
@@ -17,6 +17,15 @@ if [ -z "$VERSION" ]; then
     exit 1
 fi
 
+if [ -z "$AGENT_VERSION" ]; then
+    echo "No agent version set, agentVersionDefault will be used instead (for the metrics log intake endpoint)."
+    read -p "This should only be done when building for staging or testing. Continue (y/n)?" CONT
```
nhinsch, line 22: Instead of printing `"This should only be done when building for staging or testing. Continue (y/n)?"` Can we make it so that if you are publishing for prod, the agent version is required, but if you are publishing for staging, the default version is automatically used? That way, we can replace a prompt with hardcoded logic.

nhinsch: I think we have separate files for releasing to prod and sandbox, maybe the logic can go in those files

scripts/build_binary_and_layer_dockerized.sh
```diff
@@ -17,6 +17,15 @@ if [ -z "$VERSION" ]; then
     exit 1
 fi
 
+if [ -z "$AGENT_VERSION" ]; then
+    echo "No agent version set, agentVersionDefault will be used instead (for the metrics log intake endpoint)."
```
nhinsch, line 21: What is the "metrics log intake endpoint"? Also I don't think you need to necessarily explain this in a log line.

thedavlee: typo, just the metrics intake endpoint. `https://x-y-z-app.agent.datadoghq.com/api/beta/sketches?api_key=***`. Removed this whole warning bit

scripts/build_binary_and_layer_dockerized.sh
thedavlee: Removed this warning. The only script where the agent version is required is `publish_prod.sh`. For `publish_sandbox.sh` as well as the local testing `build_docker_python.sh`, an agent version can still be specified otherwise the default is used.

scripts/Dockerfile.build
maxday: is there a way of doing this with `go` cli?

thedavlee: I'm investigating with `go tool nm`: https://pkg.go.dev/cmd/nm

thedavlee: added a check to fail the script if the variable doesnt exist (or is renamed)

maxday (approved):

PR #65, opened 2022-06-01
[serverless] Adding env variables to support .net and java instruments

nine5two7:
What does this PR do?
Setting env variables in the dd_wrapper file to support .net and java instruments.
This is the step  1.
Step 2 is to update the corresponding datadog-ci and serverless plugin.

Motivation:
https://datadoghq.atlassian.net/browse/SLS-2190

Describe how to test/QA your changes
Manual e2e test

### DIFF
--- a/scripts/Dockerfile.build
+++ b/scripts/Dockerfile.build
@@ -41,7 +41,7 @@ RUN mkdir /extensions
 WORKDIR /extensions
 COPY --from=builder /tmp/dd/datadog-agent/cmd/serverless/datadog-agent /extensions/datadog-agent
 
-ARG DATADOG_WRAPPER=datadog_wrapper.sh
+ARG DATADOG_WRAPPER=datadog_wrapper
 COPY ./scripts/$DATADOG_WRAPPER /$DATADOG_WRAPPER
 RUN chmod +x /$DATADOG_WRAPPER
 RUN  zip -r datadog_extension.zip /extensions /$DATADOG_WRAPPER
--- a/scripts/datadog_wrapper
+++ b/scripts/datadog_wrapper
@@ -20,11 +20,18 @@ then
   fi
 fi
 
-# if it is .Net
-echo "The runtime is $AWS_EXECUTION_ENV"
-if [[ "$AWS_EXECUTION_ENV" == *"dotnet"* ]];
+if [ "$DD_LOG_LEVEL" == "debug" ]
 then
-  echo "Configuring for the .NET runtime!"
+  echo "The runtime is $AWS_EXECUTION_ENV"
+fi
+
+# if it is .Net
+if [[ "$AWS_EXECUTION_ENV" == *"dotnet"* ]]
+then
+  if [ "$DD_LOG_LEVEL" == "debug" ]
+  then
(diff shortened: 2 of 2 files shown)


scripts/datadog_wrapper.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
tianchu, line 1: Can we remove the `.sh` extension from the wrapper? the examples from AWS are all without the .sh extension, i think one benefit is that we could potentially change it to a python or non-sh script down the road without asking users to update their configurations.

agocs: I agree it Should Be Fine to remove .sh as long as we have the shebang at the beginning of the file

scripts/datadog_wrapper.sh
```diff
@@ -0,0 +1,48 @@
...
+  export AWS_LAMBDA_RUNTIME_API="127.0.0.1:9000"
+
+  if [ "$DD_LOG_LEVEL" == "debug" ]
+  then
+    echo "[bootstrap] rerouting AWS_LAMBDA_RUNTIME_API to $AWS_LAMBDA_RUNTIME_API"
+  fi
+fi
+
+# if it is .Net
+echo "The runtime is $AWS_EXECUTION_ENV"
```
maxday, line 24: I think this should be protected by a if DD_LOG_LEVEL = debug ?

scripts/datadog_wrapper.sh
```diff
@@ -0,0 +1,48 @@
...
+if [[ "$AWS_EXECUTION_ENV" == *"dotnet"* ]];
+then
+  echo "Configuring for the .NET runtime!"
```
maxday, line 27: same here

scripts/datadog_wrapper.sh
```diff
@@ -0,0 +1,48 @@
...
+  export CORECLR_ENABLE_PROFILING="1"
+  export CORECLR_PROFILER="{846F5F1C-F9AE-4B07-969E-05C26BC060D8}"
+  export CORECLR_PROFILER_PATH="/opt/datadog/Datadog.Trace.ClrProfiler.Native.so"
+  export DD_DOTNET_TRACER_HOME="/opt/datadog"
+fi
+
+# if it is java
+if [[ "$AWS_EXECUTION_ENV" == *"java"* ]];
+then
+  echo "Configuring for the Java runtime!"
```
maxday, line 37: same here

maxday (commented):
Looks good! I left some minor comments

scripts/datadog_wrapper.sh
nine5two7: I will create a new revision for it.

scripts/datadog_wrapper.sh
nine5two7: Will fix it in the next revision

scripts/datadog_wrapper.sh
nine5two7: Will fix it in the next revision

maxday (approved):

PR #66, opened 2022-06-20
Fix extension size

maxday:
In https://github.com/DataDog/datadog-lambda-extension/pull/55
the `-s` flag was removed

This flag was useful to reduce the binary size : 
```
-s : Omit the symbol table and debug information.
```
 (source: https://pkg.go.dev/cmd/link)

We cannot put it back since we need the symbol table to inject the agent-version (see screenshot)
Screen Shot 2022-06-20 at 2 07 40 PM

The proposed solution is to use a tool AFTER the build `strip` to get rid of this symbol table after we've used it to inject the information : https://sourceware.org/binutils/docs/binutils/strip.html


nine5two7 (approved):
LGTM

IvanTopolcic (approved):

PR #69, opened 2022-06-28
To find dotNet profilers from all possible directories

nine5two7:
What does this PR do?
To search .NET profiler from all possible directories.
The new dd-extension will support .NET running on arm64, while the profiler file may reside in a new different directory other than before.
This PR is to handle such scenario.

Motivation:
https://datadoghq.atlassian.net/browse/SLS-2190

Describe how to test/QA your changes
Manual e2e test

### DIFF
--- a/scripts/datadog_wrapper
+++ b/scripts/datadog_wrapper
@@ -32,9 +32,30 @@ then
   then
     echo "Configuring for the .NET runtime!"
   fi
+  # Handle the CORECLR_PROFILER_PATH
+  DEFAULT_PATH=/opt/datadog/Datadog.Trace.ClrProfiler.Native.so
+  ARM64_PATH=/opt/datadog/linux-arm64/Datadog.Trace.ClrProfiler.Native.so
+  X64_PATH=/opt/datadog/linux-x64/Datadog.Trace.ClrProfiler.Native.so
+  MUSL_X64_PATH=/opt/datadog/linux-musl-x64/Datadog.Trace.ClrProfiler.Native.so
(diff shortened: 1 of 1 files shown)


scripts/datadog_wrapper
```diff
@@ -32,9 +32,30 @@ then
   then
     echo "Configuring for the .NET runtime!"
   fi
+  # Handle the CORECLR_PROFILER_PATH
+  DEFAULT_PATH=/opt/datadog/Datadog.Trace.ClrProfiler.Native.so
+  ARM64_PATH=/opt/datadog/linux-arm64/Datadog.Trace.ClrProfiler.Native.so
+  X64_PATH=/opt/datadog/linux-x64/Datadog.Trace.ClrProfiler.Native.so
+  MUSL_X64_PATH=/opt/datadog/linux-musl-x64/Datadog.Trace.ClrProfiler.Native.so
```
maxday, line 39: I don't think there is a musl compliant runtime for .NET so I think it's safe to get rid of this one?

scripts/datadog_wrapper
```diff
@@ -32,9 +32,30 @@ then
...
+  PROFILER_PATHS=("${DEFAULT_PATH}" "${ARM64_PATH}" "${X64_PATH}" "${MUSL_X64_PATH}")
```
maxday, line 40: same here

scripts/datadog_wrapper
nine5two7: Removed

scripts/datadog_wrapper
nine5two7: Removed

maxday (approved):

PR #71, opened 2022-07-07
aj/conditionally pass compat arch

astuyve:
- hack: compatible-architectures throws an error if the region doesn't support ARM, so this is happening
- hack: wrap in brackets
- hack: needed a space

### DIFF
--- a/scripts/publish_layers.sh
+++ b/scripts/publish_layers.sh
@@ -101,17 +101,17 @@ publish_layer() {
     else
       architecture="--compatible-architectures x86_64"
     fi
-    if [ "$region" == "ap-south-1" || \
-         "$region" == "ap-southeast-1" || \
-         "$region" == "ap-southeast-2" || \
-         "$region" == "ap-northeast-1" || \
-         "$region" == "eu-central-1" || \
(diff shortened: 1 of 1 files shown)


vvo:
Have you found a good resource page as for the list of ARM compatible regions?

astuyve:
Hey @vvo! Irritatingly no, but this change was based off of a [blog post](https://aws.amazon.com/about-aws/whats-new/2021/09/better-price-performance-aws-lambda-functions-aws-graviton2-processor/#:~:text=AWS%20Lambda%20functions%20powered%20by%20AWS%20Graviton2%20processors%20are%20available,and%20US%20West%20(Oregon)) announcing GA of Graviton. We know it worked because passing `--compatible-architectures` to `publish-layer-version` will raise an error in regions that don't support Graviton (hence this change).

We used it to release a new version of the extension yesterday (even though this change is not merged). I'm going to test an x86 -> arm lambda migration today with the latest version of the Datadog extension and get back to you on [this ticket](https://github.com/DataDog/datadog-cdk-constructs/issues/110).

Thanks again!

PR #73, opened 2022-07-20
Build docker image using arm64 when on an M1 chip.

purple4reina:
For those using the new M1 macbooks, the extension and layer must be built with `arm64` architecture instead of `amd64`. We can determine this dynamically using the `uname -m` command which will return `arm64` if on an M1.

Note I have not tested this on a non M1 computer. I would like to ask someone to do this before I merge.

### DIFF
--- a/local_tests/build-docker-python.sh
+++ b/local_tests/build-docker-python.sh
@@ -2,26 +2,35 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


maxday (approved):
LGTM!

Issue #83, opened 2022-10-13
Wrapper overrides JAVA_TOOL_OPTIONS


dangothemango:
https://github.com/DataDog/datadog-lambda-extension/blob/f74c203207809bdd7d66fa591ff03227f87d05cb/scripts/datadog_wrapper#L73

On startup the datadog wrapper overrides any JAVA_OPTS the user may have set. This causes those configurations to be ignored despite that environment variable being set on the lambda. Recommend appending the required config instead of overriding.

`export JAVA_TOOL_OPTIONS="-javaagent:$DD_Agent_Jar -XX:+TieredCompilation -XX:TieredStopAtLevel=1 $JAVA_TOOL_OPTIONS"`

However this could cause issues if an option is set twice so some logic to merge to two values may be required

dangothemango:
Further research indicates that Java will take the second option if duplicates are specified. Since I assume the arguments in this script are required a solution could be as simple as:

```
export JAVA_TOOL_OPTIONS="$JAVA_TOOL_OPTIONS -javaagent:$DD_Agent_Jar -XX:+TieredCompilation -XX:TieredStopAtLevel=1"
```

This way if we set the arguments to something incompatible it is automatically fixed and we can specify our arguments as needed provided they dont conflict with the necessary ones.

Still may be worth logging a warning saying some values are being overwritten

PR #86, opened 2022-10-20
feat: Append to JAVA_TOOL_OPTIONS instead of override

astuyve:
Closes https://github.com/DataDog/datadog-lambda-extension/issues/83


nine5two7 (approved):
LGTM

PR #91, opened 2022-11-08
Release the extension on sandbox + introducing build-tools!

maxday:
- release extension on sandbox account via a dispatch workflow
- simplify bash scripts via a new build-tools binary

Screen Shot 2022-11-09 at 6 33 20 PM
Screen Shot 2022-11-09 at 6 33 34 PM


build-tools/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM amd64/ubuntu:bionic as builder
```
DarcyRaynerDD, line 1: Why not the official rust image: https://hub.docker.com/_/rust ?

build-tools/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+COPY --from=builder /usr/src/app/target/release/build_tools /bin
+RUN strip /bin/build_tools
+
+# strip the binary and zip it
+FROM amd64/ubuntu:bionic as compresser
+RUN apt-get update
+RUN apt-get install -y zip
+RUN mkdir -p /bin
+WORKDIR /bin
+COPY --from=stripper /bin/build_tools /bin/build_tools
```
DarcyRaynerDD, line 32: Why do strip our build tools? Wouldn't debug information be helpful?

build-tools/src/commands/build_command.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+fn validate_args(args: &BuildOptions) {
+    validate_architecture_from_string(args.architecture.as_str());
+}
+
+fn validate_architecture_from_string(arch: &str) -> bool {
+    match arch {
+        "arm64" => true,
```
DarcyRaynerDD, line 36: I think you can express this as an enum with Struct Opts instead, and that way there will be a better command line error message.

DarcyRaynerDD (approved):

build-tools/Dockerfile
maxday: you're right, it's not stripped anymore

build-tools/Dockerfile
maxday: good catch, modified

build-tools/src/commands/build_command.rs
maxday: yeah I did not about that, this is way better now 

Screen Shot 2022-11-11 at 3 54 03 PM

PR #92, opened 2022-11-08
Local testing with Go

Julio-Guerra:
Adapt `build-docker-runtime.sh` to add a Go function without layer to download, along with the go source code in `local_test/golang/`
 
Usage example:
```console
ARCHITECTURE=amd64 RUNTIME=go ./local_tests/build-docker-runtime.sh
```

Note that AWS doesn't seem to provide an arm64 image, so you need to force the architecture to `amd64` on our macos m1 machines for it to work. Otherwise you end up executing the arm64 builds on the amd64 aws lamdba image :-/


maxday (approved):
Thanks! LGTM

PR #94, opened 2022-11-09
Add java support for local testing

maxday:
### Usage:

From root dir :

`RUNTIME=java ./local_tests/build-docker-runtime.sh`

`RUNTIME=java ./local_tests/invoke.sh 0`

see results in `./local_tests/logs.txt`


Julio-Guerra:
Suggestion: the build steps would be better in the Dockerfile so it is reproducible in the specified image environment and doesn't need the user to install extra tools on his machine, such as gradle here. Using a multi-stage dockerfile.

maxday:
> Suggestion: the build steps would be better in the Dockerfile so it is reproducible in the specified image environment and doesn't need the user to install extra tools on his machine, such as gradle here. Using a multi-stage dockerfile.

Good idea, done ✅

Julio-Guerra (approved):

PR #95, opened 2022-11-10
Automate two manual release steps.

purple4reina:
The publish prod release script has a couple of manual steps at the end. I think we can make these easier on ourselves by automating part of the processes. Additionally, because there was yet one more command run _after_ the manual steps, it was _very_ easy to not notice the manual steps in the first place.


purple4reina:
Closing because these changes cannot be performed by gh actions

PR #114, opened 2023-02-10
Build binaries with otlp enabled.

purple4reina:
Enabling OTLP endpoints in the lambda extension will require us to include the go build tag `"otlp"`. This pull request adds this tag when building `cmd/serverless` but not when building `cmd/serverless-init`.


purple4reina:
Related pull request to enable otlp in serverless: https://github.com/DataDog/datadog-agent/pull/15321

astuyve (approved):

PR #121, opened 2023-02-27
Add a check_layer_consistency_command

maxday:
Before deploying a new layer version to all regions, we need to make sure that the version is the same everywhere.
This will be useful when a new region will pop for instance

This PRs also 

- moves the `build_layer_name` as it's now used by two different commands
- better uses the OCI layering mechanism to speed up the build process

Example when layer version does not match:
Screenshot 2023-02-28 at 3 18 12 PM


clifordshelton:
Closing stale PRs

PR #122, opened 2023-02-28
Tools for visualizing the size of the extension.

purple4reina:
This pull request updates the `scripts/visualize_size.sh` to include several more helpful tools. See the help menu included in the script for more details on each of these.

Also added is a `BUILD_TAGS` variable which allows for configuration of the go build tags used when building the extension. It defaults to `serverless`, so by default the build script will act just as it did before this change.


purple4reina:
Note that this PR adds a `BUILD_TAGS` env var which defaults to `serverless`. We can use this env var going forward to switch between otlp and not-otlp.

maxday (approved):

PR #130, opened 2023-03-06
Revert "Revert "Revert "Revert "Build binaries with otlp enabled.""""

astuyve:
lotta reverts here


purple4reina:
Sadly, I am stopping the loop and suggest we use this PR instead: https://github.com/DataDog/datadog-lambda-extension/pull/134

PR #136, opened 2023-03-17
Move from github-actions to gitlab to publish layers

maxday:
We're now using gitlab to build and publish layer
- remove binary from git
- remove aem encryption
- remove MFA as it's running on internal runner


.gitlab-ci.yml
```diff
@@ -0,0 +1,134 @@
...
+  GIT_DEPTH: 1
+  REGION_TO_DEPLOY:
+    description: "use sa-east-1 for dev, us-east-1 for RC, all for all regions"
+    value: sa-east-1
+  AGENT_BRANCH:
+    description: "datadog-agent branch you want to release"
+    value: main
+  LAYER_SUFFIX:
+    description: "Suffix to be appended to the layer name (default empty)"
+    value: ""
```
purple4reina, line 11: It would be nice to add build tags here as well, so we don't have to keep reverting PRs when we want to change the build tags.

purple4reina: Good idea. It's also less important now that we're solidly in the otlp build tag world as opposed to previously where we were going back and forth regularly.

PR #139, opened 2023-04-07
Disable function inlining, to reduce executable size

DarcyRaynerDD:
Disables function inlining, which in practice sees about a 3mb reduction in build size, or a ~11% reduction.


astuyve (approved):
Code LGTM, not sure about whitespace but then/fi is easy enough to read IMHO.

duncanista:
Closing due to stale

PR #147, opened 2023-07-06
Fix ARM build

maxday:
https://github.com/DataDog/datadog-lambda-extension/pull/145 introduced an issue where the SHA1 specified is no more compatible with multi-arch docker

This PRs use the AWS Lambda provided image + install go lang to avoid any issue with GLIBC + respect the multi arch

### DIFF
--- a/scripts/Dockerfile.build
+++ b/scripts/Dockerfile.build
@@ -1,13 +1,23 @@
 # syntax = docker/dockerfile:experimental
 
-# this fixes the version of the golang image to avoid breaking changes in glibc
-# sha1 comes from https://github.com/docker-library/repo-info/blob/b939681d85e36bae10eabe53b913f1d05e972f1f/repos/golang/remote/1.19.md
-FROM golang@sha256:2dde7ccba42da98d45e0f38f23f1c0d1474a779e4089faf665f9af0dd3f844db as builder
+FROM public.ecr.aws/lambda/provided:al2 as builder
 ARG EXTENSION_VERSION
 ARG AGENT_VERSION
 ARG CMD_PATH
 ARG BUILD_TAGS
 RUN mkdir -p /tmp/dd/datadog-agent
+
+# install go 1.19
+RUN yum install -y wget tar gzip gcc
+RUN arch="$(uname -m)"; \
+    if [ "${arch}" = 'aarch64' ]; then \
+        arch='arm64'; \
+    fi; \
+    if [ "${arch}" = 'x86_64' ]; then \
+        arch='amd64'; \
+    fi; \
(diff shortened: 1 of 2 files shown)


scripts_v2/Dockerfile.build
```diff
@@ -1,13 +1,23 @@
...
 
+# install go 1.19
+RUN yum install -y wget tar gzip gcc
+RUN arch="$(uname -m)"; \
+    if [ "${arch}" = 'aarch64' ]; then \
+        arch='arm64'; \
+    fi; \
+    if [ "${arch}" = 'x86_64' ]; then \
+        arch='amd64'; \
+    fi; \
```
purple4reina, line 17: alternatively, you could pass arch in as a build arg.

scripts_v2/Dockerfile.build
```diff
@@ -1,13 +1,23 @@
...
+    wget -O go1.19.10.linux-${arch}.tar.gz https://go.dev/dl/go1.19.10.linux-${arch}.tar.gz; \
+    tar -C /usr/local -xzf go1.19.10.linux-${arch}.tar.gz
```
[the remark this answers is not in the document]
truthbk: IMHO your build process should leverage a builder image that is already available with all dependencies like `golang` already satisfied. That said, the overhead of this should be relatively small compared to the actual build time.

scripts/Dockerfile.build
```diff
@@ -1,14 +1,24 @@
...
+# install go 1.19
+RUN yum install -y wget tar gzip gcc
+RUN arch="$(uname -m)"; \
+    if [ "${arch}" = 'aarch64' ]; then \
+        arch='arm64'; \
+    fi; \
+    if [ "${arch}" = 'x86_64' ]; then \
+        arch='amd64'; \
+    fi; \
+    wget -O go1.19.10.linux-${arch}.tar.gz https://go.dev/dl/go1.19.10.linux-${arch}.tar.gz; \
```
truthbk, line 19: I find it easier to leverage gimme in many of the agent builder images: https://github.com/travis-ci/gimme it will abstract much of what you've got here. 

But this is fine. One of the positives of this approach is you can validate the `sha256sum` of the downloaded artifacts for a more reproducible image (which is nice).

truthbk (commented):
Nice! I hope you guys can get this in soon! 🙇

scripts_v2/Dockerfile.build
maxday: yeah definitely, I'll leave this like that for now to be sure not to break backward compatibility

scripts/Dockerfile.build
maxday: ah nice I did not know about gimme, I'll definitely have a look thanks

PR #150, opened 2023-07-12
update paths to .NET tracer libs

lucaspimentel:
This PR updates the paths used in the `datadog_wrapper` script to discover the .NET tracer's libraries.

Try these paths, in order:
- the shared loader
- the tracer stand-alone library
- an older path for backwards compat with very old .NET tracer layers

### DIFF
--- a/scripts/datadog_wrapper
+++ b/scripts/datadog_wrapper
@@ -33,15 +33,20 @@ then
     echo "Configuring for the .NET runtime!"
   fi
   # Handle the CORECLR_PROFILER_PATH:
-  # Load the tracer library directly (Datadog.Tracer.Native.so),
-  # which avoids loading the shared loader (Datadog.Trace.ClrProfiler.Native.so)
-  # since we don't support Continuous Profiling yet.
-  # Keep trying Datadog.Trace.ClrProfiler.Native.so for backwards compatibility.
-  PROFILER_PATHS=("/opt/datadog/linux-x64/Datadog.Trace.ClrProfiler.Native.so"
-                  "/opt/datadog/linux-arm64/Datadog.Trace.ClrProfiler.Native.so"
-                  "/opt/datadog/linux-x64/Datadog.Tracer.Native.so"
-                  "/opt/datadog/linux-arm64/Datadog.Tracer.Native.so"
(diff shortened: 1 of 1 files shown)


scripts/datadog_wrapper
```diff
@@ -32,11 +32,15 @@ then
...
-  DEFAULT_PATH=/opt/datadog/Datadog.Trace.ClrProfiler.Native.so
-  ARM64_PATH=/opt/datadog/linux-arm64/Datadog.Trace.ClrProfiler.Native.so
-  X64_PATH=/opt/datadog/linux-x64/Datadog.Trace.ClrProfiler.Native.so
-  PROFILER_PATHS=("${DEFAULT_PATH}" "${ARM64_PATH}" "${X64_PATH}")
+  # Handle the CORECLR_PROFILER_PATH:
+  # Load the tracer library directly (Datadog.Tracer.Native.so),
+  # which avoids loading the shared loader (Datadog.Trace.ClrProfiler.Native.so)
+  # since we don't support Continuous Profiling yet.
+  # Keep trying Datadog.Trace.ClrProfiler.Native.so for backwards compatibility.
+  PROFILER_PATHS=("/opt/datadog/linux-x64/Datadog.Tracer.Native.so"
```
joeyzhao2018, line 40: I have two small concerns.
- If we start to add extra valid path, old version of the extension may not work with the new version of the layer (if the new layers no longer copy the `Datadog.Tracer.Native.so` as the `Datadog.Trace.ClrProfiler.Native.so`
- I don't know how hard it would be to support the continuous profiling feature in lambda. Maybe we can support it in the near future?

lucaspimentel: Good catch! I considered backwards compatibility in one direction (newer extension with older dotnet layer), but not the opposite direction.

lucaspimentel: @joeyzhao2018, I restored `/opt/datadog/Datadog.Trace.ClrProfiler.Native.so`, re-ordered the paths, and added comments. This should now cover all the cases and be backwards and forwards compatible.

joeyzhao2018 (approved):
LGTM

PR #151, opened 2023-07-13
Switch serverless-init deployments to use GHCR instead of Dockerhub

IvanTopolcic:
It's easier to use GHCR as we don't need to manage credentials, and we're replicating to Dockerhub, ECR, and GCR anyways. This makes GHCR the source of truth, and provides clear separation as to which registry is responsible for what.

### DIFF
--- a/.github/workflows/release-serverless-init.yml
+++ b/.github/workflows/release-serverless-init.yml
@@ -51,22 +51,18 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
-      # - name: Build binaries
-      #   working-directory: ./scripts
-      #   run: ./build_serverless_init.sh
-      #   env:
-      #     AGENT_PATH: datadog-agent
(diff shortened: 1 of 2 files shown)


.github/workflows/release-serverless-init.yml
```diff
@@ -55,56 +56,57 @@ jobs:
...
 
-      - name: Login to DockerHub
+      - name: Login to GHCR
         uses: docker/login-action@v2
         with:
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_TOKEN }}
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
maxday, line 74: have you already set the secrets? (and in 1password?)

IvanTopolcic: it's a per-repo secret set automatically by github, we don't manage it

maxday (approved):

PR #154, opened 2023-07-19
[SLES-1145] Check if the Java trace agent jar is present

DylanLovesCoffee:
Adds a check for the DD trace agent .jar prior to disabling runtime metrics. More specific to those instrumenting their container image Lambda functions, since the `AWS_EXECUTION_ENV` environment variable is not immediately identifiable as Java.

We had a support case where this specific issue came up, causing runtime metrics to be enabled and thus timing out the DD Extension's/agent's metric forwarder.

### DIFF
--- a/scripts/datadog_wrapper
+++ b/scripts/datadog_wrapper
@@ -68,7 +68,8 @@ then
 fi
 
 # if it is java
-if [[ "$AWS_EXECUTION_ENV" == *"java"* ]]
+DD_Agent_Jar=/opt/java/lib/dd-java-agent.jar
+if [[ "$AWS_EXECUTION_ENV" == *"java"* ]] || [ -f "$DD_Agent_Jar" ]
 then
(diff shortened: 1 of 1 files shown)


scripts/datadog_wrapper
```diff
@@ -68,7 +68,8 @@ then
 fi
 
 # if it is java
-if [[ "$AWS_EXECUTION_ENV" == *"java"* ]]
+DD_Agent_Jar=/opt/java/lib/dd-java-agent.jar
+if [[ "$AWS_EXECUTION_ENV" == *"java"* ]] || [ -f "$DD_Agent_Jar" ]
```
maxday, line 72: should we test AND instead of OR?

DylanLovesCoffee: Yeah, let's do that 👍

maxday (approved):

PR #155, opened 2023-07-19
Bump Go version to 1.20

IvanTopolcic:


maxday:
Thanks! I think https://github.com/DataDog/datadog-lambda-extension/blob/main/scripts/Dockerfile.build and https://github.com/DataDog/datadog-lambda-extension/blob/main/scripts_v2/Dockerfile.build would need a bump as well

maxday (approved):

PR #159, opened 2023-08-08
Update extension overhead explanation

tianchu:
Mention DD_SERVERLESS_FLUSH_STRATEGY and aws.lambda.post_runtime_extensions_duration.


README.md
```diff
@@ -21,11 +21,11 @@ Follow the [configuration instructions](https://docs.datadoghq.com/serverless/co
...
 
-For Lambda functions deployed in a region that is far from the Datadog site, for example, a Lambda function deployed in eu-west-1 reporting data to the US1 Datadog site, can observe a higher duration overhead due to the network latency.
+For Lambda functions deployed in a region that is far from the Datadog site, for example, a Lambda function deployed in eu-west-1 reporting data to the US1 Datadog site, can observe a higher duration (cost) overhead due to the network latency. To reduce the overhead, configure the extension to flush data less often, such as every minute `DD_SERVERLESS_FLUSH_STRATEGY=periodically,60000`.
```
sumedham, line 28:
```suggestion
For Lambda functions deployed in a region that is far from the Datadog site, for example, a Lambda function deployed in eu-west-1 reporting data to the US1 Datadog site, can observe a higher duration (and therefore, cost) overhead due to the network latency. To reduce the overhead, configure the extension to flush data less often, such as every minute `DD_SERVERLESS_FLUSH_STRATEGY=periodically,60000`.
```

README.md
```diff
@@ -21,11 +21,11 @@ Follow the [configuration instructions](https://docs.datadoghq.com/serverless/co
...
+By default, the Extension flushes data back to Datadog at the end of each invocation (for example, cold starts always trigger flushing). This avoids delays of data arrival for sporadic invocations from low-traffic applications, cron jobs, and manual tests. Once the Extension detects a steady and frequent invocation pattern (more than once per minute), it batches data from multiple invocations and flushes periodically at the beginning of the invocation when it's due. This means that *the busier your function is, the lower the average duration overhead per invocation*. In other words, for low-traffic applications, the duration overhead would be noticeable while the associated cost overhead is typically negligible; for high-traffic applications, the duration overhead would be barely noticeable. To understand the cost overhead casued by the duration used by the Datadog extension to flush data, use the metric `aws.lambda.post_runtime_extensions_duration` or `aws.lambda.enhanced.post_runtime_duration`.
```
sumedham, line 26:
```suggestion
By default, the Extension flushes data back to Datadog at the end of each invocation (for example, cold starts always trigger flushing). This avoids delays of data arrival for sporadic invocations from low-traffic applications, cron jobs, and manual tests. Once the Extension detects a steady and frequent invocation pattern (more than once per minute), it batches data from multiple invocations and flushes periodically at the beginning of the invocation when it's due. This means that *the busier your function is, the lower the average duration overhead per invocation*. In other words, for low-traffic applications, the duration overhead would be noticeable while the associated cost overhead is typically negligible; for high-traffic applications, the duration overhead would be barely noticeable. To understand the compute overhead casued by the duration used by the Datadog extension to flush data, use the metric `aws.lambda.post_runtime_extensions_duration` or `aws.lambda.enhanced.post_runtime_duration`. 
```

sumedham: It's a little odd to say "cost overhead" and not point to a metric with $ units, lmk if this makes sense

sumedham (approved):
minor nits

README.md
tianchu: i agree, but compute overhead is also confusing to me. I would just say duration overhead then.

DarcyRaynerDD (approved):

PR #162, opened 2023-08-16
Support building extension images for alpine

thedavl:
Support building + releasing alpine lambda extension images

This builds both amd64 & arm64 alpine images. The images will be tagged `latest-alpine` and `{supplied_version}-alpine`

### DIFF
--- a/scripts/build_and_push_docker_image.sh
+++ b/scripts/build_and_push_docker_image.sh
@@ -2,9 +2,6 @@
 
 # Use with `VERSION=<DESIRED_VERSION> ./build_docker_image.sh`
 
-# To publish an Alpine image, set ALPINE=true:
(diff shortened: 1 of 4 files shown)


scripts/build_and_push_docker_image.sh
```diff
@@ -10,7 +10,6 @@
 set -e
 
 DOCKER_REPOSITORY_NAME="datadog/lambda-extension"
-DOCKERFILE_LOCATION="scripts/Dockerfile"
```
thedavl, line 13: unused var

IvanTopolcic (approved):

astuyve (approved):
This looks good, but are we going to update the v2 scripts at all?

thedavl:
> This looks good, but are we going to update the v2 scripts at all?

afaik the v2 scripts are only used for the one click extension releases, which only deploy layers (hence no alpine). so i think we're good

PR #179, opened 2023-11-10
Svls 2771 rust shell ext

alexgallotta:
- build rust lambda extension shell
- add rust lambda extension shell as call from the wrapper


alexgallotta:
waiting for review before adding the rust shell copy to all dockerfiles

PR #201, opened 2024-03-22
feat: Set DD_APPSEC_ENABLED to false by default to avoid errors where appsec detects remoteconfig is not available

astuyve:
Fixes https://github.com/DataDog/serverless-plugin-datadog/issues/477

For Java, when RemoteConfig is disabled but AppSec is enabled, it logs an error on startup. This setting should save a few CPU cycles while also suppressing the log.

### DIFF
--- a/scripts/datadog_wrapper
+++ b/scripts/datadog_wrapper
@@ -86,6 +86,7 @@ then
   export DD_JMXFETCH_ENABLED="false"
   export DD_RUNTIME_METRICS_ENABLED="false"
   export DD_REMOTE_CONFIG_ENABLED="false"
+  export DD_APPSEC_ENABLED="false"
 
   # Removes the -XX:-TieredCompilation flag from the java command passed 
   # through from the Lambda runtime. Allows the JVM to use the C1 compiler


scripts/datadog_wrapper
```diff
@@ -86,6 +86,7 @@ then
   export DD_JMXFETCH_ENABLED="false"
   export DD_RUNTIME_METRICS_ENABLED="false"
   export DD_REMOTE_CONFIG_ENABLED="false"
+  export DD_APPSEC_ENABLED="false"
```
duncanista, line 89: Will this be overridden if the user sets it to `true`? I'm not aware what's our support parity on the AppSec side...

duncanista (approved):

scripts/datadog_wrapper
astuyve: This only disables the tracer-side appsec. For serverless appsec, we use DD_SERVERLESS_APPSEC_ENABLED env var as the implementation is in the agent and not the tracer.

duncanista: Thanks for clarifying!

purple4reina (approved):

PR #206, opened 2024-04-12
add metrics support

maxday:


bottlecap/src/event_bus/bus.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+    pub fn run(&mut self) {
+        let (tx, rx) = mpsc::channel::<Event>();
+        self.tx = Some(tx);
+        let _ = thread::spawn(move || loop {
+            let received = rx.recv();
+            if let Ok(event) = received {
+                match event {
+                    Event::Metric(event) => {
+                        println!("Metric event: {:?}", event);
```
DarcyRaynerDD, line 25: Should we use the logging functionality from the "tracing" crate instead?

bottlecap/src/event_bus/bus.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use crate::events;
+use events::Event;
+
+pub struct EventBus {
+    tx: Option<Sender<events::Event>>,
+}
+
+impl EventBus {
+    pub fn new() -> EventBus {
```
DarcyRaynerDD, line 14: You could just derive Default for this struct instead of using new.

bottlecap/src/event_bus/bus.rs
```diff
@@ -0,0 +1,43 @@
...
+        EventBus { tx: None }
+    }
+    pub fn run(&mut self) {
+        let (tx, rx) = mpsc::channel::<Event>();
+        self.tx = Some(tx);
+        let _ = thread::spawn(move || loop {
```
DarcyRaynerDD, line 20: This drops the JoinHandle for the thread. We should probably keep that, so when we shut down we can guarantee this thread has finished processing.

DarcyRaynerDD: Another thought about the design. Perhaps it would be better if this run method was more like a "class method" or associated function, that does constructions of the EventBus in one step. That way an event bus object can always have a Sender, and you can remove the panic in get_sender_copy.

bottlecap/src/main.rs
```diff
@@ -84,6 +95,12 @@ fn register() -> Result<RegisterResponse> {
 }
 
 fn main() -> Result<()> {
+    tracing_subscriber::fmt()
+        .json()
+        .with_env_filter(EnvFilter::from_default_env())
```
DarcyRaynerDD, line 100: We should set the Env filter based on what we read from config. For instance, this could be set in the datadog.yaml

bottlecap/src/metrics/dogstatsd.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use crate::events::{self, Event, MetricEvent};
+
+pub struct DogStatsD<'a> {
+    host: &'a str,
+    port: u16,
+}
+
+impl<'a> DogStatsD<'a> {
+    pub fn new(host: &'a str, port: u16) -> Self {
```
DarcyRaynerDD, line 11: Similar comment about spawning the EventBus. Don't think this needs to be tied to a struct even. Do we even need this "host" field?

DarcyRaynerDD (approved):
Left some comments with some nits and thoughts

bottlecap/src/event_bus/bus.rs
maxday: yeah that's a good idea. I think I've addressed it in the last commit, let me know if this is what you had in mind

PR #207, opened 2024-04-15
Use the right logs env var

astuyve:
- **first draft metrics**
- **add dogstatsd + event_bus**
- **pr feedback**
- **pr feedback dogstats**
- **feat: serverless_logs_enabled, not logs_enabled**

Based on Max's PR. We need `dd_serverless_logs_enabled` in order for the forwarder to push logs and enhanced metrics to the backend. This config option was allowed, but incorrectly named so Bottlecap kept failing over into the extension.


maxday (approved):
LGTM!

PR #209, opened 2024-04-16
feat: Clear env vars before running

astuyve:
`DATADOG_ROOT` is set on my machine and conflicts with this test, throwing an unexpected field error with name `root`.

This fixes it.


duncanista (approved):
LGTM – we should add setup and teardown eventually.

astuyve:
Not sure how that would work with figment jails but we can certainly try.

PR #212, opened 2024-04-25
Aj/metrics

astuyve:
Note: experimental.

Rough outline for gauge and counter metrics. Many TODOs here, pushing to instead start work on lifecycle.

Todos:
[ ] Use or remove `generation` feature
[ ] Distribution metrics
[ ] Handle multi-value metrics. Likely this means flattening them in the server thread, or refactoring MetricEvents to handle multi-value metrics.
[ ] Asynchrony to handle scheduled flushes as well as synchronous flushes.

### DIFF
--- a/bottlecap/Cargo.toml
+++ b/bottlecap/Cargo.toml
@@ -8,12 +8,15 @@ edition = "2021"
 [dependencies]
 figment = { version = "0.10.15", features = ["yaml", "env"] }
 log = "0.4.21"
-tokio = "1.37.0"
 serde = { version = "1.0.197", features = ["derive"], default-features = false }
-ureq = { version = "2.9.6", features = ["json"], default-features = false }
-
+ureq = { version = "2.9.6", features = ["json", "tls"], default-features = false }
 tracing = {version = "0.1.40", default-features = false }
 tracing-subscriber = { version = "0.3.18", default-features = false, features = ["fmt", "env-filter", "json"] }
+ustr = "1.0.0"
+thiserror = "1.0.58"
+fnv = "1.0.7"
+hashbrown = "0.14.3"
+serde_json = "1.0.116"
 
 
 [dev-dependencies]
--- a/bottlecap/Cargo.toml
+++ b/bottlecap/Cargo.toml
@@ -21,6 +21,7 @@ serde_json = "1.0.116"
 
 [dev-dependencies]
 figment = { version = "0.10.15", features = ["yaml", "env", "test"] }
+proptest = "1.4.0"
 
 
 [profile.release]
--- a/bottlecap/src/metrics/aggregator.rs
+++ b/bottlecap/src/metrics/aggregator.rs
@@ -293,6 +293,21 @@ mod tests {
     }
 
     #[test]
+    fn clear() {
+        let mut aggregator = Aggregator::<2>::new().unwrap();
+
+        let metric1 = Metric::parse("test:1|c|k:v").expect("metric parse failed");
+        let metric2 = Metric::parse("foo:1|c|k:v").expect("metric parse failed");
+
+        assert!(aggregator.insert(&metric1).is_ok());
+        assert!(aggregator.insert(&metric2).is_ok());
+
+        assert_eq!(aggregator.len(), 2);
+        aggregator.clear();
+        assert_eq!(aggregator.len(), 0);
+    }
+
+    #[test]
     fn to_series() {
         let mut aggregator = Aggregator::<2>::new().unwrap();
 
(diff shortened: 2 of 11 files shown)


bottlecap/Cargo.toml
```diff
@@ -8,16 +8,20 @@ edition = "2021"
...
-ureq = { version = "2.9.6", features = ["json"], default-features = false }
-
+ureq = { version = "2.9.6", features = ["json", "tls"], default-features = false }
 tracing = {version = "0.1.40", default-features = false }
 tracing-subscriber = { version = "0.3.18", default-features = false, features = ["fmt", "env-filter", "json"] }
+ustr = "1.0.0"
+thiserror = "1.0.58"
+fnv = "1.0.7"
+hashbrown = "0.14.3"
+serde_json = "1.0.116"
```
blt, line 19: One suggestion would be to stick with the more explicit pattern of specifying features et al above. It's annoying but you won't be surprised when you do a version upgrade and new features/deps get flagged in.

astuyve: oh good point. Done.

bottlecap/src/metrics/dogstatsd.rs
```diff
@@ -1,15 +1,36 @@
...
+    serve_handle: std::thread::JoinHandle<()>,
+    aggregator: Arc<Mutex<Aggregator<1024>>>,
+    dd_api: datadog::DdApi,
 }
 
+pub struct DogStatsDConfig {
+    pub host: String,
+    pub port: u16,
+}
+const CONTEXTS: usize = 1024;
```
blt, line 20: Suggesting moving this to `constants.rs`

bottlecap/src/metrics/dogstatsd.rs
```diff
@@ -1,15 +1,36 @@
...
+        let serializer_aggr = Arc::clone(&aggr);
+        let serve_handle = DogStatsD::run_server(&config.host, config.port, event_bus, serializer_aggr);
+        let dd_api = datadog::DdApi::new();
+        DogStatsD { serve_handle, aggregator: aggr, dd_api }
+    }
+
+    fn run_server(host: &str, port: u16, event_bus: Sender<events::Event>, aggregator: Arc<Mutex<Aggregator<CONTEXTS>>>) -> std::thread::JoinHandle<()> {
         let addr = format!("{}:{}", host, port);
-        let join_handle = std::thread::spawn(move || {
+        std::thread::spawn(move || {
```
blt, line 33: FYI https://doc.rust-lang.org/std/thread/struct.Builder.html if you think that you might not be able to create a thread in your environment but need to handle that without panicing.

bottlecap/src/metrics/dogstatsd.rs
```diff
@@ -21,16 +42,45 @@ impl DogStatsD {
                     msg,
                     src
                 );
-                let dummy_tags = vec!["tagA:valueA".to_string(), "tagB:valueB".to_string()];
-                let metric_event = MetricEvent::new("metric_name".to_string(), 1.0, dummy_tags);
+                let parsed_metric = match Metric::parse(msg) {
+                    Ok(parsed_metric) => {
+                        log::info!("parsed metric: {:?}", parsed_metric);
```
blt, line 47: In practice you will probably want to move this down to `debug` at some point, although I see the utility now.

bottlecap/src/metrics/datadog.rs
```diff
@@ -0,0 +1,148 @@
...
+use serde::{Serialize, Serializer};
+use serde_json;
+use tracing::error;
+
+/// Interface for the DogStatsD metrics intake API.
+#[derive(Debug)]
+pub struct DdApi {
+    // runtime: tokio::runtime::Runtime,
+    api_key: String,
+}
```
blt, line 15: FWIW I think you can park a `ureq::Agent` in this struct and avoid needing an Arc etc.

bottlecap/src/metrics/aggregator.rs
```diff
@@ -0,0 +1,337 @@
...
+            tags: metric.tags,
+            kind: metric.kind,
+        }
+    }
+
+    #[inline]
+    fn aged_out(self, generation: u16) -> bool {
+        self.generation.abs_diff(generation) >= 2
+    }
+}
```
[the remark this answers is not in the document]
astuyve: should be merged

bottlecap/src/metrics/dogstatsd.rs
astuyve: We should be good here, but good to know!

duncanista (approved):

PR #216, opened 2024-05-02
aj/bottlecap/lifecycle

astuyve:
- **WIP- lifecycle**
- **feat: Parse dd serverless flush strategy**
- **feat: Add lifecycle scheduler, honor DD_SERVERLESS_FLUSH_STRATEGY, adds flush controller and invocation counter**

First, we move the event bus reader into the main loop so we can decide if we should block and wait for `PlatformRuntimeDone` (so that we can flush) or not.

In the future this logic will get more complex as we may not necessarily need to flush after the function exits, but perform some form of compaction or aggregation, freeing memory, etc.

This adds a lifecycle module composed of a flush controller and invocation counter. This should implement the same flushing strategy we have in the main extension.

This also adds a parser for the DD_SERVERLESS_FLUSH_STRATEGY env var

The logic is described in the FlushControl module.

### DIFF
--- a/bottlecap/src/config.rs
+++ b/bottlecap/src/config.rs
@@ -313,7 +313,10 @@ pub mod tests {
     fn test_parse_flush_strategy_invalid_periodic() {
         figment::Jail::expect_with(|jail| {
             jail.clear_env();
-            jail.set_env("DD_SERVERLESS_FLUSH_STRATEGY", "periodically,invalid_interval");
+            jail.set_env(
+                "DD_SERVERLESS_FLUSH_STRATEGY",
+                "periodically,invalid_interval",
+            );
             let config = get_config(Path::new("")).expect("should parse config");
             assert_eq!(
                 config,
--- a/bottlecap/src/config.rs
+++ b/bottlecap/src/config.rs
@@ -49,9 +49,13 @@ impl<'de> Deserialize<'de> for FlushStrategy {
                         // "60000"
                         match interval {
                             Some(interval) => {
-                                let interval =
-                                    interval.parse().map_err(serde::de::Error::custom)?;
-                                Ok(FlushStrategy::Periodically(PeriodicStrategy { interval }))
+                                if let Ok(parsed_interval) = interval.parse() {
+                                    return Ok(FlushStrategy::Periodically(PeriodicStrategy {
+                                        interval: parsed_interval,
+                                    }));
+                                }
+                                debug!("invalid flush interval: {}, using default", interval);
+                                Ok(FlushStrategy::Default)
                             }
                             None => {
                                 debug!("invalid flush strategy: {}, using default", value);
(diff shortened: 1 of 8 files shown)


bottlecap/src/lifecycle/flush_control.rs
```diff
@@ -0,0 +1,89 @@
...
+                self.last_flush = now;
+                true
+            },
+            FlushStrategy::Periodically(periodic) => self.should_periodic_flush(periodic.interval),
+            FlushStrategy::End => true,
+        }
+    }
+
+    fn should_periodic_flush(&mut self, interval: u64) -> bool {
+        let now = time::SystemTime::now().duration_since(time::UNIX_EPOCH).unwrap().as_secs();
```
duncanista, line 51: Why seconds?

duncanista (commented):
Left some comments and asked some questions – we might want to avoid using `unwrap()` and `expect()` too

bottlecap/src/lifecycle/flush_control.rs
astuyve: I just find it easier to reason about. That's why I divide off the millis later in this method. We could do millis if you wanted. Not sure we need the precision in this case

bottlecap/src/main.rs
```diff
@@ -156,23 +165,63 @@ fn main() -> Result<()> {
...
+                                TelemetryRecord::PlatformRuntimeDone { request_id, status, .. } => {
+                                    debug!("Runtime done for request_id: {:?} with status: {:?}", request_id, status);
+                                    debug!("FLUSHING ALL");
+                                    dogstats_client.flush();
+                                    debug!("CALLING FOR NEXT EVENT");
+                                    break;
+                                },
+                                TelemetryRecord::PlatformReport { request_id, status, .. } => {
+                                    debug!("Platform report for request_id: {:?} with status: {:?}", request_id, status);
+                                },
```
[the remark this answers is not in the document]
astuyve: this is just for development at this point, we only need to catch events here which we care about. Right now, that's only PlatformRuntimeDone.

bottlecap/src/config.rs
```diff
@@ -17,6 +18,56 @@ pub enum LogLevel {
...
+                    },
+                    _ => {
+                        debug!("invalid flush strategy: {}, using default", value);
+                        Ok(FlushStrategy::Default)
+                    }
+                }
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
astuyve: I think we need each match here, but I can refactor them to `if let` which may make it easier to fall through to the default strategy, although I kinda like the explicitness

bottlecap/src/config.rs
```diff
@@ -39,6 +96,7 @@ impl Default for Config {
             // General
             site: "datadoqhq.com".to_string(),
```
maxday, line 97: maybe let's add a TODO to make sure we don't forget this?

bottlecap/src/main.rs
```diff
@@ -142,8 +148,11 @@ fn main() -> Result<()> {
         .subscribe()
         .map_err(|e| Error::new(std::io::ErrorKind::InvalidData, e.to_string()))?;
 
+    let mut flush_control = FlushControl::new(config.serverless_flush_strategy);
+
     loop {
         let evt = next_event(&r.extension_id);
+        let mut shutdown = false;
```
maxday, line 155: could we move this outside of the loop?

bottlecap/src/main.rs
```diff
@@ -156,23 +165,76 @@ fn main() -> Result<()> {
...
+                                }
+                                TelemetryRecord::PlatformReport {
+                                    request_id, status, ..
+                                } => {
+                                    debug!(
+                                        "Platform report for request_id: {:?} with status: {:?}",
+                                        request_id, status
+                                    );
+                                }
+                                _ => {}
```
maxday, line 224: let's add an warn/error log to report any unsupported event?

bottlecap/src/main.rs
```diff
@@ -156,23 +165,76 @@ fn main() -> Result<()> {
...
+                        }
+                        _ => {}
+                    }
+                } else {
+                    error!("could not get the event");
+                }
+            }
+        }
+        if shutdown {
+            dogstats_client.shutdown();
```
maxday, line 235: why can't we do that when receiving the shutdown event in the main loop?

bottlecap/src/config.rs
```diff
@@ -183,4 +241,90 @@ pub mod tests {
             Ok(())
         });
     }
+
```
maxday, line 244: 💯

duncanista (approved):

bottlecap/src/config.rs
astuyve: Yup will do

bottlecap/src/main.rs
astuyve: sure

bottlecap/src/main.rs
astuyve: I'll just make this a debug for now because I expect we move them up into the match as we further develop things.

bottlecap/src/main.rs
astuyve: `shutdown` is a long-poll response from the `/next` endpoint, which means until that point we are frozen.

I moved `shutdown` out of main loop and into this conditional after we block on the telemetry API so that we can receive the final logs and metrics (eg `report` line) from the telemetry API before we shutdown. I noticed this because on shutdown events, we never flushed the last metric/log.

bottlecap/src/config.rs
maxday: oh this is Default sorry! Yes my bad, all good

PR #217, opened 2024-05-02
feat(bottlecap): send logs to Datadog

duncanista:
# What?

Send Telemetry API logs to Datadog.

# How?

Created the Logs Agent which handles the coordination for:
1. Logs Aggregator (needs batching mechanism).
2. Logs Processor: convert `TelemetryEvents` into `IntakeLog`s.
3. Logs API `datadog.rs`: sending data to backend.

# Testing

- [x] manual testing
- [x] unit testing for processing, logs agent e2e testing will be needed later 

# Motivation

[SVLS-4809](https://datadoghq.atlassian.net/browse/SVLS-4809)

[SVLS-4809]: https://datadoghq.atlassian.net/browse/SVLS-4809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/logs/agent.rs
+++ b/bottlecap/src/logs/agent.rs
@@ -1,0 +1,78 @@
+use std::sync::mpsc::{self, Sender};
+use std::sync::{Arc, Mutex};
+use std::thread;
+
+use tracing::{debug, error};
+
+use crate::logs::aggregator::Aggregator;
+use crate::logs::datadog::DdApi;
+use crate::logs::processor::Processor;
+use crate::telemetry::events::TelemetryEvent;
+
+pub struct LogsAgent {
+    function_arn: Option<String>,
+    dd_api: DdApi,
+    aggregator: Arc<Mutex<Aggregator>>,
+    tx: Sender<TelemetryEvent>,
+    join_handle: std::thread::JoinHandle<()>,
+}
+
+impl LogsAgent {
+    pub fn run(function_arn: &str) -> LogsAgent {
+        let function_arn = function_arn.to_string();
+        let aggregator: Arc<Mutex<Aggregator>> = Arc::new(Mutex::new(Aggregator::new()));
+        let processor: Processor = Processor::new(function_arn.clone());
+
+        let cloned_aggregator = Arc::clone(&aggregator);
+
+        let (tx, rx) = mpsc::channel::<TelemetryEvent>();
+        let join_handle = thread::spawn(move || loop {
+            let received = rx.recv();
+            if let Ok(event) = received {
+                match processor.process(event) {
+                    Ok(log) => {
+                        cloned_aggregator.lock().expect("lock poisoned").add(log);
+                    }
+                    Err(e) => {
+                        error!("Error processing event: {}", e);
+                    }
+                }
+            }
+        });
+        LogsAgent {
+            tx,
+            join_handle,
+            dd_api: DdApi::new(),
+            aggregator,
+            function_arn: None,
+        }
+    }
+
+    pub fn get_sender_copy(&self) -> Sender<TelemetryEvent> {
+        self.tx.clone()
+    }
+
+    pub fn set_function_arn(&mut self, function_arn: String) {
+        self.function_arn = Some(function_arn);
+    }
+
+    pub fn flush(&self) {
+        let logs = self.aggregator.lock().expect("lock poisoned").get_batch();
+        self.dd_api
+            .send(&logs)
+            .expect("failed to send logs to Datadog");
+    }
+
+    pub fn shutdown(self) {
+        debug!("Shutting down LogsAgent");
+        self.flush();
+        match self.join_handle.join() {
+            Ok(_) => {
+                debug!("LogsAgent thread has been shutdown");
+            }
+            Err(e) => {
+                debug!("Error shutting down the LogsAgent thread: {:?}", e);
+            }
+        }
+    }
+}
--- a/bottlecap/src/logs/agent.rs
+++ b/bottlecap/src/logs/agent.rs
@@ -11,7 +11,7 @@ use crate::logs::processor::Processor;
 use crate::telemetry::events::TelemetryEvent;
 
 pub struct LogsAgent {
-    dd_api: datadog::DdApi,
+    dd_api: datadog::Api,
     aggregator: Arc<Mutex<Aggregator>>,
     tx: Sender<TelemetryEvent>,
     join_handle: std::thread::JoinHandle<()>,
@@ -23,7 +23,7 @@ impl LogsAgent {
         let aggregator: Arc<Mutex<Aggregator>> = Arc::new(Mutex::new(Aggregator::default()));
         let processor: Processor = Processor::new(function_arn, Arc::clone(&datadog_config));
 
-        let cloned_aggregator = Arc::clone(&aggregator);
+        let cloned_aggregator = aggregator.clone();
 
         let (tx, rx) = mpsc::channel::<TelemetryEvent>();
         let join_handle = thread::spawn(move || loop {
@@ -40,8 +40,7 @@ impl LogsAgent {
             }
         });
 
-        let dd_api =
-            datadog::DdApi::new(datadog_config.api_key.clone(), datadog_config.site.clone());
+        let dd_api = datadog::Api::new(datadog_config.api_key.clone(), datadog_config.site.clone());
         LogsAgent {
             tx,
             join_handle,
--- a/bottlecap/src/logs/agent.rs
+++ b/bottlecap/src/logs/agent.rs
@@ -28,14 +28,19 @@ impl LogsAgent {
         let (tx, rx) = mpsc::channel::<TelemetryEvent>();
         let join_handle = thread::spawn(move || loop {
             let received = rx.recv();
-            if let Ok(event) = received {
-                match processor.process(event) {
-                    Ok(log) => {
-                        cloned_aggregator.lock().expect("lock poisoned").add(log);
-                    }
-                    Err(e) => {
-                        error!("Error processing event: {}", e);
-                    }
+            // TODO(duncanista): we might need to create a Event::Shutdown
+            // to signal shutdown and make it easier to handle any floating events
+            let Ok(event) = received else {
+                debug!("Failed to received event in Logs Agent");
+                break;
+            };
+
+            match processor.process(event) {
+                Ok(log) => {
+                    cloned_aggregator.lock().expect("lock poisoned").add(log);
+                }
+                Err(e) => {
+                    error!("Error processing event: {}", e);
                 }
             }
         });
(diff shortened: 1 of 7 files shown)


bottlecap/src/logs/aggregator.rs
```diff
@@ -0,0 +1,23 @@
+use crate::logs::processor::IntakeLog;
+
+pub struct Aggregator {
+    messages: Vec<IntakeLog>,
+}
+
+impl Aggregator {
```
maxday, line 7: Here you could use the #derive Default and remove the new() function

bottlecap/src/logs/aggregator.rs
```diff
@@ -0,0 +1,23 @@
...
+            messages: Vec::new(),
+        }
+    }
+
+    pub fn add(&mut self, log: IntakeLog) {
+        self.messages.push(log);
+    }
+
+    pub fn get_batch(&mut self) -> Vec<IntakeLog> {
+        let messages = self.messages.clone();
```
maxday, line 19: why do you clone messages here?

bottlecap/src/logs/datadog.rs
```diff
@@ -0,0 +1,48 @@
...
+
+pub struct DdApi {
+    api_key: String,
+    ureq_agent: ureq::Agent,
+}
+
+impl DdApi {
+    pub fn new() -> Self {
+        DdApi {
+            api_key: env::var("DD_API_KEY").expect("unable to read envvars, catastrophic failure"),
```
maxday, line 17: the API key could live in datadog.yml file as well, maybe it would be easier to get it from the config?

bottlecap/src/logs/datadog.rs
```diff
@@ -0,0 +1,48 @@
...
+            ureq_agent: ureq::AgentBuilder::new().build(),
+        }
+    }
+
+    pub fn send(&self, logs: &Vec<IntakeLog>) -> Result<(), Box<dyn Error>> {
+        let api_key = self.api_key.clone();
```
maxday, line 23: why do you need a clone here?

bottlecap/src/logs/processor.rs
```diff
@@ -0,0 +1,120 @@
...
+pub struct Processor {
+    function_arn: String,
+}
+
+impl Processor {
+    pub fn new(function_arn: String) -> Processor {
+        Processor { function_arn }
+    }
+
+    pub fn process(&self, event: TelemetryEvent) -> Result<IntakeLog, Box<dyn Error>> {
```
maxday, line 44: Could you have unit tests around this?

bottlecap/src/main.rs
```diff
@@ -123,6 +125,7 @@ fn main() -> Result<()> {
 
     let r = register().map_err(|e| Error::new(std::io::ErrorKind::InvalidData, e.to_string()))?;
 
+    let mut logs_agent = LogsAgent::run("TODO"); // todo how to add function arn?
```
maxday, line 128: I think you can build the arn from:
- env vars
- accountID from Lambda-Extension-Accept-Feature (see: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#extensions-registration-api-a)

bottlecap/src/main.rs
```diff
@@ -156,6 +159,8 @@ fn main() -> Result<()> {
                     deadline_ms,
                     invoked_function_arn
                 );
+                logs_agent.set_function_arn(invoked_function_arn);
```
maxday, line 162: it could be removed if you can build the arn from env (see previous comment)
and arn would not need to be an Option the LogAgent

maxday:
Great work! Left some minor comments, let me know!

bottlecap/src/logs/datadog.rs
duncanista: Yup, I'm going to create another PR for this

bottlecap/src/logs/datadog.rs
```diff
@@ -0,0 +1,48 @@
...
+                .ureq_agent
+                .post(LOGS_API_INTAKE)
+                .set("DD-API-KEY", &api_key)
+                .set("DD-PROTOCOL", "agent-json")
+                .set("Content-Type", "application/json")
+                .send_json(logs);
+
+            match resp {
+                Ok(resp) => {
+                    if resp.status() != 200 {
```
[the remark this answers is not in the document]
duncanista: Yes, I'm modifying this whole block

bottlecap/src/logs/processor.rs
duncanista: Yes, there's not that many unit tests yet

bottlecap/src/main.rs
duncanista: Oh, nice perfect, thanks for this

bottlecap/src/logs/datadog.rs
duncanista: Modified it

bottlecap/src/logs/processor.rs
duncanista: LMK if I should add more!

bottlecap/src/logs/agent.rs
```diff
@@ -0,0 +1,76 @@
...
+    join_handle: std::thread::JoinHandle<()>,
+}
+
+impl LogsAgent {
+    pub fn run(function_arn: &str, datadog_config: Arc<config::Config>) -> LogsAgent {
+        let function_arn = function_arn.to_string();
+        let aggregator: Arc<Mutex<Aggregator>> = Arc::new(Mutex::new(Aggregator::default()));
+        let processor: Processor = Processor::new(function_arn, Arc::clone(&datadog_config));
+
+        let cloned_aggregator = Arc::clone(&aggregator);
```
DarcyRaynerDD, line 26: Can you just call .clone()?

bottlecap/src/logs/processor.rs
```diff
@@ -0,0 +1,339 @@
...
+            },
+            TelemetryRecord::PlatformStart {
+                request_id,
+                version,
+            } => {
+                let version = version.unwrap_or("$LATEST".to_string());
+                log.message.message =
+                    format!("START RequestId: {} Version: {}", request_id, version);
+                log.message.lambda.request_id = request_id;
+                log.message.status = "info".to_string();
```
maxday, line 99: since "info".to_string() is repeated, maybe we could use a const &str ?

maxday (approved):

bottlecap/src/logs/agent.rs
```diff
@@ -0,0 +1,76 @@
...
+        let logs = self.aggregator.lock().expect("lock poisoned").get_batch();
+        self.dd_api
+            .send(&logs)
+            .expect("failed to send logs to Datadog");
+    }
+
+    pub fn shutdown(self) {
+        debug!("Shutting down LogsAgent");
+        self.flush();
+        match self.join_handle.join() {
```
DarcyRaynerDD, line 67: Calling .join won't actually shutdown the thread. Based on the logic you used up above, the loop will continue until the tx channel closes. But you aren't closing that here. Maybe a better name, "waitForShutdown"

bottlecap/src/logs/datadog.rs
```diff
@@ -0,0 +1,48 @@
+use std::error::Error;
+
+use tracing::{debug, error};
+
+use crate::logs::processor::IntakeLog;
+
+pub struct DdApi {
```
DarcyRaynerDD, line 7: Maybe give this a more descriptive name than DdApi. Or just call it Api, and use the fact that it's namespaced under logs.

bottlecap/src/logs/processor.rs
duncanista: Modified it so `info` is the default, later, we'd have to set them as a `const &str = "error";` and so on for the others probably!

bottlecap/src/logs/processor.rs
```diff
@@ -0,0 +1,339 @@
...
+mod tests {
+    use crate::telemetry::events::{
+        InitPhase, InitType, ReportMetrics, RuntimeDoneMetrics, Status,
+    };
+
+    use super::*;
+
+    use chrono::{TimeZone, Utc};
+
+    macro_rules! process_tests {
```
DarcyRaynerDD, line 147: I'm impressed you've grokked the macro rules stuff so quickly. I'd recommend that this could probably just be a helper function. Maintaining macro rules blocks tends to be a bit of a pain, and the added verbosity isn't that much. Alternatively, you can use something like https://docs.rs/duplicate/latest/duplicate/ to do the same thing. One other thing to consider, if you want a lot of tests here, and the IntakeOutput might change frequently as we add features, you could use a snapshot testing tool like [insta](https://docs.rs/insta/latest/insta/), which can cut down a lot of verbosity in the code.

bottlecap/src/logs/processor.rs
```diff
@@ -0,0 +1,339 @@
...
+            datadog_config,
+        }
+    }
+
+    pub fn process(&self, event: TelemetryEvent) -> Result<IntakeLog, Box<dyn Error>> {
+        let service = self
+            .datadog_config
+            .service
+            .clone()
+            .unwrap_or("".to_string());
```
DarcyRaynerDD, line 55: You can use "unwrap_or_default" instead, as the default for String is ""

bottlecap/src/logs/processor.rs
```diff
@@ -0,0 +1,339 @@
...
+        let tags = self.datadog_config.tags.clone().unwrap_or("".to_string());
+        let mut log = IntakeLog {
+            hostname: self.function_arn.clone(),
```
DarcyRaynerDD, line 58: Future possible optimization, (not for now), instead of copying basically immutable strings all the time we can use a Copy On Write, (Cow), [string](https://doc.rust-lang.org/std/borrow/enum.Cow.html).

bottlecap/src/logs/agent.rs
DarcyRaynerDD: Another thing to consider. You may want to do the flush after the aggregator thread is shutdown, in case there are inflight requests.

bottlecap/src/logs/agent.rs
```diff
@@ -0,0 +1,76 @@
...
+        let function_arn = function_arn.to_string();
+        let aggregator: Arc<Mutex<Aggregator>> = Arc::new(Mutex::new(Aggregator::default()));
+        let processor: Processor = Processor::new(function_arn, Arc::clone(&datadog_config));
+
+        let cloned_aggregator = Arc::clone(&aggregator);
+
+        let (tx, rx) = mpsc::channel::<TelemetryEvent>();
+        let join_handle = thread::spawn(move || loop {
+            let received = rx.recv();
+            if let Ok(event) = received {
```
DarcyRaynerDD, line 31: Stylistically you can write this as:
```
let Ok(event) = received else {
   break;
}
```
This let's us avoid the creeping indent problem. You also never break from this loop, even when we start getting errors from the channel, (like when it's closed).

bottlecap/src/logs/processor.rs
duncanista: I think I'll address this in another PR to use duplicate/insta on the whole repo, WDYT?

bottlecap/src/logs/agent.rs
duncanista: Resolved!

bottlecap/src/logs/aggregator.rs
duncanista: I dediced to just take them so they are empty now

PR #218, opened 2024-05-03
feat(bottlecap): make configuration an atomic reference counter

duncanista:
# What?

Update configuration to be an `Arc<Config>` so we can use across services and threads.

# How?

Updated `main.rs` to reflect this changes.
Also updated the Metrics API to use the configuration instead of reading from environment variables.

### DIFF
--- a/bottlecap/src/config.rs
+++ b/bottlecap/src/config.rs
@@ -8,7 +8,7 @@ use figment::{
     Figment,
 };
 
-#[derive(Debug, PartialEq, Deserialize, Default)]
+#[derive(Clone, Copy, Debug, PartialEq, Deserialize, Default)]
 pub enum LogLevel {
     Trace,
     Debug,
@@ -18,12 +18,12 @@ pub enum LogLevel {
     Error,
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, PartialEq)]
 pub struct PeriodicStrategy {
     pub interval: u64,
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, PartialEq)]
 pub enum FlushStrategy {
     Default,
     End,
--- a/bottlecap/src/config.rs
+++ b/bottlecap/src/config.rs
@@ -18,12 +18,12 @@ pub enum LogLevel {
     Error,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct PeriodicStrategy {
     pub interval: u64,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum FlushStrategy {
     Default,
     End,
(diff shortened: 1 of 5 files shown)


bottlecap/src/config.rs
```diff
@@ -18,12 +18,12 @@ pub enum LogLevel {
     Error,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
```
DarcyRaynerDD, line 21: This type should probably implement Copy as well.

DarcyRaynerDD, line 21: These types should probably both derive Copy as well, since they are plain old data types, (ie, don't own any heap memory/pointers).

bottlecap/src/metrics/datadog.rs
```diff
@@ -30,37 +28,24 @@ pub enum ShipError {
...
+    pub fn new(api_key: String) -> Self {
+        DdApi {
+            api_key,
             ureq_agent: ureq::AgentBuilder::new().build(),
         }
     }
-}
-
-impl DdApi {
-    /// Create a new `DdApi` instance
```
DarcyRaynerDD, line 43: Makes sense to delete this. By convention, in Rust we use default() instead of new for empty constructors.

bottlecap/src/main.rs
```diff
@@ -118,21 +119,23 @@ fn main() -> Result<()> {
     // First load the configuration
     let lambda_directory = std::env::var("LAMBDA_TASK_ROOT").unwrap_or("".to_string());
     let config = match config::get_config(Path::new(&lambda_directory)) {
-        Ok(config) => config,
+        Ok(config) => Arc::new(config),
```
DarcyRaynerDD, line 122: Using a plain Arc for this makes a lot of sense, since config should never be mutated after this point.

bottlecap/src/logger.rs
```diff
@@ -4,7 +4,7 @@ use log::{LevelFilter, Log, Metadata, Record};
 pub struct SimpleLogger {}
 
 impl SimpleLogger {
-    pub fn init(level: LogLevel) -> Result<(), log::SetLoggerError> {
+    pub fn init(level: &LogLevel) -> Result<(), log::SetLoggerError> {
```
DarcyRaynerDD, line 7: LogLevel derivces Copy and is a small enum, so there isn't much point borrowing (with &) it here vs just copying/moving (without).

DarcyRaynerDD (approved):

bottlecap/src/config.rs
duncanista: On it!

bottlecap/src/metrics/datadog.rs
duncanista: Thanks – I didn't know that!

bottlecap/src/logger.rs
duncanista: The one we are using doesn't, so I'm going to add it, remove the borrowing and move it directly!

bottlecap/src/metrics/datadog.rs
astuyve: ah my bad, good to know

PR #219, opened 2024-05-03
feat(bottlecap): get AWS Account ID from Extension register

duncanista:
# How?

Telling in the request that we are opting for the `accountId` field.

# Motivation

Reconstruct function ARN.
Unblocks #217 

# Testing

Logged the account ID on receipt.

### DIFF
--- a/bottlecap/src/main.rs
+++ b/bottlecap/src/main.rs
@@ -31,8 +31,10 @@ use serde::Deserialize;
 
 const EXTENSION_HOST: &str = "0.0.0.0";
 const EXTENSION_NAME: &str = "datadog-agent";
+const EXTENSION_FEATURES: &str = "accountId";
 const EXTENSION_NAME_HEADER: &str = "Lambda-Extension-Name";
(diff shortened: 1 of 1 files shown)


bottlecap/src/main.rs
```diff
@@ -96,17 +102,23 @@ fn register() -> Result<RegisterResponse> {
...
-    Ok(RegisterResponse {
-        extension_id: ext_id.to_string(),
-    })
+    let extension_id = resp
+        .header(EXTENSION_ID_HEADER)
+        .unwrap_or_default()
+        .to_string();
+    let mut register_response = resp.into_json::<RegisterResponse>()?;
+
+    register_response.extension_id = extension_id;
```
DarcyRaynerDD, line 119: Why the skip_serializing on the extension_id field, and why do we overwrite it here?

DarcyRaynerDD (approved):

bottlecap/src/main.rs
duncanista: It comes as a header, not available in the payload.

PR #224, opened 2024-05-08
feat(bottlecap): batch logs properly

duncanista:
# What?

Add batching logic for the logs `aggregator.rs`.

# How?

Using Datadog's HTTP Logs API limits described in the [docs](https://docs.datadoghq.com/api/latest/logs/?code-lang=curl).

# Motivation

We need it, or else logs are dropped if not sent correctly.
Also [SVLS-4837](https://datadoghq.atlassian.net/browse/SVLS-4837)

# Notes
- Might need to update the flushing mechanism too -> if there are multiple batches on shutdown, we need to try to flush everything on the queue.

[SVLS-4837]: https://datadoghq.atlassian.net/browse/SVLS-4837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/logs/agent.rs
+++ b/bottlecap/src/logs/agent.rs
@@ -62,6 +62,16 @@ impl LogsAgent {
         dd_api.send(&logs).expect("Failed to send logs to Datadog");
     }
 
+    fn flush_shutdown(aggregator: &Arc<Mutex<Aggregator>>, dd_api: &datadog::Api) {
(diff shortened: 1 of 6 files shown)


astuyve (approved):
We may want to use a regular vector at some point and let the delimmiters be handled by the system library, but this is okay for now

PR #226, opened 2024-05-09
Route all logs into the tracing subsystem

blt:
This commit removes the `SimpleLogger` entirely, using the `tracing-log` crate to route any `log` logs into the tracing subsystem. This preserves debug logs et al from `ureq` and avoids contention between two logging subsystems. I've maintained the simple formatter previously used. I've also converted every `log::` call made by our code to the use of `tracing`, further reducing confusion between the two subsystems.

`json` feature is removed from tracing, removing a dependency on `serde_json` and `tracing-serde`. Small win.

### DIFF
--- a/bottlecap/src/config.rs
+++ b/bottlecap/src/config.rs
@@ -150,9 +150,10 @@ pub enum ConfigError {
 
 pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
     let path = config_directory.join("datadog.yaml");
-    let figment = Figment::new().merge(Yaml::file(path));
-    // .merge(Env::prefixed("DATADOG_"))
-    // .merge(Env::prefixed("DD_"));
+    let figment = Figment::new()
+        .merge(Yaml::file(path))
+        .merge(Env::prefixed("DATADOG_"))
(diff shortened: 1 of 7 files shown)


bottlecap/src/config.rs
```diff
@@ -120,10 +150,9 @@ pub enum ConfigError {
 
 pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
     let path = config_directory.join("datadog.yaml");
-    let figment = Figment::new()
-        .merge(Yaml::file(path))
-        .merge(Env::prefixed("DATADOG_"))
-        .merge(Env::prefixed("DD_"));
+    let figment = Figment::new().merge(Yaml::file(path));
+    // .merge(Env::prefixed("DATADOG_"))
```
astuyve, line 154: need these

blt: Oops, my bad.

blt: Addressed in [ce37057](https://github.com/DataDog/datadog-lambda-extension/pull/226/commits/ce370575f00e85c82d4430597d596433e020f686)

astuyve (approved):

PR #235, opened 2024-05-10
Add clippy lints to the project, address some dings 

blt:
This commit is a follow-on from a suggestion I had early on to make aggressive use of clippy. The changes here are mostly mechanical however some needless clones, branches have been caught and fixed. I have left some of the precision loss dings in place, note that lib.rs has a section of lints explicitly allowed. These should be an hour or two to grind out. Some of these lints are also a matter of project taste. 

Assumes that https://github.com/DataDog/datadog-lambda-extension/pull/227 will merge. Should be rebased once that PR is in `main`.

### DIFF
--- a/bottlecap/src/config.rs
+++ b/bottlecap/src/config.rs
@@ -37,6 +37,7 @@ impl AsRef<str> for LogLevel {
 
 impl LogLevel {
     /// Construct a `log::LevelFilter` from a `LogLevel`
+    #[must_use]
     pub fn as_level_filter(self) -> log::LevelFilter {
         match self {
             LogLevel::Error => log::LevelFilter::Error,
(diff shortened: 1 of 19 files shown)


bottlecap/src/telemetry/listener.rs
```diff
@@ -116,13 +122,18 @@ pub struct TelemetryListenerConfig {
 }
 
 impl TelemetryListener {
+    /// Run the `TelemetryListener`
+    ///
+    /// # Errors
+    ///
+    /// Fucntion will error if the passed address cannot be bound.
```
duncanista, line 129:
```suggestion
    /// Function will error if the passed address cannot be bound.
```

bottlecap/src/config.rs
```diff
@@ -68,36 +69,32 @@ impl<'de> Deserialize<'de> for FlushStrategy {
...
+                        Ok(FlushStrategy::Default)
+                    } else {
                         debug!("invalid flush strategy: {}, using default", value);
                         Ok(FlushStrategy::Default)
                     }
                 }
+                _ => {
+                    debug!("invalid flush strategy: {}, using default", value);
+                    Ok(FlushStrategy::Default)
+                }
```
duncanista, line 97: Thanks for improving this

duncanista (approved):
LGTM – left a nit.

bottlecap/src/config.rs
blt: 100% a clippy lint fix.

PR #236, opened 2024-05-10
Use jemalloc as global allocator

blt:
I imagine that the libc allocator in a lambda environment is glibc. While it's improved over time jemalloc is often a better all-around allocator. This commit also demonstrates how one sets a global allocator for a Rust project, park it in the `main.rs` of your binary and everything linked into that binary -- except for C code, as the allocator has a prefix and doesn't overwrite `malloc` et al -- uses the allocator.

### DIFF
--- a/bottlecap/Cargo.lock
+++ b/bottlecap/Cargo.lock
@@ -103,6 +103,7 @@ dependencies = [
  "serde",
  "serde_json",
  "thiserror",
+ "tikv-jemallocator",
  "tracing",
  "tracing-core",
  "tracing-log",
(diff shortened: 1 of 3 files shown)


blt:
Note if you wanted to periodically print allocations: https://docs.rs/tikv-jemalloc-ctl/latest/tikv_jemalloc_ctl/

The base crate also allows you to make builds -- via a feature flag -- that dumps profiling material around heap allocations out. Potentially of interest.

blt:
Worth also noting that this allocator takes up space in the binary, where the system allocator does not.

astuyve (approved):

PR #240, opened 2024-05-10
aj/bound event bus

astuyve:
- **feat: bind event bus**
- **event bus constants**

### DIFF
--- a/bottlecap/src/event_bus/bus.rs
+++ b/bottlecap/src/event_bus/bus.rs
@@ -2,7 +2,7 @@ use std::sync::mpsc::{self, SyncSender};
 
 use crate::events;
 
-use super::constants::MAX_EVENTS;
+use crate::event_bus::constants::MAX_EVENTS;
 
 pub struct EventBus {
     tx: SyncSender<events::Event>,
--- a/bottlecap/src/event_bus/bus.rs
+++ b/bottlecap/src/event_bus/bus.rs
@@ -1,20 +1,21 @@
-use std::sync::mpsc;
-use std::sync::mpsc::Sender;
+use std::sync::mpsc::{self, SyncSender};
 
 use crate::events;
 
+use super::constants::MAX_EVENTS;
+
 pub struct EventBus {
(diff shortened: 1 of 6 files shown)


bottlecap/src/event_bus/constants.rs
```diff
@@ -0,0 +1 @@
+pub const MAX_EVENTS: usize = 100;
```
duncanista, line 1: Why 100?

bottlecap/src/event_bus/bus.rs
```diff
@@ -1,20 +1,21 @@
-use std::sync::mpsc;
-use std::sync::mpsc::Sender;
+use std::sync::mpsc::{self, SyncSender};
 
 use crate::events;
 
+use super::constants::MAX_EVENTS;
```
duncanista, line 5: why `super` and not `crate::constants::MAX_EVENTS`?

duncanista (approved):
Left some questions – but overall, LGTM! :shipit:

bottlecap/src/event_bus/bus.rs
astuyve: This is just how my editor imported it, but I think we should use crate here in case we want to move the constants file somewhere else in the future. Will change.

bottlecap/src/event_bus/constants.rs
astuyve: Just a number I picked. We can make it configurable, I can add a TODO item and ticket

duncanista: Yeah – I'm just afraid we might drop data in scenarios where customer is sending tons of metrics + logs + eventually traces

astuyve: doesn't drop, just blocks until the messages are acknowledged by the main thread.

duncanista: Ah thank you, that's good to know! Thank you for explaining it

PR #243, opened 2024-05-15
feat(bottlecap): add logs processing rules

duncanista:
# What?

Allows users to set `DD_LOGS_PROCESSING_RULES` to mask, exclude, or include, logs as desired.

An AWS Lambda with the configuration:
```
DD_LOGS=PROCESSING_RULES=[{"type": "mask_sequences", "name": "masked_value", "replace_placeholder": "masked", "pattern": "my-secret-value"},{"type": "exclude_at_match", "name": "exclude", "pattern": "exclude-me"}]
```
And the following lines in the code:
```js
console.log('this is my secret value: my-secret-value')
console.log('this log will be excluded because it matches the pattern "exclude-me"');
```
Will produce the following logs:
Screenshot 2024-05-15 at 1 50 52 PM

Where the first log was masked, and the second one was omitted.

# How?

Created a new module under `config`: `processing_rules.rs` which should use a custom deserializer set on the config.

# Motivation
Parity and [SLS-4807](https://datadoghq.atlassian.net/browse/SVLS-4807)

# Notes:
1. This doesn't work properly for YAML, since the field is defined as:
```yaml
// ... rest of config
logs_config:
    processing_rules:
        - type: exclude_at_match
           name: "exclude"
           pattern: "exclude-me"
        // ... more rules
```
And our current config doesn't support nested fields. This will be an issue for `trace_config`, and others, in the future.

### DIFF
--- a/bottlecap/src/logs/processor.rs
+++ b/bottlecap/src/logs/processor.rs
@@ -2,8 +2,10 @@ use std::error::Error;
 use std::sync::{Arc, Mutex};
 
 use serde::Serialize;
+use tracing::debug;
 
 use crate::config;
+use crate::config::processing_rule;
 use crate::logs::aggregator::Aggregator;
 use crate::telemetry::events::{TelemetryEvent, TelemetryRecord};
 
@@ -54,6 +56,13 @@ pub struct IntakeLog {
     pub source: String,
 }
 
+#[derive(Clone, Debug)]
+pub struct Rule {
+    kind: processing_rule::Kind,
+    regex: regex::Regex,
+    placeholder: String,
+}
+
 #[derive(Clone)]
 pub struct Processor {
     function_arn: String,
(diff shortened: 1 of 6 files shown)


bottlecap/src/logs/processor.rs
```diff
@@ -200,9 +287,12 @@ impl Processor {
     }
 
     pub fn process(&mut self, event: TelemetryEvent, aggregator: &Arc<Mutex<Aggregator>>) {
-        if let Ok(log) = self.make_log(event) {
+        if let Ok(mut log) = self.make_log(event) {
             let mut aggregator = aggregator.lock().expect("lock poisoned");
-            aggregator.add(log);
+            let should_send_log = self.apply_rules(&mut log);
```
astuyve, line 292: ah I didn't realize that log processing rules encompassed both redaction _and_ deciding if we should keep or drop a log.

That API makes sense after some consideration, but it was jarring to me at first

duncanista: Yup, I'm not happy with the readability tbh, I'm open for suggestions! 😿

astuyve: Nah I think it's okay, ship it!

PR #245, opened 2024-05-15
aj/tags

astuyve:
Creates a concept of a tag provider, which currently implements a Lambda tag provider.

Uses tags on all metrics.

several TODOs here including a few additional tags which are tricky to find.


bottlecap/src/tags/lambda/tags.rs
```diff
@@ -0,0 +1,162 @@
...
+// TODO(astuyve): set runtime
+// const RUNTIME_VAR: &str = "AWS_EXECUTION_ENV";
+const MEMORY_SIZE_VAR: &str = "AWS_LAMBDA_FUNCTION_MEMORY_SIZE";
+const INIT_TYPE: &str = "AWS_LAMBDA_INITIALIZATION_TYPE";
+const INIT_TYPE_KEY: &str = "init_type";
+
+// FunctionARNKey is the tag key for a function's arn
+const FUNCTION_ARN_KEY: &str = "function_arn";
+// FunctionNameKey is the tag key for a function's name
+const FUNCTION_NAME_KEY: &str = "functionname";
```
duncanista, line 22: Should we add `AWS_LAMBDA_FUNCTION_NAME` as a tag here too? or we don't care about it in this module?

astuyve: we [don't care about that](https://github.com/DataDog/datadog-agent/blob/main/pkg/serverless/tags/tags.go#L242)

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -161,10 +163,16 @@ fn main() -> Result<()> {
 
     let logs_agent = LogsAgent::run(&function_arn, Arc::clone(&config));
     let event_bus = EventBus::run();
+    let mut metadata_hash = hash_map::HashMap::new();
+    metadata_hash.insert("function_arn".to_string(), function_arn.clone());
```
duncanista, line 167: Can't we just initialize it with the data?

bottlecap/src/logs/processor.rs
```diff
@@ -522,7 +522,7 @@ mod tests {
                     arn: "test-arn".to_string(),
                     request_id: Some("test-request-id".to_string()),
                 },
-                timestamp: 1673061827000,
+                timestamp: 1_673_061_827_000,
```
duncanista, line 525: I think these were already solved, can we merge main into this branch?

bottlecap/src/bin/bottlecap/main.rs
astuyve: TIL hashmap::from() was introduced in 1.56

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -168,10 +170,16 @@ fn main() -> Result<()> {
 
     let logs_agent = LogsAgent::run(&function_arn, Arc::clone(&config));
     let event_bus = EventBus::run();
+    let metadata_hash =
+        hash_map::HashMap::from([("function_arn".to_string(), function_arn.clone())]);
```
duncanista, line 174:
```suggestion
        hash_map::HashMap::from([(tags::lambda::tags::FUNCTION_ARN_KEY, function_arn.clone())]);
```
?

PR #246, opened 2024-05-16
feat: Tags 4 logs

astuyve:
Adds the tag provider for logs instead of simply the config tags.

This adds global tags like function ARN:
image

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -167,21 +167,20 @@ fn main() -> Result<()> {
     let function_name =
         env::var("AWS_LAMBDA_FUNCTION_NAME").expect("could not read AWS_LAMBDA_FUNCTION_NAME");
     let function_arn = build_function_arn(&r.account_id, &region, &function_name);
-
-    let logs_agent = LogsAgent::run(&function_arn, Arc::clone(&config));
-    let event_bus = EventBus::run();
     let metadata_hash = hash_map::HashMap::from([(
         lambda::tags::FUNCTION_ARN_KEY.to_string(),
         function_arn.clone(),
     )]);
     let tags_provider =
-        provider::Provider::new(Arc::clone(&config), "lambda".to_string(), &metadata_hash);
+        Arc::new(provider::Provider::new(Arc::clone(&config), "lambda".to_string(), &metadata_hash));
+    let logs_agent = LogsAgent::run(&function_arn, Arc::clone(&tags_provider), Arc::clone(&config));
+    let event_bus = EventBus::run();
     let dogstatsd_config = DogStatsDConfig {
         host: EXTENSION_HOST.to_string(),
         port: DOGSTATSD_PORT,
         datadog_config: Arc::clone(&config),
         function_arn: function_arn.clone(),
-        tags_provider: Arc::new(tags_provider),
+        tags_provider: Arc::clone(&tags_provider),
     };
     let mut dogstats_client = DogStatsD::run(&dogstatsd_config, event_bus.get_sender_copy());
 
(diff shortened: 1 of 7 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -167,21 +167,27 @@ fn main() -> Result<()> {
...
-    let event_bus = EventBus::run();
     let metadata_hash = hash_map::HashMap::from([(
         lambda::tags::FUNCTION_ARN_KEY.to_string(),
         function_arn.clone(),
     )]);
-    let tags_provider =
-        provider::Provider::new(Arc::clone(&config), "lambda".to_string(), &metadata_hash);
+    let tags_provider = Arc::new(provider::Provider::new(
+        Arc::clone(&config),
+        "lambda".to_string(),
```
duncanista, line 176: If this keeps repeating, we might have to set it to a constant, I think I've seen `"lambda".to_string()` a couple times now

bottlecap/src/logs/agent.rs
```diff
@@ -20,10 +21,15 @@ pub struct LogsAgent {
...
+    pub fn run(
+        function_arn: &str,
+        tags_provider: Arc<provider::Provider>,
+        datadog_config: Arc<config::Config>,
+    ) -> LogsAgent {
         let function_arn = function_arn.to_string();
         let aggregator: Arc<Mutex<Aggregator>> = Arc::new(Mutex::new(Aggregator::default()));
-        let mut processor: Processor = Processor::new(function_arn, Arc::clone(&datadog_config));
+        let mut processor: Processor =
+            Processor::new(function_arn, tags_provider, Arc::clone(&datadog_config));
```
duncanista, line 32: If the `tags_provider` contains the function_arn, should we even be sending it now?

duncanista: Or is it too hard since the tags provider only gives us the complete set?

bottlecap/src/bin/bottlecap/main.rs
astuyve: It should be a constant though, at least then the compiler will intern it

bottlecap/src/tags/lambda/tags.rs
```diff
@@ -105,6 +105,16 @@ fn tags_from_env(
     if let Ok(memory_size) = std::env::var(MEMORY_SIZE_VAR) {
         tags_map.insert(MEMORY_SIZE_KEY.to_string(), memory_size);
     }
+
+    if let Some(tags) = &config.tags {
+        for tag in tags.split(',') {
```
duncanista, line 110: This looks like it might fail if someone adds a tag like `a:b,c:b:d` ignoring the second one, should we warn or do something when this happens?

duncanista (approved):
LGTM – left some comments

bottlecap/src/tags/lambda/tags.rs
astuyve: Actually that's completely allowed and we do this intentionally in the version tag. We treat `b:d` as one string.

You can see it in the main agent: https://github.com/DataDog/datadog-agent/blob/main/pkg/serverless/tags/tags.go#L120

which is faithfully replicated here in lambda tags.

bottlecap/src/logs/agent.rs
astuyve: yeah technically we're duplicating here, but the tags_provider uses the function ARN for it's own purposes and immediately carves it up, so it's not meant to be the canonical holder of function ARN.

We should refactor the logs agent to process this from a metadata hashmap (like we did with tag providers) so that we're not coupled to Lambda though.

bottlecap/src/bin/bottlecap/main.rs
duncanista: Ok! I think I saw it in logs too, but maybe they are there for different purposes!

bottlecap/src/logs/agent.rs
duncanista: Sounds good

bottlecap/src/tags/lambda/tags.rs
duncanista: Thanks!

PR #247, opened 2024-05-17
feat: Add lambda enhanced invocation and error metric.

astuyve:
- Refactor metrics to receive an aggregator so we can share it behind a mutex.
- Fixes a tag issue where we'd send a vector of strings, which may contain commas.
- Flattens them and merges tags into one array instead.
- includes an [update](https://github.com/DataDog/saluki/pull/54) to save 10-16kb per distribution context


duncanista (approved):
Amazing work AJ!

PR #248, opened 2024-05-17
chore(bottlecap): decouple Lambda logic from the logs Processor

duncanista:
# What?

Decouple Logs processor from Lambda specific logic.

# How?

Traits and generics.

# Motivation

It's cleaner, and we should be using traits for this.


bottlecap/src/logs/agent.rs
```diff
@@ -28,8 +28,8 @@ impl LogsAgent {
     ) -> LogsAgent {
         let function_arn = function_arn.to_string();
         let aggregator: Arc<Mutex<Aggregator>> = Arc::new(Mutex::new(Aggregator::default()));
-        let mut processor: Processor =
-            Processor::new(function_arn, tags_provider, Arc::clone(&datadog_config));
+        let mut processor: LambdaProcessor =
+            LambdaProcessor::new(function_arn, tags_provider, Arc::clone(&datadog_config));
```
astuyve, line 32: we'll need the logs agent to accept some enum for the processor, instead of the lambda processor, like we do for the lambda tags provider.

duncanista: Per our chat, going to do in another PR, since the change seems quite long!

PR #250, opened 2024-05-20
feat: Create errors metric and timeouts metric

astuyve:
Increments the error metric if the runtimeDone status is not success.
Increments the timeout metric if the status is an error.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -28,7 +28,8 @@ use bottlecap::{
     },
     tags::{lambda, provider},
     telemetry::{
-        self, client::TelemetryApiClient, events::TelemetryRecord, listener::TelemetryListener,
+        self, client::TelemetryApiClient, events::Status, events::TelemetryRecord,
+        listener::TelemetryListener,
     },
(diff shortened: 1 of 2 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -272,14 +273,31 @@ fn main() -> Result<()> {
...
+                                            {
+                                                Ok(()) => {}
+                                                Err(e) => {
+                                                    error!(
+                                                        "Failed to increment timeout metric: {e:?}"
+                                                    );
+                                                }
+                                            }
+                                        }
+                                    }
```
duncanista, line 294: Too many `match`es for code where we aren't doing anything with `Ok`, can't we just do
```rust
if let Err(e) = lambda_enhanced_metrics.increment_errors_metric() {
        error!("Failed to increment errors metric: {e:?}");
};
```

duncanista (approved):
LGTM – let's figure out a pattern so we can handle errors in a different way for the future, I'm afraid we'd eventually end up with a bunch of debug logs for errors

astuyve:
We could modify the methods to handle any failures internally, but all we can really do here is log anyway.

PR #253, opened 2024-05-21
set arch tag

astuyve:
Sets the arch tag from rust's constant. Re-maps aarch64 to arm64

### DIFF
--- a/bottlecap/src/tags/lambda/tags.rs
+++ b/bottlecap/src/tags/lambda/tags.rs
@@ -62,6 +62,13 @@ pub struct Lambda {
     tags_map: hash_map::HashMap<String, String>,
 }
 
+fn arch_to_platform<'a>() -> &'a str {
+    match ARCH {
+        "aarch64" => "arm64",
+        _ => ARCH,
+    }
+}
+
 fn tags_from_env(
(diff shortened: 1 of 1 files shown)


bottlecap/src/tags/lambda/tags.rs
```diff
@@ -105,6 +106,12 @@ fn tags_from_env(
     if let Ok(memory_size) = std::env::var(MEMORY_SIZE_VAR) {
         tags_map.insert(MEMORY_SIZE_KEY.to_string(), memory_size);
     }
+    let arch = match ARCH {
+        "aarch64" => "arm64",
```
duncanista, line 110: I'm not sure if `x86` is going to ever appear, should we: test it until we find it, or remap it either way to `x86` maps to `x86_64`, or naw?

duncanista (approved):

bottlecap/src/tags/lambda/tags.rs
astuyve: x86 would be 32bit.

Here are our current invocations:
image

duncanista: great tysm for clearing that out

PR #254, opened 2024-05-21
feat(bottlecap): add more enhanced metrics

duncanista:
# What?

Adds three new enhanced metrics:
- `runtime_duration`,
- `post_runtime_duration`, and
- `estimated_cost`.

Screenshot 2024-06-07 at 2 43 00 PM

# Motivation

[SVLS-4906](https://datadoghq.atlassian.net/browse/SVLS-4906), [SVLS-4907](https://datadoghq.atlassian.net/browse/SVLS-4907), and [SVLS-4908](https://datadoghq.atlassian.net/browse/SVLS-4908)

# How?

Added an `InvocationContext` buffer which keeps the latest 5 invocations to tracks data to calculate metrics as needed.

[SVLS-4906]: https://datadoghq.atlassian.net/browse/SVLS-4906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SVLS-4907]: https://datadoghq.atlassian.net/browse/SVLS-4907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SVLS-4908]: https://datadoghq.atlassian.net/browse/SVLS-4908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -31,7 +31,9 @@ use bottlecap::{
     },
     tags::{lambda, provider},
     telemetry::{
-        self, client::TelemetryApiClient, events::Status, events::TelemetryRecord,
+        self,
+        client::TelemetryApiClient,
+        events::{Status, TelemetryRecord},
         listener::TelemetryListener,
     },
     DOGSTATSD_PORT, EXTENSION_ACCEPT_FEATURE_HEADER, EXTENSION_FEATURES, EXTENSION_HOST,
@@ -325,11 +327,16 @@ fn main() -> Result<()> {
                                     if let Some(invocation_context) =
                                         invocation_context_buffer.get(&request_id)
                                     {
-                                        let post_runtime_duration_ms = metrics.duration_ms
-                                            - invocation_context.runtime_duration_ms;
-                                        lambda_enhanced_metrics.set_post_runtime_duration_metric(
-                                            post_runtime_duration_ms,
-                                        );
+                                        if invocation_context.runtime_duration_ms > 0.0 {
+                                            let post_runtime_duration_ms = metrics.duration_ms
+                                                - invocation_context.runtime_duration_ms;
+                                            lambda_enhanced_metrics.set_post_runtime_duration_metric(
+                                                post_runtime_duration_ms,
+                                            );
+                                        } else {
+                                            debug!("Impossible to compute post runtime duration for request_id: {:?}", request_id);
+                                        }
+                                        
                                     }
 
                                     if shutdown {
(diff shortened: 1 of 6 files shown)


bottlecap/src/lifecycle/invocation_context.rs
duncanista, line 1: Would have to use this file to replace `execution_context` in Lambda `processor`. Do we want to keep one buffer per process? Do we want to create just one and share it between them?

bottlecap/src/lifecycle/invocation_context.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn default() -> Self {
+        InvocationContextBuffer {
+            buffer: VecDeque::<InvocationContext>::with_capacity(5),
+        }
+    }
+}
+
+impl InvocationContextBuffer {
+    pub fn insert(&mut self, invocation_context: InvocationContext) {
+        if self.buffer.len() == self.buffer.capacity() {
```
alexgallotta, line 25: any reason why using a growable ring buffer?
This seems https://docs.rs/ringbuffer/latest/ringbuffer/struct.ConstGenericRingBuffer.html to do the same stuff in a simpler way
> // Because capacity is reached the next push will be the first item of the buffer.
and with stack allocation it should be more efficient, right?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -264,6 +270,12 @@ fn main() -> Result<()> {
                             debug!("Metric event: {:?}", event);
                         }
                         Event::Telemetry(event) => match event.record {
```
alexgallotta, line 272: Main function is growing a lot with different abstraction levels and loops. Is it reasonable to extract a few methods to make more readable?

alexgallotta (approved):

bottlecap/src/lifecycle/invocation_context.rs
alexgallotta: probably not a good idea to duplicate information/logic around

bottlecap/src/bin/bottlecap/main.rs
duncanista: Yeah, I think it'd make sense to start figuring out how to split them – let's log it!

astuyve (approved):
LGTM

PR #256, opened 2024-05-23
chore(bottlecap): reduce lock contention in logs

duncanista:
# What?

1. Drop logs `Aggregator` lock as soon as possible.
2. Make the Telemetry API listener to send the events directly to the logs agent, and send events of interest to the main thread.
- Previously, we were sending first the events to the main thread event bus, which was bounded to 100 elements, and we had to wait for them to be processed there to be forwarded to the logs agent. This is not ideal.

### DIFF
--- a/bottlecap/src/logs/aggregator.rs
+++ b/bottlecap/src/logs/aggregator.rs
@@ -38,10 +38,16 @@ impl Aggregator {
         }
     }
 
-    pub fn add<T: Serialize>(&mut self, log: T) {
+    fn add<T: Serialize>(&mut self, log: T) {
         match serde_json::to_string(&log) {
             Ok(log) => self.messages.push_back(log),
             Err(e) => debug!("Failed to serialize log: {}", e),
+        }
+    }
+
+    pub fn add_batch<T: Serialize>(&mut self, logs: Vec<T>) {
+        for log in logs {
+            self.add(log);
         }
     }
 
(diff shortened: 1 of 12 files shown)


bottlecap/src/logs/aggregator.rs
```diff
@@ -1,11 +1,10 @@
-use serde::Serialize;
-use std::collections::VecDeque;
+use crossbeam::queue::SegQueue;
```
blt, line 1: FWIW SegQueue is an unbounded multi-producer, multi-consumer. Minor issue, the multi-consumer bit requires more coordination in the implementation so you're leaving performance on the table by not using a single-consumer. More pressing, this is an unbounded queue and represents a potential site for unbounded allocations.

duncanista: I'm aware of the tradeoffs, I have to clean this PR still, will mark it as draft

blt: SegQueue is also theoretically lock free but it uses Acquire/Release semantics internally and so the semantics are equivalent to a mutex. Except Acq/Rel done on atomic structures cuts the Linux scheduler out so your latency will be worse generally speaking, except in specialized circumstances where the acq/rel using software has the highest CPU priority anyway and is always scheduled first. 

Tokio's mpsc is good, will play well with the Linux scheduler and also nicely between async/sync code.

duncanista:
Closing due to other PRs doing this

PR #259, opened 2024-05-30
fix(bottlecap): fix how Telemetry API stream is being read

duncanista:
# How?

Created a new `from_stream` method, which reads headers until `content-length` is found, and proceeds to read the exact remaining body size.

# Motivation

- Data loss and timeouts were happening due to a stream hanging waiting for data
- Stream sometimes received the data chunked, which we were never reading.

First developed in #256, also [SVLS-4932](https://datadoghq.atlassian.net/browse/SVLS-4932)

Screenshot 2024-05-31 at 2 39 34 PM

[SVLS-4932]: https://datadoghq.atlassian.net/browse/SVLS-4932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/telemetry/listener.rs
+++ b/bottlecap/src/telemetry/listener.rs
@@ -18,6 +18,8 @@ pub struct HttpRequestParser {
 
 const CR: u8 = b'\r';
 const LR: u8 = b'\n';
+const MAX_BUFFER_SIZE: usize = 262_144;
+const CHUNK_SIZE: usize = 32_768;
 
 impl HttpRequestParser {
     /// Create a `HttpRequestParser` from a `TcpStream`
@@ -41,7 +43,7 @@ impl HttpRequestParser {
                 break;
             }
 
-            let mut temp_buf = [0u8; 32_768];
+            let mut temp_buf = [0u8; CHUNK_SIZE];
             let bytes_read = match stream.read(&mut temp_buf) {
                 Ok(n) => n,
                 Err(ref e) if e.kind() == std::io::ErrorKind::WouldBlock => {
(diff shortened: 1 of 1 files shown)


bottlecap/src/telemetry/listener.rs
```diff
@@ -18,21 +18,62 @@ pub struct HttpRequestParser {
...
+            if parser.headers.contains_key("content-length") {
+                break;
+            }
+        }
+
+        let body_start_index = parser.parse_headers(&headers_buf)?;
+        let content_length = parser
+            .headers
+            .get("content-length")
+            .expect("unfallible")
```
astuyve, line 63: do we want to expect here? can't we just drop the request if it's malformed or something broke in our assumption?

duncanista: The loop before that ensures that the key exist, which means that this is infallible, wdyt?

bottlecap/src/telemetry/listener.rs
```diff
@@ -161,15 +201,10 @@ impl TelemetryListener {
     }
 
     fn handle_stream(
-        stream: &mut impl Read,
-        mut buf: [u8; 262_144],
+        stream: &TcpStream,
```
astuyve, line 204: ah yeah I made this change in my branch too, a bit easier to handle than receiving a generic which implements Read.

astuyve (approved):

bottlecap/src/telemetry/listener.rs
astuyve: hmm, I think that's okay. It just makes me nervous

duncanista: I could definitely add another `if let Some(...)` block if that makes you feel more confident

PR #261, opened 2024-05-31
add kms handling

alexgallotta:


bottlecap/src/secrets/decrypt.rs
```diff
@@ -43,23 +56,75 @@ pub fn resolve_secrets(config: Config) -> Result<Config> {
...
 struct AwsConfig {
     region: String,
     aws_access_key_id: String,
     aws_secret_access_key: String,
     aws_session_token: String,
+    function_name: String,
 }
 
-fn manual_decrypt(secret_arn: String, aws_config: AwsConfig) -> Result<String> {
+fn decrypt_aws_kms(client: Client, kms_key: String, aws_config: AwsConfig) -> Result<String> {
```
duncanista, line 74: If we are already returning a result, shouldn't we also remove all the expects in the code and bubble up the possible errors?

bottlecap/src/secrets/decrypt.rs
```diff
@@ -43,23 +56,75 @@ pub fn resolve_secrets(config: Config) -> Result<Config> {
...
+        },
+    );
+
+    let v = request(json_body, headers, client);
 
-    let client = reqwest::blocking::Client::builder()
-        .use_rustls_tls()
-        .build()
-        .expect("Failed to create reqwest client for aws decrypt");
+    return if let Some(secret_string) = v["SecretString"].as_str() {
```
duncanista, line 120: Return is not needed if last line is already there, same suggestion as before, maybe we should do a Result<String, ...> where we can bubble up errors?

bottlecap/src/secrets/decrypt.rs
```diff
@@ -74,32 +139,28 @@ fn manual_decrypt(secret_arn: String, aws_config: AwsConfig) -> Result<String> {
         .text()
         .expect("Cannot deserialize body");
     let v: Value = serde_json::from_str(&body).expect("Failed to parse JSON");
-
-    return if let Some(secret_string) = v["SecretString"].as_str() {
-        Ok(secret_string.to_string())
-    } else {
-        Err(Error::new(std::io::ErrorKind::InvalidData, v.to_string()))
-    };
+    v
```
duncanista, line 142: Same comment as before, should we bubble all errors so we don't have expects here?

bottlecap/src/secrets/decrypt.rs
alexgallotta: That sounds right, thanks!
I used a box dyn error for trying to catch different error types. Let me know if there is a better way to do it!

bottlecap/src/secrets/decrypt.rs
alexgallotta: I'm not a big fan of omitting return, but I will stick with the idiom!

bottlecap/src/secrets/decrypt.rs
alexgallotta: done!

PR #262, opened 2024-05-31
chore(bottlecap): Re-architect Telemetry API events forwarding

duncanista:
# What?

Updates how Telemetry API events are being passed through channels.

Instead of sending them to the main Event Bus, and then forwarding them to the Logs Agent. We directly send them to the Logs Agent, and there we only send the 3 events of interest back to the main Event Bus.

Another side effect is that this optimization reduces lock contention, by sending events as batches, instead of individually.

# Motivation

This makes it generally faster, and reduces the lock contention.

Also [SVLS-4821](https://datadoghq.atlassian.net/browse/SVLS-4821), first developed in #256 

[SVLS-4821]: https://datadoghq.atlassian.net/browse/SVLS-4821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/Cargo.lock
+++ b/bottlecap/Cargo.lock
@@ -81,12 +81,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atomic-waker"
-version = "1.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
-
-[[package]]
 name = "autocfg"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -258,61 +263,58 @@ fn main() -> Result<()> {
                         Event::Metric(event) => {
                             debug!("Metric event: {:?}", event);
                         }
-                        Event::Telemetry(event) => {
-                            logs_agent.send_event(event.clone());
```
DarcyRaynerDD, line 262: I'm having a bit of trouble reading this diff, the only difference is you removed this send_event line, and it changed the identation of the rest of the code right?

bottlecap/Cargo.lock
```diff
@@ -80,12 +80,6 @@ dependencies = [
  "bytemuck",
 ]
 
-[[package]]
```
DarcyRaynerDD, line 83: Nice to see all these dependencies going

DarcyRaynerDD (approved):
LGTM

bottlecap/src/bin/bottlecap/main.rs
duncanista: Correct, it's not being sent anymore, since we changed the place where it's being sent

bottlecap/Cargo.lock
duncanista: Yeah!
This was also causing errors on compiling

PR #266, opened 2024-06-06
Aj/add tokio

astuyve:


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -84,38 +79,50 @@ enum NextEventResponse {
...
+        .send()
+        .await
         .map_err(|e| Error::new(std::io::ErrorKind::InvalidData, e.to_string()))?;
 
     assert!(resp.status() == 200, "Unable to register extension");
 
     let extension_id = resp
-        .header(EXTENSION_ID_HEADER)
-        .unwrap_or_default()
+        .headers()
```
duncanista, line 116: Should we try to avoid using `unwrap` and `expect` here?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -141,6 +150,8 @@ fn main() -> Result<()> {
         }
     };
 
+    error!("astuyve config elapsed: {:?}", env_start.elapsed());
```
duncanista, line 153:
```suggestion
```

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -165,16 +176,23 @@ fn main() -> Result<()> {
...
-    let config = match resolve_secrets(env_config) {
+    let register_duration = std::time::Instant::now();
+    let r = register(&client)
+        .await
+        .map_err(|e| Error::new(std::io::ErrorKind::InvalidData, e.to_string()))?;
+    error!(
+        "astuyve register elapsed: {:?}",
+        register_duration.elapsed()
+    );
+    let config = match resolve_secrets(env_config).await {
```
duncanista, line 189:
```suggestion
        .map_err(|e| Error::new(std::io::ErrorKind::InvalidData, e.to_string()))?;

    let config = match resolve_secrets(env_config).await {
```

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -165,16 +176,23 @@ fn main() -> Result<()> {
     tracing::subscriber::set_global_default(subscriber).expect("setting default subscriber failed");
 
     info!("logging subsystem enabled");
+    let client = reqwest::Client::new();
 
-    let config = match resolve_secrets(env_config) {
+    let register_duration = std::time::Instant::now();
```
duncanista, line 181:
```suggestion
```

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -189,12 +207,20 @@ fn main() -> Result<()> {
...
         Arc::clone(&tags_provider),
         Arc::clone(&config),
         event_bus.get_sender_copy(),
     );
+    let logs_agent_channel = logs_agent.get_sender_copy();
+    let logs_flusher = LogsFlusher::new(Arc::clone(&config), Arc::clone(&logs_agent.aggregator));
+    tokio::spawn(async move {
+        logs_agent.spin().await;
+    });
+    error!("astuyve logs elapsed: {:?}", logs_duration.elapsed());
```
duncanista, line 222:
```suggestion
    let mut logs_agent = LogsAgent::new(
        Arc::clone(&tags_provider),
        Arc::clone(&config),
        event_bus.get_sender_copy(),
    );
    let logs_agent_channel = logs_agent.get_sender_copy();
    let logs_flusher = LogsFlusher::new(Arc::clone(&config), Arc::clone(&logs_agent.aggregator));
    tokio::spawn(async move {
        logs_agent.spin().await;
    });
```

bottlecap/src/metrics/datadog.rs
```diff
@@ -37,38 +37,42 @@ impl DdApi {
...
         match resp {
-            Ok(_resp) => Ok(()),
-            Err(ureq::Error::Status(code, response)) => Err(ShipError::Failure {
-                status: code,
-                body: response
-                    .into_string()
-                    .expect("body could not be converted to string"),
-            }),
+            Ok(resp) => match resp.status() {
+                reqwest::StatusCode::OK => Ok(()),
```
duncanista, line 61:
```suggestion
```

Do we really need it, if the API will always return either 201/202?

bottlecap/src/metrics/datadog.rs
```diff
@@ -84,20 +88,23 @@ impl DdApi {
...
```
duncanista, line 101: Same here

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -27,6 +27,19 @@ impl Lambda {
         self.increment_metric(constants::TIMEOUTS_METRIC)
     }
 
+    pub fn set_init_duration_metric(&self, init_duration_ms: f64) -> Result<(), errors::Insert> {
```
duncanista, line 30: Did you figure out the issue on why sometimes it wasn't appearing? And is this a remediation to it?

bottlecap/src/telemetry/client.rs
```diff
@@ -15,26 +17,26 @@ impl TelemetryApiClient {
...
-        });
-
-        let resp = ureq::put(&url)
-            .set("Content-Type", "application/json")
-            .set(EXTENSION_ID_HEADER, &self.extension_id)
-            .send_json(data);
+        let resp = reqwest::Client::new()
+            .put(&url)
+            .header(EXTENSION_ID_HEADER, &self.extension_id)
+            .json(&serde_json::json!({
```
duncanista, line 25: Is `.json` applying the same method we are doing behind the scenes? If so, can we remove the `&serde_json::json!()` macro?

scripts/Dockerfile.bottlecap.build
```diff
@@ -6,12 +6,14 @@ RUN yum install -y curl gcc gcc-c++ make unzip openssl openssl-devel
 COPY ./scripts/install-protoc.sh /
 RUN chmod +x /install-protoc.sh && /install-protoc.sh
 RUN curl https://sh.rustup.rs -sSf | \
-    sh -s -- --default-toolchain stable -y
+    sh -s -- --default-toolchain nightly-x86_64-unknown-linux-gnu -y
```
duncanista, line 9: This has to be changed so `ARCHITECTURE` passed through the build process is honored.

bottlecap/src/telemetry/listener.rs
```diff
@@ -30,22 +46,27 @@ impl HttpRequestParser {
...
 
         let mut headers_buf = [0u8; HEADERS_BUFFER_SIZE];
         loop {
-            match stream.read(&mut headers_buf) {
+            // select here
+            match stream.read(&mut headers_buf).await {
+                Ok(0) => {
+                    error!("astuyve Connection closed by client");
+                    return Err(TcpError::Close("Connection closed by client".to_string()));
+                }
```
duncanista:
```suggestion
            match stream.read(&mut headers_buf).await {
```
Do we really need to check a connection close?

bottlecap/src/telemetry/listener.rs
```diff
@@ -15,6 +12,25 @@ pub struct HttpRequestParser {
...
 
+#[derive(Debug, PartialEq)]
+pub enum TcpError {
+    Close(String),
+    Read(String),
+    Write(String),
+    Parse(String),
+}
+
+impl ToString for TcpError {
```
duncanista: I don't think the name is adequate in this context, because `Parse` doesn't go with the TCP context, would it make sense to rename it? WDYT?

bottlecap/src/telemetry/listener.rs
```diff
@@ -60,14 +81,18 @@ impl HttpRequestParser {
             .headers
             .get("content-length")
             .expect("infallible")
-            .parse::<usize>()?;
+            .parse::<usize>()
+            .map_err(|e| TcpError::Parse(e.to_string()))?;
         let body_bytes_read = headers_buf.len() - body_start_index;
-        let missing_body_lenght = content_length - body_bytes_read;
-        let mut body_buf = vec![0u8; missing_body_lenght];
+        let missing_body_length = content_length - body_bytes_read;
```
duncanista, line 87: lol thanks for fixing this typo

bottlecap/src/bin/bottlecap/main.rs
astuyve: in the test?

duncanista (commented):
Looks amazing AJ, great work on this, left a bunch of nits mostly

bottlecap/src/bin/bottlecap/main.rs
astuyve: the unwrap or default from before was cold comfort - we can't continue if there's no extension ID header in the response. That would be a major failure, and we'd crash anyway. In the future we'll enter a no-op loop, but for now I think this is okay

bottlecap/src/metrics/datadog.rs
astuyve: Removed

bottlecap/src/metrics/enhanced/lambda.rs
astuyve: Not exactly. We don't want to wait on the report log to set init duration because we'll know it from the init report. So this is just an improvement.

bottlecap/src/telemetry/client.rs
astuyve: no, we can drop the macro if we roll this into a struct and derive serde's `Serialize` method, which we do in other places where we already have a struct, but this is just easier

bottlecap/src/telemetry/listener.rs
astuyve: Yes we do, this fixes a bug in our current implementation where we'll sit around forever holding this socket open for this client until we exit (even after the client hung up).

0 is success/`OK` result so it falls into our blocking `wait` state. But if you read the docs for `stream.read` it says `0 means EOF`, so the client closed the connection.

bottlecap/src/telemetry/listener.rs
astuyve: I guess I interpreted it as we can't parse the message on the TCP socket, but wdyt?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -257,19 +296,21 @@ fn main() -> Result<()> {
...
+                            error!("Metric event: {:?}", event);
                         }
                         Event::Telemetry(event) => match event.record {
                             TelemetryRecord::PlatformInitReport {
                                 initialization_type,
                                 phase,
                                 metrics,
                             } => {
-                                debug!("Platform init report for initialization_type: {:?} with phase: {:?} and metrics: {:?}", initialization_type, phase, metrics);
+                                error!("Platform init report for initialization_type: {:?} with phase: {:?} and metrics: {:?}", initialization_type, phase, metrics);
```
duncanista, line 311:
```suggestion
                                debug!("Platform init report for initialization_type: {:?} with phase: {:?} and metrics: {:?}", initialization_type, phase, metrics);
```

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -257,19 +296,21 @@ fn main() -> Result<()> {
...
         if flush_control.should_flush() || shutdown {
             loop {
-                let received = event_bus.rx.recv();
-                if let Ok(event) = received {
+                let received = event_bus.rx.recv().await;
+                if let Some(event) = received {
                     match event {
                         Event::Metric(event) => {
-                            debug!("Metric event: {:?}", event);
+                            error!("Metric event: {:?}", event);
```
duncanista, line 303:
```suggestion
                            debug!("Metric event: {:?}", event);
```

bottlecap/src/telemetry/listener.rs
duncanista: Not 100% convinced, but I understand the interpretation, we can leave it as is for now

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -292,8 +333,8 @@ fn main() -> Result<()> {
                                     "Runtime done for request_id: {:?} with status: {:?}",
                                     request_id, status
                                 );
-                                logs_agent.flush();
-                                dogstats_client.flush();
+                                logs_flusher.flush().await;
+                                statsd_flusher.flush().await;
```
alexgallotta, line 337: Can `statsd_flusher.flush()` start before the `logs_flusher.flush()` is awaited, to do them concurrently?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -323,9 +364,8 @@ fn main() -> Result<()> {
         }
 
         if shutdown {
-            logs_agent.shutdown();
-            dogstats_client.shutdown();
-            telemetry_listener.shutdown();
+            dogstats_cancel_token.cancel();
+            telemetry_listener_cancel_token.cancel();
```
alexgallotta, line 368: should this be a join/wait for the flush to finish rather than a canceling the flush?

bottlecap/src/telemetry/listener.rs
astuyve: ahh you're right, we need to match on this.

bottlecap/src/bin/bottlecap/main.rs
astuyve: they are now concurrent w/ `tokio::join`

bottlecap/src/bin/bottlecap/main.rs
astuyve: canceling stops the main loop tasks, not any of the running tasks to aggregate/flush data. So cancelling first then flushing should be okay. We will test and confirm.

bottlecap/src/telemetry/listener.rs
astuyve: Fixed

scripts/Dockerfile.bottlecap.build
astuyve: Done

scripts/Dockerfile.bottlecap.build
```diff
@@ -1,26 +1,27 @@
...
-    sh -s -- --default-toolchain stable -y
+    sh -s -- --default-toolchain nightly-$PLATFORM-unknown-linux-gnu -y
     ENV PATH=/root/.cargo/bin:$PATH
+RUN rustup component add rust-src --toolchain nightly-$PLATFORM-unknown-linux-gnu
 RUN mkdir -p /tmp/dd
 COPY ./bottlecap /tmp/dd/bottlecap
-RUN cd /tmp/dd/bottlecap && cargo build --release
-RUN cp /tmp/dd/bottlecap/target/release/bottlecap /tmp/dd/bottlecap/bottlecap
+ENV RUSTFLAGS="-C panic=abort -Zlocation-detail=none"
+RUN cd /tmp/dd/bottlecap && cargo +nightly build -Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort --release --target x86_64-unknown-linux-gnu
```
duncanista, line 16: I think we also missed it here

bottlecap/src/secrets/decrypt.rs
```diff
@@ -7,11 +7,11 @@ use sha2::{Digest, Sha256};
...
 use std::io::{Error, Result};
 use std::time::Instant;
-use tracing::debug;
+use tracing::{debug, error};
 
-pub fn resolve_secrets(config: Config) -> Result<Config> {
+pub async fn resolve_secrets(config: Config) -> Result<Config> {
     if !config.api_key.is_empty() {
-        debug!("DD_API_KEY found, not trying to resolve secrets");
+        error!("DD_API_KEY found, not trying to resolve secrets");
```
alexgallotta, line 14: we might want to remove this error too

scripts/Dockerfile.bottlecap.build
astuyve: ahh yeah you're right

bottlecap/src/bin/bottlecap/main.rs
astuyve: Yeah main _desperately_ needs automated tests, you're completely right.

I'm HOPING that we can carve out time to emulate the telemetry API (and maybe share it with the rest of the community?), which would enable a really good e2e test for main.

bottlecap/src/telemetry/listener.rs
```diff
@@ -30,22 +47,26 @@ impl HttpRequestParser {
...
         };
 
         let mut headers_buf = [0u8; HEADERS_BUFFER_SIZE];
         loop {
-            match stream.read(&mut headers_buf) {
+            // select here
+            match stream.read(&mut headers_buf).await {
+                Ok(0) => {
+                    return Err(TcpError::Close("Connection closed by client".to_string()));
+                }
```
duncanista, line 62:
```suggestion
            match stream.read(&mut headers_buf).await {
                Ok(0) => Err(TcpError::Close("Connection closed by client".to_string())),
```

bottlecap/src/telemetry/listener.rs
```diff
@@ -131,29 +157,39 @@ impl HttpRequestParser {
...
-            Some(length) => length.parse::<usize>()?,
-            None => return Err(Box::from("content-length header not found")),
+            Some(length) => length
+                .parse::<usize>()
+                .map_err(|e| TcpError::Parse(e.to_string()))?,
+            None => {
+                return Err(TcpError::Parse(
+                    "content-length header not found".to_string(),
+                ))
+            }
```
duncanista, line 169:
```suggestion
            None => Err(TcpError::Parse(
                    "content-length header not found".to_string(),
                )),
```

bottlecap/src/telemetry/listener.rs
```diff
@@ -167,86 +203,91 @@ impl TelemetryListener {
...
+            listener,
+        })
+    }
+
+    pub async fn spin(self) {
+        loop {
+            tokio::select! {
+                biased;
+                stream = self.listener.accept() => {
+                    match stream {
```
duncanista, line 226: Is it possible to remove this assignment by just doing it inline?

duncanista (approved):

bottlecap/src/telemetry/listener.rs
astuyve: not that I'm aware of? Tokio wants a future to block on and this is how you assign the result of that future. Possibly we could refactor this all into another method but I think we'll pass on that

bottlecap/src/telemetry/listener.rs
astuyve: we want to break out of the loop so we need the return

bottlecap/src/telemetry/listener.rs
astuyve: same here, want to break the loop

PR #278, opened 2024-06-14
Svls 4840 metrics api limit

alexgallotta:
- create methods for serializing the payload within byte size/entry size limits. Not sure if the entry size limit makes sense though

From the docs of `api/v2/series` there is only a byte restriction https://docs.datadoghq.com/api/latest/metrics/#submit-metrics. For the `api/beta/sketches` there is no public docs available I think

- flusher loops over them until everything that can be send, is sent. The skipped metrics/sketches will still pollute the buffer though
- flusher should not lock for all that time. A good solution to avoid excess mutex usage could be to copy the outgoing buffer

### DIFF
--- a/bottlecap/src/metrics/aggregator.rs
+++ b/bottlecap/src/metrics/aggregator.rs
@@ -1,15 +1,16 @@
 //! The aggregation of metrics.
 
-use crate::metrics::metric;
-use crate::metrics::{
-    constants, datadog, errors,
-    metric::{Metric, Type},
+use crate::{
+    metrics::{constants, datadog, errors, metric},
+    tags::provider,
 };
+use datadog::Metric as DatadogMetric;
+use metric::{Metric, Type};
 use std::sync::Arc;
 
-use crate::tags::provider;
 use datadog_protos::metrics::{Dogsketch, Sketch, SketchPayload};
 use hashbrown::hash_table;
+use protobuf::Message;
 use std::time;
 use tracing::error;
 use ustr::Ustr;
@@ -176,6 +177,8 @@ impl Entry {
 pub struct Aggregator<const CONTEXTS: usize> {
     tags_provider: Arc<provider::Provider>,
     map: hash_table::HashTable<Entry>,
+    max_batch_entries_size: usize,
+    max_content_size_bytes: usize,
 }
 
 impl<const CONTEXTS: usize> Aggregator<CONTEXTS> {
@@ -195,6 +198,8 @@ impl<const CONTEXTS: usize> Aggregator<CONTEXTS> {
         Ok(Self {
             tags_provider,
             map: hash_table::HashTable::new(),
+            max_batch_entries_size: constants::MAX_ENTRIES_NUMBER,
+            max_content_size_bytes: constants::MAX_CONTENT_SIZE_BYTES,
         })
     }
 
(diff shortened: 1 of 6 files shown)


bottlecap/src/metrics/aggregator.rs
```diff
@@ -43,11 +44,13 @@ struct DistributionMetric {
 }
 
 #[derive(Debug, Clone)]
+#[repr(transparent)]
```
duncanista, line 47: What does this do?

bottlecap/src/metrics/aggregator.rs
```diff
@@ -174,6 +177,8 @@ impl Entry {
 pub struct Aggregator<const CONTEXTS: usize> {
     tags_provider: Arc<provider::Provider>,
     map: hash_table::HashTable<Entry>,
+    max_batch_entries_size: usize,
```
duncanista, line 180: Nice, this is a great pattern for testing too!

bottlecap/src/metrics/aggregator.rs
```diff
@@ -239,32 +246,78 @@ impl<const CONTEXTS: usize> Aggregator<CONTEXTS> {
...
         sketch_payload
     }
 
+    #[must_use]
+    pub fn distributions_to_protobuf_serialized(&self) -> Vec<u8> {
+        let mut buffer: Vec<u8> = Vec::with_capacity(self.max_content_size_bytes);
+        let mut count_entries = 0;
+        let now = time::SystemTime::now()
+            .duration_since(time::UNIX_EPOCH)
+            .expect("unable to poll clock, unrecoverable")
```
duncanista, line 263: Is there something we can do about this expect? Can we bubble it up?

bottlecap/src/metrics/aggregator.rs
```diff
@@ -239,32 +246,78 @@ impl<const CONTEXTS: usize> Aggregator<CONTEXTS> {
...
+            .as_secs()
+            .try_into()
+            .unwrap_or_default();
+        for entry in &self.map {
+            let Some(sketch) = self.build_sketch(now, entry) else {
```
duncanista, line 268: Should we do something (maybe in another PR) if we cannot build the sketch?

bottlecap/src/metrics/aggregator.rs
```diff
@@ -239,32 +246,78 @@ impl<const CONTEXTS: usize> Aggregator<CONTEXTS> {
...
+                continue;
+            };
+            let pb_size = sketch.compute_size();
+            if pb_size > self.max_content_size_bytes as u64
+                || pb_size + buffer.len() as u64 > self.max_content_size_bytes as u64
```
duncanista, line 273: Do we really need to check if the current proto is bigger than the max content size?
Wouldn't this be caught on `pb_size + buffer.len() ...` ? Even if it was the first one, it would be something like `size + 0 > ...`

bottlecap/src/metrics/aggregator.rs
alexgallotta: This was existing existing code I moved. I think it's worth to address all the expect/panics in a specific PR

bottlecap/src/metrics/aggregator.rs
alexgallotta: It hints the rust compiler to compile the struct using a lightweight representation that have just one primitive field.
In this way, the compiled version will be guaranteed to be the same as only that one field, and all the struct methods will be guaranteed to be 0 cost abstraction

bottlecap/src/metrics/aggregator.rs
alexgallotta: Yes, good point. I mentioned it here
> flusher loops over them until everything that can be send, is sent. The skipped metrics/sketches will still pollute the buffer though

When a sketch is not valid, it makes sense to just remove it, since it will never be valid.
Before addressing this policy, I wanted to check in about the buffer swap solution, since it would simplify this issue

bottlecap/src/metrics/aggregator.rs
alexgallotta: oh good catch, I missed it in this method

bottlecap/src/metrics/aggregator.rs
```diff
@@ -274,56 +324,92 @@ impl<const CONTEXTS: usize> Aggregator<CONTEXTS> {
...
-                kind,
-                points: [point; 1],
-                tags: final_tags,
+            let metric = self.build_metric(entry);
+            let serialized_metric = match serde_json::to_vec(&metric) {
+                Ok(serialized_metric) => serialized_metric,
+                Err(e) => {
+                    error!("failed to serialize metric: {:?}", e);
+                    continue;
+                }
```
duncanista, line 396: Can this code be changed to another pattern? Something like an `if let`? It just looks very strange

duncanista (approved):
LGTM!

bottlecap/src/metrics/aggregator.rs
alexgallotta: can you give me an example?
I am aware of if let with a Some type, but with Result and using the error I don't know whether it is possible

duncanista: Maybe an `unwrap_or_else` might work here:
```
let serialized_metric = serde_json::to_vec(&metric).unwrap_or_else(|e| {
    error!("failed to serialize metric: {:?}", e);
    continue;
});
```
?

PR #281, opened 2024-06-17
Move vuln scan action from agent repo.

purple4reina:
In order to be more proactive on vulnerabilities in the serverless extension, run a vulnerability scan nightly on the agent `main` branch.

We will be able to subscribe in slack to this workflow which I will do once we're done testing it.

### DIFF
--- a/.github/workflows/serverless-vuln-scan.yml
+++ b/.github/workflows/serverless-vuln-scan.yml
@@ -1,0 +1,78 @@
+name: "Serverless Vulnerability Scan"
+
+on:
+  pull_request:
+    paths:
+      - 'cmd/serverless/**'
+      - 'cmd/serverless-init/**'
+      - 'pkg/serverless/**'
+      - '.github/workflows/serverless-vuln-scan.yml'
+
+env:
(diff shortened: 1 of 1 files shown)


.github/workflows/serverless-vuln-scan.yml
```diff
@@ -0,0 +1,75 @@
...
+
+env:
+  VERSION: 1  # env var required when building extension
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout datadog-agent repository
+        uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
```
duncanista, line 16: Why are we adding hashes instead of the version?

purple4reina: I copied this from the agent repo. It was changed to hashes for security reasons.

purple4reina:
Also note, this doesn't scan serverless-init or bottlecap. We can always add those later.

DylanLovesCoffee (approved):
🥹

PR #284, opened 2024-06-20
Refactor metricvalue

alexgallotta:
- removing some extra code
- decouple Type from the aggregator
- remove redundant type in entry
- remove wrapper types over tuples for counter/gauge/ddsketches. There is already 3+ structs for them

### DIFF
--- a/bottlecap/src/metrics/aggregator.rs
+++ b/bottlecap/src/metrics/aggregator.rs
@@ -35,7 +35,6 @@ struct Entry {
     id: u64,
     name: Ustr,
     tags: Option<Ustr>,
-    kind: metric::Type,
     metric_value: MetricValue,
 }
 
@@ -105,7 +104,6 @@ impl Entry {
             id,
             name: metric.name,
             tags: metric.tags,
-            kind: metric.kind,
             metric_value,
         }
     }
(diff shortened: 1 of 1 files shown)


astuyve (approved):

bottlecap/src/metrics/aggregator.rs
```diff
@@ -291,10 +220,11 @@ impl<const CONTEXTS: usize> Aggregator<CONTEXTS> {
...
 
     fn build_sketch(&self, now: i64, entry: &Entry) -> Option<Sketch> {
-        if entry.kind != metric::Type::Distribution {
+        if let MetricValue::Distribution(_) = entry.metric_value {
+        } else {
             return None;
-        }
-        let sketch = entry.metric_value.get_sketch();
+        };
+        let sketch = entry.metric_value.get_sketch()?;
```
duncanista, line 227: I don't understand this change, you're not using the value, why change the logic here?

duncanista (approved):

bottlecap/src/metrics/aggregator.rs
duncanista: I think it's due to how logic was modified to instead of checking the kind, now this is "the way" of checking which type of metric it is. Is there a more idiomatic way of doing this?

PR #290, opened 2024-06-25
Alexgallotta/svls 4840 fix metric limit

alexgallotta:
- re-applyed changes for metric limits
- consume the entries when shipping payloads
- creating a vec of payloads, each within the batch entry/size limit
- there is a corner case: when 1 element is bigger than the max bytes size, it is added anyway in a separate batch. It should not be realistic though

### DIFF
--- a/bottlecap/src/metrics/aggregator.rs
+++ b/bottlecap/src/metrics/aggregator.rs
@@ -90,7 +90,7 @@ impl Entry {
     }
 
     /// Return an iterator over key, value pairs
-    fn tag(&self) -> impl Iterator<Item=(Ustr, Ustr)> {
+    fn tag(&self) -> impl Iterator<Item = (Ustr, Ustr)> {
         self.tags.into_iter().filter_map(|tags| {
             let mut split = tags.split(',');
             match (split.next(), split.next()) {
@@ -101,16 +101,20 @@ impl Entry {
     }
 }
 
+#[derive(Debug, Clone)]
+enum Limit {
+    ApiLimit(usize, usize),
+}
+
 #[derive(Clone)]
 // NOTE by construction we know that intervals and contexts do not explore the
 // full space of usize but the type system limits how we can express this today.
 pub struct Aggregator<const CONTEXTS: usize> {
     tags_provider: Arc<provider::Provider>,
     map: hash_table::HashTable<Entry>,
-    max_batch_entries_size_single_metric: usize,
-    max_content_size_bytes_single_metric: usize,
-    max_batch_entries_size_sketch_metric: usize,
-    max_content_size_bytes_sketch_metric: usize,
+    no_limit: Limit,
+    series_limit: Limit,
+    distribution_limit: Limit,
 }
 
 impl<const CONTEXTS: usize> Aggregator<CONTEXTS> {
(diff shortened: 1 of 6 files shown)


bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -233,29 +236,31 @@ mod tests {
...
         max_mem_sketch.insert(256.0);
-        assert_eq!(
+        if let Some(ValueVariant::DDSketch(variant)) =
             aggr.get_value_by_id(constants::MEMORY_SIZE_METRIC.into(), None)
-                .unwrap(),
-            max_mem_sketch
-        );
+        {
+            assert_eq!(variant, max_mem_sketch);
+        }
```
duncanista, line 264: We should panic on else here, or else, if the value doesn't resolve into the expected, it will execute as a false positive.

bottlecap/src/metrics/aggregator.rs
```diff
@@ -1,15 +1,19 @@
...
+use datadog::Metric as DatadogMetric;
+use metric::{Metric, Type};
 use std::sync::Arc;
 
-use crate::tags::provider;
+use crate::metrics::datadog::Series;
 use datadog_protos::metrics::{Dogsketch, Sketch, SketchPayload};
+use ddsketch_agent::DDSketch;
 use hashbrown::hash_table;
+use log::warn;
```
duncanista, line 15: We shouldn't use `log` crate

bottlecap/src/metrics/aggregator.rs
```diff
@@ -1,15 +1,19 @@
...
-use crate::metrics::metric;
-use crate::metrics::{
-    constants, datadog, errors,
-    metric::{Metric, Type},
+use crate::{
+    metrics::{constants, datadog, errors, metric},
+    tags::provider,
 };
+use datadog::Metric as DatadogMetric;
+use metric::{Metric, Type};
```
duncanista, line 8: Double import for `metric`, `datadog`, and `metrics`

bottlecap/src/metrics/aggregator.rs
duncanista, line 1: Really big change 😿

bottlecap/src/metrics/aggregator.rs
alexgallotta: oh damn auto import, good catch, ty!

bottlecap/src/metrics/aggregator.rs
alexgallotta: Not sure what is a double import, but we are using 2 different metrics structs. I changed names and nested everything for clarity

bottlecap/src/metrics/enhanced/lambda.rs
alexgallotta: thanks, missed that!

bottlecap/src/metrics/aggregator.rs
alexgallotta: happy to go through the changes together if it makes it easier!

duncanista (approved):

bottlecap/src/metrics/flusher.rs
```diff
@@ -16,31 +16,30 @@ impl Flusher {
...
-        if !current_points.series.is_empty() {
-            debug!("flushing {} series to datadog", current_points.series.len());
-            match &self.dd_api.ship_series(&current_points).await {
+        let all_series;
+        let all_distributions;
+        {
+            let locked_aggr = &mut self.aggregator.lock().expect("lock poisoned");
+            all_series = locked_aggr.consume_metrics();
+            all_distributions = locked_aggr.consume_distributions();
+        }
```
tobz, line 25: Seems like this should just be:

```suggestion
        let (all_series, all_distributions) = {
            let mut aggregator = self.aggregator.lock().expect("lock poisoned");
            (aggregator.consume_metrics(), aggregator.consume_distributions())
        };
```

alexgallotta: oh thanks, the mut was in the wrong spot, ty!

PR #292, opened 2024-06-27
chore(bottlecap): log failover reason

duncanista:


bottlecap/src/config/mod.rs
```diff
@@ -77,7 +77,10 @@ pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
         .merge(Env::prefixed("DD_"));
 
     let config = figment.extract().map_err(|err| match err.kind {
-        figment::error::Kind::UnknownField(field, _) => ConfigError::UnsupportedField(field),
+        figment::error::Kind::UnknownField(field, _) => {
+            println!("{{\"reason\":\"{field}\"}}");
```
duncanista, line 81: Should we have a different name for this? Instead of `reason` maybe something more verbose like `DD_LAMBDA_FAILOVER_REASON`?

At the same, it shouldn't matter that much, since this will be received as an `Extension` log in the Go Agent, so do we know which type of logs we have.

astuyve: maybe `DD_EXTENSION_FAILOVER_REASON`? not sure it matters, we'll know what this is.

astuyve (approved):

PR #294, opened 2024-06-27
Add APM tracing support

astuyve:
- Adds support for APM tracing via v4 and v5 endpoints, sending data as v7 to the backend

- mostly a copy paste of the mini agent, but a bunch of stuff changed as we can't flush on a schedule.


bottlecap/Cargo.toml
```diff
@@ -5,12 +5,20 @@ edition = "2021"
 publish = false
 
 [dependencies]
+async-trait = "0.1.64"
+anyhow = "1.0"
```
duncanista, line 9: can we have them with `default-features = false`?

scripts/Dockerfile.bottlecap.build
```diff
@@ -16,7 +17,7 @@ COPY ./bottlecap/Cargo.toml /tmp/dd/bottlecap/Cargo.toml
 COPY ./bottlecap/Cargo.lock /tmp/dd/bottlecap/Cargo.lock
 ENV RUSTFLAGS="-C panic=abort -Zlocation-detail=none"
 WORKDIR /tmp/dd/bottlecap
-RUN --mount=type=cache,target=/usr/local/cargo/registry cargo +nightly build -Z build-std=std,panic_abort -Z build-std-features=panic_immediate_abort --release --target $PLATFORM-unknown-linux-gnu
+RUN --mount=type=cache,target=/usr/local/cargo/registry cargo +nightly build --release --target $PLATFORM-unknown-linux-gnu
```
duncanista, line 20: Why was this required?

bottlecap/src/traces/trace_processor.rs
```diff
@@ -0,0 +1,329 @@
...
+impl TraceProcessor for ServerlessTraceProcessor {
+    async fn process_traces(
+        &self,
+        config: Arc<config::Config>,
+        req: Request<Body>,
+        tx: Sender<trace_utils::SendData>,
+        tags_provider: Arc<provider::Provider>,
+        version: ApiVersion,
+    ) -> http::Result<Response<Body>> {
+        info!("Recieved traces to process");
```
duncanista, line 54:
```suggestion
        debug!("Received traces to process");
```

Maybe make it a debug? Might need an import

bottlecap/src/traces/http_utils.rs
```diff
@@ -0,0 +1,162 @@
...
+/// Response body format:
+/// {
+///     "message": message
+/// }
+pub fn log_and_create_http_response(
+    message: &str,
+    status: StatusCode,
+) -> http::Result<Response<Body>> {
+    if status.is_success() {
+        info!("{message}");
```
duncanista, line 27:
```suggestion
        debug!("{message}");
```

Maybe we don't want to info this?, for the error I think that's fine, but not for the success, WDYT?

bottlecap/src/traces/http_utils.rs
```diff
@@ -0,0 +1,162 @@
...
+/// Takes a request's header map, and verifies that the "content-length" header is present, valid,
+/// and less than the given `max_content_length`.
+///
+/// Will return None if no issues are found. Otherwise logs an error (with the given prefix) and
+/// returns and HTTP Response with the appropriate error status code.
+pub fn verify_request_content_length(
+    header_map: &HeaderMap,
+    max_content_length: usize,
+    error_message_prefix: &str,
+) -> Option<http::Result<Response<Body>>> {
```
duncanista, line 44: Should this be a Result instead of an Option?

bottlecap/src/traces/trace_processor.rs
```diff
@@ -0,0 +1,329 @@
...
+        let (parts, body) = req.into_parts();
+
+        if let Some(response) = http_utils::verify_request_content_length(
+            &parts.headers,
+            MAX_CONTENT_LENGTH,
+            "Error processing traces",
+        ) {
+            return response;
+        }
```
duncanista, line 63: Reading this made me think if it would be better to change the return from the function into a Result<(), Err>, or else rename this `response` to something negative, since it wasn't clear that `verify_request_content_length` was supposed to return `Some(response)` which in turn is used for a negative behavior.

bottlecap/src/traces/trace_processor.rs
```diff
@@ -0,0 +1,329 @@
...
+
+        let payload = trace_utils::collect_trace_chunks(
+            traces,
+            &tracer_header_tags,
+            |chunk, _root_span_index| {
+                // trace_utils::set_serverless_root_span_tags(
+                //     &mut chunk.spans[root_span_index],
+                //     config.function_name.clone(),
+                //     &config.env_type,
+                // );
```
duncanista, line 98: Should we remove this or keep it?

bottlecap/src/traces/trace_processor.rs
```diff
@@ -0,0 +1,329 @@
...
+                chunk.spans.retain(|span| {
+                    (span.resource != "127.0.0.1" || span.resource != "0.0.0.0")
+                        && span.name != "dns.lookup"
+                });
```
duncanista, line 102: Might be good to refactor it into a function which specifies which spans we don't want?

bottlecap/src/traces/trace_flusher.rs
```diff
@@ -0,0 +1,64 @@
...
+            self.flush_traces(buffer.to_vec()).await;
+            buffer.clear();
+        }
+    }
+
+    async fn flush_traces(&self, traces: Vec<SendData>) {
+        if traces.is_empty() {
+            return;
+        }
+        info!("Flushing {} traces", traces.len());
```
duncanista, line 52:
```suggestion
        debug!("Flushing {} traces", traces.len());
```

Might need an import

bottlecap/src/traces/trace_flusher.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        for traces in trace_utils::coalesce_send_data(traces) {
+            match traces.send().await.last_result {
+                Ok(_) => info!("Successfully flushed traces"),
```
duncanista, line 56:
```suggestion
                Ok(_) => debug!("Successfully flushed traces"),
```

bottlecap/src/traces/trace_agent.rs
```diff
@@ -0,0 +1,203 @@
...
+
+        let make_svc = make_service_fn(move |_| {
+            let trace_processor = trace_processor.clone();
+            let trace_tx = trace_tx.clone();
+
+            let stats_processor = stats_processor.clone();
+            let stats_tx = stats_tx.clone();
+
+            let endpoint_config = endpoint_config.clone();
+            let tags_provider = tags_provider.clone();
```
duncanista, line 92: Rust Q, sooo many clones, can't we just move them all the way down? 🤔

bottlecap/src/traces/stats_processor.rs
```diff
@@ -0,0 +1,94 @@
...
+        info!("Recieved trace stats to process");
+        let (parts, body) = req.into_parts();
+
+        if let Some(response) = http_utils::verify_request_content_length(
+            &parts.headers,
+            MAX_CONTENT_LENGTH,
+            "Error processing trace stats",
+        ) {
+            return response;
+        }
```
duncanista, line 48: Same comment that I had for the other file

bottlecap/src/config/mod.rs
```diff
@@ -33,6 +34,13 @@ pub struct Config {
     pub apm_enabled: bool,
     pub lambda_handler: String,
     pub serverless_flush_strategy: FlushStrategy,
+    pub trace_enabled: bool,
+    pub serverless_trace_enabled: bool,
+    pub capture_lambda_payload: bool,
```
duncanista, line 39: I was gonna say should we remove this? Then I remembered that we are currently targeting Node + Python

duncanista (commented):
Left some comments – impressive work as always! 🔥

bottlecap/src/traces/http_utils.rs
astuyve: It can be, this is from the [mini agent](https://github.com/DataDog/libdatadog/blob/main/trace-mini-agent/src/http_utils.rs). We can tweak it though, just might make it more challenging to re-merge later if we go that route.

scripts/Dockerfile.bottlecap.build
astuyve: it's not, I'll revert

bottlecap/src/traces/trace_processor.rs
astuyve: Good question, as far as I can tell we need these tags on every span, not just the root span. We can discuss this on slack a bit.

bottlecap/src/traces/trace_processor.rs
astuyve: can do that if you'd rather that then a closure

bottlecap/src/traces/trace_processor.rs
astuyve: We can, but we'd need to do it in the [mini agent](https://github.com/DataDog/libdatadog/blob/main/trace-mini-agent/src/http_utils.rs)

bottlecap/src/traces/trace_agent.rs
astuyve: nah, need to clone into make_service_fn and then again into service_fn. This all changed in hyper 1.x, so maybe we can look forward to fixing it then?

bottlecap/src/traces/trace_processor.rs
duncanista: Maybe not now, but I wonder if in the future we will need more spans that we want to ignore

bottlecap/src/traces/trace_agent.rs
duncanista: Sg!

duncanista (approved):
LGTM – :shipit:

PR #299, opened 2024-07-03
fix: Use resolved API key

astuyve:
My mistake, I thought the API key would be updated on the config.

Instead we just need to pass the resolved key.

### DIFF
--- a/bottlecap/src/traces/stats_flusher.rs
+++ b/bottlecap/src/traces/stats_flusher.rs
@@ -29,6 +29,7 @@ pub trait StatsFlusher {
 pub struct ServerlessStatsFlusher {
     pub buffer: Arc<Mutex<Vec<pb::ClientStatsPayload>>>,
     pub config: Arc<config::Config>,
+    pub resolved_api_key: String,
 }
 
 #[async_trait]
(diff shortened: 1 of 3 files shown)


bottlecap/src/traces/stats_flusher.rs
```diff
@@ -29,6 +29,7 @@ pub trait StatsFlusher {
 pub struct ServerlessStatsFlusher {
     pub buffer: Arc<Mutex<Vec<pb::ClientStatsPayload>>>,
     pub config: Arc<config::Config>,
+    pub resolved_api_key: String,
```
duncanista, line 32: Shouldn't we just name this field `api_key` and then pass the resolved one?

duncanista (approved):
LGTM – I'd prefer renaming the field to `api_key` instead of `resolved_api_key`, but up to you!

bottlecap/src/traces/stats_flusher.rs
astuyve: The resolved_api_key is the name we used elsewhere. config also has an API key field which I think makes it a little confusing, so I chose this name.

We are already passing the resolved one, see the changes in main.

PR #302, opened 2024-07-03
feat: Require DD_EXTENSION_VERSION: next

astuyve:
1. Requires `DD_EXTENSION_VERSION: next` to be set, otherwise we fail over to the main extension.
2. Fixes an issue where `DD_LOG_LEVEL` was case sensitive.

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -42,6 +42,7 @@ pub struct Config {
     pub logs_injection: bool,
     pub merge_xray_traces: bool,
     pub serverless_appsec_enabled: bool,
+    pub extension_version: Option<String>,
 }
 
 impl Default for Config {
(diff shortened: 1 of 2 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -109,6 +111,16 @@ pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
         return Err(ConfigError::UnsupportedField("appsec_enabled".to_string()));
     }
 
+    match config.extension_version.as_deref() {
+        Some("next") => {}
+        Some(&_) | None => {
```
alexgallotta, line 116: why &_ rather than _?

alexgallotta (approved):

bottlecap/src/config/mod.rs
astuyve: I originally had it as a separate branch and had some logic, but removed it. Will replace!

duncanista (approved):

PR #306, opened 2024-07-08
feat: Remove info, use debug

astuyve:
Replace info! with debug! and use tracing lib instead of logging.

### DIFF
--- a/bottlecap/src/traces/trace_agent.rs
+++ b/bottlecap/src/traces/trace_agent.rs
@@ -9,7 +9,7 @@ use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Instant;
 use tokio::sync::mpsc::{self, Receiver, Sender};
-use tracing::{error, info};
+use tracing::{error, debug};
 
 use crate::config;
 use crate::tags::provider;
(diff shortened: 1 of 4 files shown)


bottlecap/src/traces/trace_agent.rs
```diff
@@ -112,8 +112,8 @@ impl TraceAgent {
 
         let server = server_builder.serve(make_svc);
 
-        info!("Trace Agent started: listening on port {TRACE_AGENT_PORT}");
-        info!(
+        debug!("Trace Agent started: listening on port {TRACE_AGENT_PORT}");
```
alexgallotta, line 115: Should we collapse the 2 debugs in one line?

alexgallotta (approved):

bottlecap/src/traces/trace_agent.rs
astuyve: eh, I like the newline between them when I check the logs. It's nice to see if the port got goofed or if the duration took longer than expected. It's just personal preference on my end though, if you think it should be changed I have no problem with that. Happy to slap a `\n` in here and call it good!

duncanista (approved):

PR #307, opened 2024-07-08
feat: honor enhanced metrics bool

astuyve:
Some people will turn this off but still want logs, this change allows that.

### DIFF
--- a/bottlecap/src/metrics/enhanced/lambda.rs
+++ b/bottlecap/src/metrics/enhanced/lambda.rs
@@ -21,15 +21,15 @@ impl Lambda {
     }
 
     pub fn increment_invocation_metric(&self) {
-        self.increment_metric(constants::INVOCATIONS_METRIC)
+        self.increment_metric(constants::INVOCATIONS_METRIC);
     }
 
     pub fn increment_errors_metric(&self) {
-        self.increment_metric(constants::ERRORS_METRIC)
+        self.increment_metric(constants::ERRORS_METRIC);
     }
 
     pub fn increment_timeout_metric(&self) {
-        self.increment_metric(constants::TIMEOUTS_METRIC)
+        self.increment_metric(constants::TIMEOUTS_METRIC);
     }
 
     pub fn set_init_duration_metric(&self, init_duration_ms: f64) {
--- a/bottlecap/src/metrics/enhanced/lambda.rs
+++ b/bottlecap/src/metrics/enhanced/lambda.rs
@@ -8,12 +8,16 @@ use tracing::error;
 
 pub struct Lambda {
     pub aggregator: Arc<Mutex<Aggregator<1024>>>,
+    pub config: Arc<crate::config::Config>,
 }
 
 impl Lambda {
     #[must_use]
-    pub fn new(aggregator: Arc<Mutex<Aggregator<1024>>>) -> Lambda {
-        Lambda { aggregator }
+    pub fn new(
+        aggregator: Arc<Mutex<Aggregator<1024>>>,
+        config: Arc<crate::config::Config>,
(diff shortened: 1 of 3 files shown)


bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -29,6 +33,9 @@ impl Lambda {
     }
 
     pub fn set_init_duration_metric(&self, init_duration_ms: f64) -> Result<(), errors::Insert> {
+        if !self.config.enhanced_metrics {
```
alexgallotta, line 36: Should we use Option<Error> instead of result with unit?

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -55,6 +65,9 @@ impl Lambda {
     }
 
     pub fn set_runtime_duration_metric(&self, duration_ms: f64) {
+        if !self.config.enhanced_metrics {
+            return;
```
alexgallotta, line 69: Why those methods are not returning anything, unlike the others above?

duncanista (approved):

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -8,12 +8,16 @@ use tracing::error;
 
 pub struct Lambda {
     pub aggregator: Arc<Mutex<Aggregator<1024>>>,
+    pub config: Arc<crate::config::Config>,
```
alexgallotta, line 11: do we really need to pass this config around?
It seems it is only used for `self.config.enhanced_metrics`.
I would either just pass that single boolean or, even better, have 2 Lambda Impl:
- EnhancedLambda when the boolean is true
- NoEnahcnedLambda otherwise
so we save all those ifs around the code too

astuyve: if we have no enhanced lambda then we have to duplicate every method in an entire module, that seems overly complicated because we'd need to always ensure both modules have the same methods and are tested.

This way at least it's clear they become no-ops.

We could pass just the boolean though, but it's an ARC for a slice of memory that always lives for the lifetime of our application. Passing the boolean will actually allocate more memory (albeit one additional bit, maybe 2 for the pointer)

alexgallotta: Good point, I have the feeling there is a way to remove the extra methods/ifs and handle the enhanced metrics, but maybe it will be clearer later on

astuyve: Yeah I agree, but I only had a minute or two to write this code and didn't think of it. Honestly this is pretty pragmatic already, and prevents us from wrapping the rest of the usages with conditionals.

bottlecap/src/metrics/enhanced/lambda.rs
astuyve: not sure what you mean by result with unit?

bottlecap/src/metrics/enhanced/lambda.rs
astuyve: I agree we should standardize. I think removing the result is the most clean way, because we don't retry/reattempt incrementing metrics if it fails, right?

astuyve: Done

alexgallotta (approved):

bottlecap/src/metrics/enhanced/lambda.rs
alexgallotta: Unit type is (). Since it doesn't hold anything, in this case Some(Err)/None can be used

PR #310, opened 2024-07-09
add mock for integration test for metrics

alexgallotta:
- add integration test for metrics and logs
- removed aggregator generic constant (discussed on slack https://dd.slack.com/archives/C6YASP83Z/p1719514829557779 )


bottlecap/src/logs/flusher.rs
```diff
@@ -1,53 +1,69 @@
...
 
 pub struct Flusher {
     api_key: String,
-    site: String,
+    fqdn_site: String,
     client: reqwest::Client,
     aggregator: Arc<Mutex<Aggregator>>,
 }
 
+#[must_use]
```
astuyve, line 14: Could be `#[inline]` as well

bottlecap/src/logs/flusher.rs
```diff
@@ -1,53 +1,69 @@
...
         Flusher {
             api_key,
-            site,
+            fqdn_site: site,
             client,
             aggregator,
         }
     }
-    pub async fn flush(&self) {
+    pub async fn flush(&self) -> HashMap<String, Result<(), String>> {
```
astuyve, line 30: I don't think we should return anything from this method, especially not a hashmap for results? We're doing nothing with them.

bottlecap/tests/metrics_integration_test.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::sync::{Arc, Mutex};
+
+mod common;
+
+#[tokio::test]
+async fn test_enhanced_metrics() {
+    let dd_api_key = "my_test_key";
+
+    // payload looks like
+    // aws.lambda.enhanced.invocations"_dd.compute_stats:1"architecture:x86_64"function_arn:test-arn:�൴      �?!      �?)      �?1      �?:�B
```
astuyve, line 18: got some non unicode symbols here

astuyve (changes requested):
Tests look good, let's remove the changes to application code though. We don't need to do these allocations.

bottlecap/src/logs/flusher.rs
alexgallotta: It is just to verify that the request was executed correctly. Right now it is used only on tests, but I guess it will be needed when we do the retry

bottlecap/tests/metrics_integration_test.rs
alexgallotta: I didn't want to spend too much time trying to output the raw string there. I am matching the readable parts only

bottlecap/src/logs/flusher.rs
alexgallotta: good idea!

PR #311, opened 2024-07-09
aj/add hello route

astuyve:
the `hello` route was a path used to tell the tracers that the agent is ready to receive data. It goes back to the [original PR](https://github.com/DataDog/datadog-agent/pull/6591) for the extension.

It was eventually [removed](https://github.com/DataDog/datadog-agent/pull/9782) but somehow lives on.

We've since removed it in all lambda layers, but a smoke test revealed that a user was still using it as late as layer v85.

We should remove this entirely at some point, but that would effectively be a breaking change for all versions of the clients/tracers.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -37,11 +37,11 @@ use bottlecap::{
         listener::TelemetryListener,
     },
     traces::{
+        hello_agent,
         stats_flusher::{self, StatsFlusher},
         stats_processor, trace_agent,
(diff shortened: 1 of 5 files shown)


duncanista (commented):
Are there any cold start implications by starting another server?

astuyve:
I doubt it, but it doesn't matter. We have to support it until we fully deprecate it. I'd love if we could do that today, but an alpha user already ran into this so I don't think it's quite time to deprecate it on the server-side.

duncanista (approved):

PR #317, opened 2024-07-12
feat: multi-metric support

astuyve:
TODO:
- add test for multi-line metric
- Add test for single metric
- Talk with George on SMP

### DIFF
--- a/bottlecap/src/metrics/dogstatsd.rs
+++ b/bottlecap/src/metrics/dogstatsd.rs
@@ -1,8 +1,6 @@
-use tokio::sync::mpsc::Sender;
-
+use std::str::Split;
 use tracing::{debug, error};
 
-use crate::events::{self, Event, MetricEvent};
 use crate::metrics::aggregator::Aggregator;
 use crate::metrics::metric::Metric;
 use std::sync::{Arc, Mutex};
@@ -11,7 +9,6 @@ pub struct DogStatsD {
     cancel_token: tokio_util::sync::CancellationToken,
     aggregator: Arc<Mutex<Aggregator>>,
     socket: tokio::net::UdpSocket,
-    event_bus: Sender<events::Event>,
 }
 
 pub struct DogStatsDConfig {
@@ -24,7 +21,6 @@ impl DogStatsD {
     pub async fn new(
         config: &DogStatsDConfig,
         aggregator: Arc<Mutex<Aggregator>>,
-        event_bus: Sender<events::Event>,
         cancel_token: tokio_util::sync::CancellationToken,
     ) -> DogStatsD {
         let addr = format!("{}:{}", config.host, config.port);
@@ -36,12 +32,12 @@ impl DogStatsD {
             cancel_token,
             aggregator,
             socket,
-            event_bus,
         }
     }
 
     pub async fn spin(self) {
-        loop {
+        let mut spin_cancelled = false;
+        while !spin_cancelled {
             // TODO(astuyve) this should be dynamic
             let mut buf = [0; 1024]; // todo, do we want to make this dynamic? (not sure)
             let (amt, src) = self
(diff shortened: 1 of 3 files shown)


bottlecap/src/metrics/dogstatsd.rs
```diff
@@ -35,63 +56,115 @@ impl DogStatsD {
...
-                Ok(parsed_metric) => {
-                    debug!("parsed metric: {:?}", parsed_metric);
-                    parsed_metric
-                }
+        let mut spin_cancelled = false;
+        while !spin_cancelled {
+            self.consume_statsd().await;
+            spin_cancelled = self.cancel_token.is_cancelled();
+        }
+    }
```
duncanista, line 69: Can we use a tokio select here instead? with a bias maybe?

bottlecap/src/metrics/dogstatsd.rs
```diff
@@ -35,63 +56,115 @@ impl DogStatsD {
...
+    async fn consume_statsd(&self) {
+        let (buf, src) = self
+            .buffer_reader
+            .read()
+            .await
+            .expect("didn't receive data");
+        let msgs = std::str::from_utf8(&buf).expect("couldn't parse as string");
+        debug!(
+            "received message: {} from {}, sending it to the bus",
+            msgs, src
```
duncanista, line 80: We are not sending anything to a bus anymore tho, maybe update/remove this debug log?

bottlecap/src/metrics/dogstatsd.rs
```diff
@@ -35,63 +56,115 @@ impl DogStatsD {
...
+    use crate::metrics::aggregator::ValueVariant::Value;
+    use crate::metrics::dogstatsd::{BufferReader, DogStatsD};
+    use crate::tags::provider::Provider;
+    use crate::LAMBDA_RUNTIME_SLUG;
+    use std::collections::HashMap;
+    use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+    use std::sync::{Arc, Mutex};
+
+    #[tokio::test]
+    async fn test_dogstatsd() {
```
duncanista, line 118: Would it make sense to also add a single metric test? here we are testing multiple metrics at once, WDYT?

bottlecap/src/metrics/dogstatsd.rs
```diff
@@ -35,63 +56,115 @@ impl DogStatsD {
...
+    }
+
+    fn insert_metrics(&self, msg: Split<char>) {
+        let all_valid_metrics: Vec<Metric> = msg
+            .filter_map(|m| match Metric::parse(m) {
+                Ok(metric) => Some(metric),
                 Err(e) => {
-                    error!("failed to parse metric: {:?}\n message: {:?}", msg, e);
-                    continue;
+                    error!("failed to parse metric: {}", e);
```
duncanista, line 91: Should we also add the message in the error? so we know which metric we failed to parse and from where?

duncanista (approved):
LGTM – left some nits, might be good to release a beta version for SMP!

bottlecap/src/metrics/dogstatsd.rs
alexgallotta: Is it worth it?
I don't like to add extra macros and stuff if it's not really helpful.
Also, the reason why the flag is there, it's to have a do-while. We probably want to consume at least once before checking if it's cancelled, in case of quick shut down

bottlecap/src/metrics/dogstatsd.rs
alexgallotta: I will correct the message. It's still useful for debugging for the source addresser

bottlecap/src/metrics/dogstatsd.rs
alexgallotta: good idea!

bottlecap/src/metrics/dogstatsd.rs
alexgallotta: I think the string parsing should be tested as unit test in the metric parser.
I will add a 1 metric example just to make sure the newlines won't play fun

PR #318, opened 2024-07-12
chore(bottlecap): switch flushing strategy to race

astuyve:
WIP (lol had a baby)

- Pick a better time value. I'd suggest 1-3 seconds
- override with a custom-scheduled `periodic` value
- skip if the user specifies `end`
- drop the `smart flushing` algo?

# What?

Allows flushing strategies to be expanded, allowing to periodically flush during runtime, improving experience for long running processes.

# How?

Using `DD_SERVERLESS_FLUSH_STRATEGY`:
1. Not set counts as default, this strategy flushes periodically every second, and at the end of the invocation.
2. `end`, flushing only at the end of the invocation.
3. `end,<ms>` flush periodically every defined `<ms>`, and at the end of the invocation. As the default, but with more control.
4. `periodically,<ms>` flush only periodically every defined `<ms>`.

# Note

- The adapting algorithm after 20 invocations gets removed.
- Not removing AJs PR description bc of historical reasons! 🍼


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -331,10 +328,19 @@ async fn extension_loop_active(
...
 
+    // TODO(duncanista): move this into a separate module
+    let mut periodic_flush_timer = match config.serverless_flush_strategy {
+        FlushStrategy::Periodically(period) => {
+            tokio::time::interval(tokio::time::Duration::from_millis(period.interval))
+        }
+        FlushStrategy::Default => tokio::time::interval(tokio::time::Duration::from_millis(1000)),
+        FlushStrategy::End => tokio::time::interval(tokio::time::Duration::MAX),
+    };
+    periodic_flush_timer.tick().await; // discard first tick, which is instantaneous
```
duncanista, line 342: Let's definitely move this directly into a module when this strategy matures

scripts/Dockerfile.bottlecap.build
```diff
@@ -19,7 +19,7 @@ RUN mkdir -p /tmp/dd
 COPY ./bottlecap/src /tmp/dd/bottlecap/src
 COPY ./bottlecap/Cargo.toml /tmp/dd/bottlecap/Cargo.toml
 COPY ./bottlecap/Cargo.lock /tmp/dd/bottlecap/Cargo.lock
-ENV RUSTFLAGS="-C panic=abort -Zlocation-detail=none"
+ENV RUSTFLAGS="-C panic=abort -Zlocation-detail=none --cfg tokio_unstable"
```
duncanista, line 22: Removing this for prod, we should use it to keep collecting metrics in the future dev

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -222,8 +221,11 @@ fn enable_logging_subsystem(config: &Arc<Config>) {
 
     let subscriber = tracing_subscriber::fmt::Subscriber::builder()
         .with_env_filter(
-            EnvFilter::try_new(config.log_level)
-                .expect("could not parse log level in configuration"),
+            EnvFilter::try_new(format!(
+                "{:?},hyper=warn,reqwest=warn,rustls=warn,datadog-trace-mini-agent=warn",
+                config.log_level
+            ))
+            .expect("could not parse log level in configuration"),
```
duncanista, line 228: If customers sets log level to `debug`, some crates might spam tons of logs, this allows us to set specific logging level per crate

duncanista: Solved in #329

bottlecap/src/config/flush_strategy.rs
```diff
@@ -26,28 +27,39 @@ impl<'de> Deserialize<'de> for FlushStrategy {
...
-                            return Ok(FlushStrategy::Periodically(PeriodicStrategy {
-                                interval: parsed_interval,
-                            }));
-                        }
-                        debug!("Invalid flush interval: {}, using default", interval);
-                        Ok(FlushStrategy::Default)
-                    } else {
-                        debug!("Invalid flush strategy: {}, using default", value);
-                        Ok(FlushStrategy::Default)
+            // "end,1000"
```
alexgallotta, line 30: this seems a great use cases for match-pattern

bottlecap/src/lifecycle/flush_control.rs
```diff
@@ -1,14 +1,11 @@
 use crate::config::flush_strategy::FlushStrategy;
-use crate::lifecycle::invocation_times::InvocationTimes;
-use ::std::time;
-use tracing::debug;
+use tokio::time::Interval;
 
-const TWENTY_SECONDS: u64 = 20 * 1000;
+const DEFAULT_FLUSH_INTERVAL: u64 = 1000; // 1s
 
+#[derive(Clone, Copy, Debug, PartialEq)]
```
alexgallotta, line 6: Do we need all of that?
Maybe with an enum rather than struct it could be easier, instead of wrapping the enum on a struct. Also there won't be need of all those ifs later in the logic

bottlecap/src/config/flush_strategy.rs
duncanista: I was thinking that at the beginning, but then I have to duplicate logic for interval parsing in both places + all the debugs when it fails, wdyt?

bottlecap/src/lifecycle/flush_control.rs
duncanista: Clone + Copy the compiler was complaining about, debug and partial eq for debugging

duncanista: I also suspect we might need to pass timeout racing + signals for buffers, that's why I didn't remove the structure yet

alexgallotta: an enum should be able to get arguments, already implemented those traits and be enough for this use case

bottlecap/src/config/flush_strategy.rs
alexgallotta: Mh not sure,, the mapping can be done with
match
    end,periodic or periodic,end => do same stuff

right?

duncanista: Addressed it!

bottlecap/src/lifecycle/flush_control.rs
duncanista: If I do it, then initialization would end up with something like
```rust
let flush_control = match config.serverless_flush_strategy {
    FlushStrategy::Default => FlushControl::Default,
    FlushStrategy::Periodically(p) => FlushControl::Periodically { interval: p.interval },
    FlushStrategy::EndPeriodically(p) => FlushControl::EndPeriodically { interval: p.interval },
    FlushStrategy::End => FlushControl::End,
};```

duncanista: Will refactor in another PR

alexgallotta (approved):

PR #326, opened 2024-07-18
chore(bottlecap): fallback on `datadog.yaml` usage

duncanista:
# What?

Failovers to Go binary when `datadog.yaml` is present.

# Motivation

`figment` crate doesn't match properly fields from `YAML` and environment variables. Therefore, if users had some configuration values enabled, we might not be able to honor them properly, due to how they're processed by figment.

# Note

This should be reverted once configuration is done accordingly for environment variables and `YAML`

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -4,11 +4,9 @@ pub mod processing_rule;
 
 use std::path::Path;
 
-use figment::{
-    providers::{Env, Format, Yaml},
-    Figment,
-};
+use figment::{providers::Env, Figment};
 use serde::Deserialize;
+use tracing::debug;
 
 use crate::config::flush_strategy::FlushStrategy;
 use crate::config::log_level::LogLevel;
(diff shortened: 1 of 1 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -108,6 +106,13 @@ pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
         _ => ConfigError::ParseError(err.to_string()),
     })?;
 
+    // TODO(duncanista): revert to using datadog.yaml when we have a proper serializer
+    if path.exists() {
+        log_failover_reason("datadog_yaml");
+        debug!("datadog.yaml is not supported, use environment variables instead");
+        return Err(ConfigError::UnsupportedField("datadog_yaml".to_string()));
```
alexgallotta, line 113: . instead of _?

duncanista: This is going to be sent as a metric, so the tag of the metric I think cannot contain `.`

alexgallotta: oh ok sounds good

alexgallotta (approved):

PR #327, opened 2024-07-18
feat(bottlecap): add `runtime` tag

alexgallotta:


bottlecap/src/tags/lambda/tags.rs
```diff
@@ -113,6 +116,16 @@ fn tags_from_env(
     if let Ok(memory_size) = std::env::var(MEMORY_SIZE_VAR) {
         tags_map.insert(MEMORY_SIZE_KEY.to_string(), memory_size);
     }
+    if let Ok(exec_runtime_env) = std::env::var(RUNTIME_VAR) {
+        // AWS_Lambda_java8
```
astuyve, line 120: if this takes ~25ms can we add an opt-out for people who may know what their runtime is and not care?

alexgallotta: it seems the 25ms delay is coded in goagent (more details in the jira https://datadoghq.atlassian.net/browse/SVLS-4904 ) so we are not adding anything here

alexgallotta:
It seems serialized-size byte tests are failing both here and in main.
I am not sure if it depends on my arch building protobuf or some other changes. If anyone else can replicate the green test or the failing it would be extremely helpful

bottlecap/src/tags/lambda/tags.rs
```diff
@@ -1,13 +1,16 @@
+use crate::config;
+use log::warn;
```
duncanista, line 2: don't use `log` crate

bottlecap/src/tags/lambda/tags.rs
```diff
@@ -132,6 +145,41 @@ fn tags_from_env(
     tags_map
 }
 
+fn resolve_provided_runtime(path: &str) -> String {
+    let path = Path::new(path);
+
+    let file = match File::open(path) {
+        Err(why) => {
+            warn!(
```
duncanista, line 153: Maybe a debug? I'm afraid this might spam users

bottlecap/src/tags/lambda/tags.rs
```diff
@@ -167,12 +215,15 @@ impl Lambda {
 mod tests {
     use super::*;
     use crate::config::Config;
+    use serial_test::serial;
+    use std::io::Write;
 
     #[test]
+    // #[serial]
```
duncanista, line 222: ?

duncanista (approved):
LGTM – left some comments!

duncanista:
Tests are failing because a new tag adds size to the metric!

bottlecap/src/tags/lambda/tags.rs
alexgallotta: ops missed this, thanks

bottlecap/src/tags/lambda/tags.rs
alexgallotta: ty! My IDE doesn't do the right thing whenever I add new loggers

PR #328, opened 2024-07-18
feat(bottlecap): add `dd_extension_version` to tags

duncanista:
# What?

Adds `dd_extension_version` to the tags.

# Motivation

We need it.
Also [SVLS-5154](https://datadoghq.atlassian.net/browse/SVLS-5154)

# Note

Currently, we are setting a hardcoded value `61-next` which should indicate the current version of the package. This needs to be updated to not be hardcoded. Keeping it only for closed beta.

[SVLS-5154]: https://datadoghq.atlassian.net/browse/SVLS-5154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/metrics/aggregator.rs
+++ b/bottlecap/src/metrics/aggregator.rs
@@ -429,6 +429,8 @@ mod tests {
     use std::collections::hash_map;
     use std::sync::Arc;
 
+    const SINGLE_METRIC_SIZE: usize = 205;
+
     fn create_tags_provider() -> Arc<provider::Provider> {
         let config = Arc::new(config::Config::default());
         Arc::new(provider::Provider::new(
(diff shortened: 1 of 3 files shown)


DylanLovesCoffee (approved):

bottlecap/src/metrics/aggregator.rs
```diff
@@ -604,8 +604,8 @@ mod tests {
 
     #[test]
     fn consume_distributions_batch_bytes() {
-        let single_proto_size = 104;
-        let max_bytes = 250;
+        let single_proto_size = 135;
+        let max_bytes = 300;
```
alexgallotta, line 608: max_bytes should not be changed, unless it moves the threshold enough to prevent expected batching

duncanista: This was needed because or else, batches would mismatch, since the size change, 250 was max 2 batches of 104, now that the number is 135, we need at least 270 for proper batching

PR #329, opened 2024-07-19
fix(bottlecap): filter debug logs from external crates

duncanista:
# What?

External crates, `h2` and `hyper` specifically, were adding logs to our result whenever the log level was set to debug. This is not ideal because we don't want other crates interfering with our logging strategy.

# How?

Updated the env filter to disable some crates. 

# Motivation
We don't want to spam customer with these obscure debug logs.
Also [SVLS-4914](https://datadoghq.atlassian.net/browse/SVLS-4914)

# Note
- Removed unnecessary files
- Updated some messages to be capitalized

[SVLS-4914]: https://datadoghq.atlassian.net/browse/SVLS-4914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bottlecap/src/lifecycle/flush_control.rs
```diff
@@ -37,7 +37,7 @@ impl FlushControl {
         let now = match time::SystemTime::now().duration_since(time::UNIX_EPOCH) {
             Ok(now) => now.as_secs(),
             Err(e) => {
-                debug!("failed to get current time: {:?}", e);
+                debug!("Failed to get current time: {:?}", e);
```
alexgallotta, line 40: I think this should be very unlikely, should be warn/err?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -344,7 +338,7 @@ async fn extension_loop_active(
                 invoked_function_arn,
             }) => {
                 debug!(
-                    "[extension_next] Invoke event {}; deadline: {}, invoked_function_arn: {}",
+                    "Invoke event {}; deadline: {}, invoked_function_arn: {}",
```
alexgallotta, line 343: Why is `extension_next` only here?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -213,16 +213,10 @@ fn load_configs() -> (AwsConfig, Arc<Config>) {
...
 
 fn enable_logging_subsystem(config: &Arc<Config>) {
-    // Bridge any `log` logs into the tracing subsystem. Note this is a global
-    // registration.
-    tracing_log::LogTracer::builder()
-        .with_max_level(config.log_level.as_level_filter())
-        .init()
-        .expect("failed to set up log bridge");
-
+    let env_filter = format!("h2=off,hyper=off,rustls=off,datadog-trace-mini-agent=off,{:?}", config.log_level);
```
alexgallotta: Not a big deal for now, but an option to enable more verbose debugging might come handy. Especially for `datadog-trace-mini-agent`

bottlecap/src/bin/bottlecap/main.rs
duncanista: i think AJ added it in a previous PR, but I just fixed it in the formatter

bottlecap/src/bin/bottlecap/main.rs
duncanista: Although I'd like it, we'd have to upstream fix their info logging, they're logging whenever they send a request, it's not ideal because it shows at the info level

bottlecap/src/lifecycle/flush_control.rs
duncanista: Yeah, maybe an error

duncanista: Actually, this is removed during my racing flushing PR, won't fix

alexgallotta (approved):

PR #333, opened 2024-07-23
fix(bottlecap): remove log that might confuse customers

duncanista:
# What?

When loading Rust, a `println` explicitly marked an `Error` which would end up in customers Lambdas, this might appear as confusing, since we end up going into failover with Go

# Motivation

We should have removed this, but we forgot that this could appear on every customer 🤦🏽.
Also `#incident-29136`

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -6,7 +6,6 @@ use std::path::Path;
 
 use figment::{providers::Env, Figment};
 use serde::Deserialize;
-use tracing::debug;
 
 use crate::config::flush_strategy::FlushStrategy;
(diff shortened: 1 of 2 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -109,7 +108,6 @@ pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
     // TODO(duncanista): revert to using datadog.yaml when we have a proper serializer
     if path.exists() {
         log_failover_reason("datadog_yaml");
-        debug!("datadog.yaml is not supported, use environment variables instead");
```
duncanista, line 112: Also this, because the logging subsystem is not enabled, so this would never appear

Issue #331, opened 2024-07-23
Datadog Lambda Extension v61 causes lambdas to crash on startup


JavaScriptBach:
The latest version of the Datadog Lambda Extension (v61) causes my lambdas to crash on startup with the error `UnsupportedField("extension_version")`.

I am running Node.js lambdas using the AWS Node 20 base container image in x86. I'm guessing https://github.com/DataDog/datadog-lambda-extension/pull/328 is related?

DylanLovesCoffee:
hey @JavaScriptBach - apologies for this. We're working on reproducing and fixing the issue at the moment.

JavaScriptBach:
Thanks! In the meantime, my lambdas are working again on v60.

Some more context in case it helps:

- My Dockerfile base image looks like this:
```
FROM public.ecr.aws/lambda/nodejs:20.2024.05.08.11-x86_64
COPY --from=public.ecr.aws/datadog/lambda-extension:latest /opt/extensions/ /opt/extensions
```

- I'm using the `datadog-lambda-js` library version 7.104.0.
- My lambda's DD_ env variables are `DD_API_KEY_SECRET_ARN`, `DD_ENV`, `DD_SERVICE`, `DD_SITE`, and `DD_VERSION`.
- And my Cloudwatch logs look like this:

```

2024-07-23T01:01:25.388Z{     "DD_EXTENSION_FAILOVER_REASON": "extension_version" } | {"DD_EXTENSION_FAILOVER_REASON":"extension_version"}
-- | --
  | 2024-07-23T01:01:25.396ZError loading configuration: UnsupportedField("extension_version") | Error loading configuration: UnsupportedField("extension_version")
  | 2024-07-23T01:01:25.399ZEXTENSION	Name: datadog-agent	State: Started	Events: [] | EXTENSION Name: datadog-agent State: Started Events: []
  | 2024-07-23T01:01:25.399ZEXTENSION	Name: datadog-agent-go	State: Started	Events: [] | EXTENSION Name: datadog-agent-go State: Started Events: []
  | 2024-07-23T01:01:25.399ZINIT_REPORT Init Duration: 103.79 ms	Phase: init	Status: error	Error Type: Extension.Crash | INIT_REPORT Init Duration: 103.79 ms Phase: init Status: error Error Type: Extension.Crash
[...]

PR #342, opened 2024-07-31
feat: Remove excepts from secrets, drop into idle loop when local

astuyve:
Followup from #331 for local emulators.

There's no telemetry API so we can't process logs or handle the lifecycle, so we'll just idle and call `next` infinitely.

Because of the missing API keys and lack of secrets, this drops us into the `can't decrypt secrets` catch-all. We may consider creating one specifically for local emulation, but my guess is the key would be switching over when we detect there is no telemetry API.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -78,7 +78,7 @@ struct RegisterResponse {
     // body, but as a header. Header is extracted and set manually.
     #[serde(skip_deserializing)]
     extension_id: String,
-    account_id: String,
+    account_id: Option<String>,
 }
 
 #[derive(Deserialize)]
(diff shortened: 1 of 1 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -190,13 +190,11 @@ async fn main() -> Result<()> {
...
-        aws_secret_access_key: env::var("AWS_SECRET_ACCESS_KEY")
-            .expect("AWS_SECRET_ACCESS_KEY not set"),
-        aws_session_token: env::var("AWS_SESSION_TOKEN").expect("AWS_SESSION_TOKEN not set"),
-        function_name: env::var("AWS_LAMBDA_FUNCTION_NAME")
-            .expect("AWS_LAMBDA_FUNCTION_NAME not set"),
+        region: env::var("AWS_DEFAULT_REGION").unwrap_or("us-east-1".to_string()),
+        aws_access_key_id: env::var("AWS_ACCESS_KEY_ID").unwrap_or("".to_string()),
+        aws_secret_access_key: env::var("AWS_SECRET_ACCESS_KEY").unwrap_or("".to_string()),
+        aws_session_token: env::var("AWS_SESSION_TOKEN").unwrap_or("".to_string()),
+        function_name: env::var("AWS_LAMBDA_FUNCTION_NAME").unwrap_or("".to_string()),
```
duncanista, line 197: I think an `unwrap_or_default` might be better here

astuyve: yup

astuyve: sorry, new baby brain. literally forgot about that method

duncanista (approved):

PR #345, opened 2024-08-03
chore(bottlecap): reduce allocations for logs processing

duncanista:
# What?

Reduces allocations for logs components
- Aggregator
- Processor

# Motivation

Just save some allocations by pre-allocating vectors and not generating them everytime.

# Notes


astuyve (approved):
plz stop working on the weekends

PR #348, opened 2024-08-05
feat: Allow trace disabled plugins

astuyve:
Allows `DD_TRACE_DISABLED_PLUGINS` for people who want to optionally disable some plugins in the tracer.


duncanista (approved):
Might be good to add `DD_TRACE_DEBUG=<bool>` too?

PR #354, opened 2024-08-08
feat: Allowlist additional env vars

astuyve:
Allowing more env vars which we either don't use in serverless or aren't used by the lambda extension

except profiling, which we will fail over into the main agent if it's present and enabled


duncanista (approved):
Should we add the Git Repo tag too?

PR #358, opened 2024-08-13
aj/allow apm replace tags array

astuyve:
APM replace tags can be both a string and a JSON object. But we don't care about its value - that's for the tracers.

This change introduces a new `ObjectIgnore::Ignore` enum to save us the string alloc on the heap and to avoid failing over when customers have more complicated replace rules.

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -1,5 +1,6 @@
 pub mod flush_strategy;
 pub mod log_level;
+pub mod object_ignore;
 pub mod processing_rule;
 
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -29,7 +31,7 @@ pub struct Config {
     #[serde(deserialize_with = "deserialize_processing_rules")]
     pub logs_config_processing_rules: Option<Vec<ProcessingRule>>,
     pub apm_enabled: bool,
-    pub apm_replace_tags: Option<String>,
+    pub apm_replace_tags: Option<ObjectIgnore>,
```
alexgallotta, line 34: Can this be just a String, instead of using a customer deserializer?

alexgallotta (approved):

PR #363, opened 2024-08-15
fix(bottlecap): set explicit deny list and allow yaml usage

duncanista:
# What?

- Allows usage of `datadog.yaml`
- Removes the strict deny list, by now creating one of non-supported features: appsec and profiling.

# Motivation

`datadog.yaml` needs to be allowed, and we can't keep on adding more environment variables which we don't support, since they are not related to the project.

# Testing

Added unit tests


bottlecap/src/config/mod.rs
```diff
@@ -27,47 +44,22 @@ pub struct Config {
...
-    pub runtime_metrics_enabled: bool,
-    pub git_repository_url: Option<String>,
 }
 
 impl Default for Config {
     fn default() -> Self {
         Config {
             // General
-            site: "datadoghq.com".to_string(),
+            site: String::default(),
```
duncanista, line 62: Changed it because if you had in `datadog.yaml` something like `datadoghq.eu`, this would be overwritten by the default value, this is wrong. Instead, I just set it after if it's empty.

astuyve (approved):
Can you remove the ObjectIgnore module now that it's not used?

PR #366, opened 2024-08-19
[SVLS-5280] bundle .NET tracer script with serverless-init

DylanLovesCoffee:
work-in-progress for incident-29739

Shell script provided by Storms: https://github.com/DataDog/serverless-self-monitoring/compare/storms/incident-29739 

https://datadoghq.atlassian.net/browse/SVLS-5280

### DIFF
--- a/scripts/Dockerfile.serverless-init.alpine.build
+++ b/scripts/Dockerfile.serverless-init.alpine.build
@@ -2,3 +2,4 @@ FROM scratch
 ARG TARGETARCH
 
 COPY ./bin/datadog_extension-$TARGETARCH-alpine/extensions/datadog-agent /datadog-init
+COPY ./bin/datadog_extension-$TARGETARCH/extensions/dotnet.sh /datadog-init/dotnet.sh
--- a/scripts/Dockerfile.serverless-init.alpine.build
+++ b/scripts/Dockerfile.serverless-init.alpine.build
@@ -2,5 +2,4 @@ FROM scratch
 ARG TARGETARCH
 
 COPY ./bin/datadog_extension-$TARGETARCH-alpine/extensions/datadog-agent /datadog-init
-COPY ./scripts/serverless_init_dotnet.sh /dotnet.sh
-
+COPY ./bin/serverless_init_dotnet.sh /dotnet.sh
--- a/scripts/Dockerfile.serverless-init.alpine.build
+++ b/scripts/Dockerfile.serverless-init.alpine.build
@@ -2,4 +2,5 @@ FROM scratch
 ARG TARGETARCH
 
 COPY ./bin/datadog_extension-$TARGETARCH-alpine/extensions/datadog-agent /datadog-init
-COPY ./bin/datadog_extension-$TARGETARCH/extensions/dotnet.sh /datadog-init/dotnet.sh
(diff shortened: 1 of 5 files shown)


scripts/Dockerfile.serverless-init.alpine.build
```diff
@@ -2,3 +2,4 @@ FROM scratch
 ARG TARGETARCH
 
 COPY ./bin/datadog_extension-$TARGETARCH-alpine/extensions/datadog-agent /datadog-init
+COPY ./bin/serverless_init_dotnet.sh /dotnet.sh
```
joeyzhao2018, line 5: do we have to add this for some reason for alpine?

joeyzhao2018 (approved):
1 minor comment. non-blocking.  🚢

scripts/Dockerfile.serverless-init.alpine.build
DylanLovesCoffee: ah yeah, you're right, that's unnecessary for alpine ✏️

scripts/serverless_init_dotnet.sh
```diff
@@ -0,0 +1,25 @@
...
+if [ -f "/etc/alpine-release" ]; then
+    echo "Alpine images are not supported for .NET automatic instrumentation with serverless-init"
+    exit 1
+fi
+
+# Make sure curl is installed
+apt-get update && apt-get install -y curl
+
+# Get latest released version of dd-trace-dotnet
+TRACER_VERSION=$(curl -s https://api.github.com/repos/DataDog/dd-trace-dotnet/releases/latest | jq -r '.tag_name' | cut -d '"' -f 4 | cut -c2-)
```
duncanista, line 13: What if they hit rate limit? Should we hardcode a fallback version?

DylanLovesCoffee: yeah good point 👍

joeyzhao2018: had this page open and just saw this and thinking it's probably safer to let the operation fail (fast) than fallback to any version which might cause unexpected issue.

duncanista: > it's probably safer to let the operation fail (fast)
wouldn't this crash your customer if it fails?

joeyzhao2018: that's a very good question. I was thinking that we should favor failing at deployment time over any fallback solution. [more about the evil of fallback](https://aws.amazon.com/builders-library/avoiding-fallback-in-distributed-systems/) 
However, this is a good question on when it's "crashing". I assumed that the image build is built once and reused when scaling happens, which is the most sensitive time.
If it's not reused, crashing during scaling may be indeed undesirable.
But could it even create a scenario where some instances running with one version and others using fallback versions?

duncanista: > But could it even create a scenario where some instances running with one version and others using fallback versions?

I don't know, you tell me!

I agree, I don't like fallbacks, yet I don't want us to crash customers. Pick as your product needs 😛

DylanLovesCoffee: ah yeah, that's true. It might not be so obvious to the user that there's a fallback version too should the version request fail when they're running this in their pipelines. I'll revert

PR #371, opened 2024-08-22
chore(bottlecap): fast failover

duncanista:
# What?

Moves the `Failover` specific configuration to its own struct, allowing us to failover fast.

# Motivation

It's easier to maintain having this deny list, by quickly checking the values of `extension_version` from Env and Yaml before checking other values expected.

# Notes

Also updated how we merge logs processing rules, as Env has precedence over Yaml

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -122,6 +122,13 @@ fn failsover(figment: &Figment) -> Result<(), ConfigError> {
         ));
     }
 
+    let datadog_wrapper_set =
+        std::env::var("AWS_LAMBDA_EXEC_WRAPPER").unwrap_or_default() == "/opt/datadog_wrapper";
+    if datadog_wrapper_set {
+        log_failover_reason("datadog_wrapper");
+        return Err(ConfigError::UnsupportedField("datadog_wrapper".to_string()));
+    }
+
     if failover_config.serverless_appsec_enabled || failover_config.appsec_enabled {
         log_failover_reason("appsec_enabled");
         return Err(ConfigError::UnsupportedField("appsec_enabled".to_string()));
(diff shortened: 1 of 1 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -104,12 +148,18 @@ pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
...
+
+    // Failover
+    failsover(&figment)?;
 
     let (mut config, yaml_config): (Config, YamlConfig) =
         match (figment.extract(), yaml_figment.extract()) {
             (Ok(env_config), Ok(yaml_config)) => (env_config, yaml_config),
-            (_, Err(err)) | (Err(err), _) => return Err(ConfigError::ParseError(err.to_string())),
+            (_, Err(err)) | (Err(err), _) => {
+                println!("Failed to parse Datadog config: {err}");
```
duncanista, line 160: Safe to log, as we know we are opted in

bottlecap/src/config/mod.rs
```diff
@@ -93,6 +97,46 @@ fn log_failover_reason(reason: &str) {
...
+        return Err(ConfigError::UnsupportedField("appsec_enabled".to_string()));
+    }
+
+    if failover_config.profiling_enabled {
+        log_failover_reason("profiling_enabled");
+        return Err(ConfigError::UnsupportedField(
+            "profiling_enabled".to_string(),
+        ));
+    }
+
```
astuyve, line 136: Let's add a failover condition for `AWS_LAMBDA_WRAPPER_SCRIPT` as well. That's required for Java and DotNet and would probably catch most universal instrumentation users who may opt in without reading the docs fully.

astuyve (approved):

bottlecap/src/config/mod.rs
duncanista: Sg

PR #377, opened 2024-08-27
feat: break if we shouldn't wait for the end

astuyve:
Technically our `periodic` strategy has always been `periodic` + end flushing.
Because we no longer attempt the smart/adaptive flush strategy and instead default by flushing every 1s + at the end of every invocation, we can now diverge from the main agent and allow the `periodic` strategy to skip flushing at the `end`, and thus not block on the `platform.runtimeDone` event.

This also unblocks people who are using local lambda emulation to set `DD_SERVERLESS_FLUSH_STRATEGY=periodically,1`. This prevents the extension from waiting for an event which will never come from the runtime-interface-emulator.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -465,6 +465,9 @@ async fn extension_loop_active(
                         trace_flusher.manual_flush(),
                         stats_flusher.manual_flush()
                     );
+                    if !flush_control.should_flush_end() {
+                        break;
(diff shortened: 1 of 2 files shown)


duncanista (approved):
LGTM – is there any way we could include a unit test in the future which used the local emulation?

astuyve:
Not really because we're entirely dependent on the telemetry API working (which it doesn't right now). Any test we write would be pretty trivial without that support.

PR #380, opened 2024-09-03
feat: increase to 10k contexts

astuyve:
I tested this out with 10k contexts along with Python APM tracing enabled and we barely touched 74mb, so I think 10x-ing this is fine and unblocks our customer for now.
image

We can replace this with a bump allocator or dynamic flushing as needed, but I think this is the fastest path to resolution

### DIFF
--- a/bottlecap/src/traces/trace_agent.rs
+++ b/bottlecap/src/traces/trace_agent.rs
@@ -45,10 +45,6 @@ pub enum ApiVersion {
 impl TraceAgent {
     pub async fn start_trace_agent(&self) -> Result<(), Box<dyn std::error::Error>> {
         let now = Instant::now();
-        debug!(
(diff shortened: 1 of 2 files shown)


bottlecap/src/traces/trace_agent.rs
```diff
@@ -45,10 +45,6 @@ pub enum ApiVersion {
 impl TraceAgent {
     pub async fn start_trace_agent(&self) -> Result<(), Box<dyn std::error::Error>> {
         let now = Instant::now();
-        debug!(
```
duncanista, line 48: We still need this?

duncanista (approved):

bottlecap/src/traces/trace_agent.rs
duncanista: Nvm – misread, thought you were adding it

PR #385, opened 2024-09-09
feat: Specify aws_lc_rs as our default crypto provider

astuyve:
Required now for as rustls will be upset if both `ring` and `aws-lc-rs` are present.

### DIFF
--- a/bottlecap/Cargo.toml
+++ b/bottlecap/Cargo.toml
@@ -41,6 +41,7 @@ base64 = { version = "0.22.0", default-features = false }
 rmp-serde = { version = "1.3.0", default-features = false }
 rmpv = { version = "1.3.0", default-features = false }
 rmp = { version = "0.8.14", default-features = false }
+rustls = { version = "0.23.12", default-features = false, features = ["aws-lc-rs"] }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


bottlecap/Cargo.toml
```diff
@@ -41,6 +41,7 @@ base64 = { version = "0.22.0", default-features = false }
 rmp-serde = { version = "1.3.0", default-features = false }
 rmpv = { version = "1.3.0", default-features = false }
 rmp = { version = "0.8.14", default-features = false }
+rustls = { version = "0.23.12", default-features = false, features = ["aws-lc-rs"] }
```
tobz, line 44: FWIW, it's generally better to avoid specifying patch versions (or potentially minor versions, when dependencies are at a major version 1 or later) as Cargo's default semver resolver treats `0.23.12` as `>= 0.23.12, < 0.24`.... unless you need _at least_ that version or above.

Doing so just lets Cargo potentially be a little more flexible with how it resolves which version to use across the entire dependency graph.

astuyve: I ran `cargo add rustls --no-default-features`, should I be removing the minor version it adds?

astuyve: Will drop patch version specs in another PR

duncanista (approved):

PR #386, opened 2024-09-09
fix: licenses

astuyve:
Fix licenses file


duncanista (approved):
cannot see any changes really, so :shipit:

PR #395, opened 2024-09-16
feat: add async resolution of runtime

alexgallotta:
I went though multiple implementations and eventually settled for OnceLock:
- using RWLock there are no guarantees that the producer will be done before the consumers are called, risking deadlocks, same for broadcast channels 
- https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html can be used to initialize the runtime tag value and be passed around behind a ARC. OnceLock won't block on `get` and can be initialize once. If the value is not ready, the tag won't be added for that datapoint. Datadog aggregation should handle and fill that properly
- the runtime resolution logic will contains retry and such, it's the easy part. Not yet added
- tag provider is used as a struct in different ways, adding anything async there will exponentially complicate the logic and force huge refactors. It's probably a wrong mix of data and logic 
- dogstatsd will need changes too, so another PR in libdatadog must be added. This will be a breaking change. To avoid further issue, the OnceLock will have a struct rather than String so more async values can be added later

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -9,6 +9,7 @@
 #![deny(missing_copy_implementations)]
 #![deny(missing_debug_implementations)]
 
+use bottlecap::tags::lambda::tags::resolve_runtime;
 use bottlecap::{
     base_url,
     config::{self, AwsConfig, Config},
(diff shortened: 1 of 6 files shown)


blt (commented):
I buy the use of a `OnceLock`. Tidy stuff. I'm missing in this PR where you are `set`ing or `try_insert`ing but I've been travelling and am sleepy so I might just be missing it. 

I would also note that you might be able to elide the use of the `Arc` by having access to the runtime be mediated through a function, have that function interact with a static `OnceLock`. That would change your testing strategy some but it would save needing to pass an `Arc` around, dunno that it would have much impact on runtime behavior. It's a trick I like in my work though.

PR #401, opened 2024-09-27
feature(bottlecap): add trace context extractor

duncanista:
# What?

Adds an implementation of trace context propagator, currently only focusing on extraction.

# Motivation

Inferred spans and [SVLS-5289](https://datadoghq.atlassian.net/browse/SVLS-5289)

# Notes

[SVLS-5289]: https://datadoghq.atlassian.net/browse/SVLS-5289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bottlecap/src/traces/propagation/text_map_propagator.rs
```diff
@@ -0,0 +1,508 @@
...
+    sampling_priority: Option<i8>,
+    origin: Option<String>,
+    lower_order_trace_id: Option<String>,
+}
+
+#[derive(Clone, Copy)]
+pub struct TraceparentPropagator;
+
+impl TextMapPropagator for TraceparentPropagator {
+    fn extract(&self, carrier: &dyn Extractor) -> SpanContext {
```
alexgallotta, line 209: do we really need struct + trait + methods here for traceparent and tracestate propagators?
Can we get away with extract_parent and extract_state?

bottlecap/src/traces/propagation/carrier.rs
```diff
@@ -0,0 +1,68 @@
+/// Code inspired, and copied, by OpenTelemetry Rust project.
+/// <https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/propagation/mod.rs>
+///
+use std::collections::HashMap;
+
+/// Injector provides an interface for a carrier to be used
+/// with a Propagator to inject a Context into the carrier.
+///
+pub trait Injector {
```
alexgallotta, line 9: Same for these 2. Can we avoid/postpone creating extra traits?

alexgallotta (commented):
Looks amazing! Great job! Left 2 comments on some stuff that might be removed and avoid some extra code/layering

bottlecap/src/traces/propagation/text_map_propagator.rs
duncanista: > `traceparent and tracestate propagators`
Do you mean The `Traceparent` propagator which includse the extraction of the `tracestate` + `traceparent`?

If so, I tried using this `struct + trait + methods` in order to make the code easier to maintain, and also following the current OTel implemenetation, which I think was simple enough to understand, what do you think?

bottlecap/src/traces/propagation/carrier.rs
duncanista: This allows us to directly implement this traits for carriers which we might not have control of, for example, here I tend to convert `hyper::HeaderMap` into a `HashMap` so I can use the `extract` and `inject` methods on the struct right away.

At the top of the file I explain that I directly copied this from the OTel rust implementation for extractor/injectors.

I think this would help us whenever we have to expand, and its also very generic, what do you think of this?

bottlecap/src/traces/propagation/text_map_propagator.rs
```diff
@@ -0,0 +1,508 @@
...
+#[derive(Clone, Copy)]
+pub struct DatadogPropagator;
+
+impl TextMapPropagator for DatadogPropagator {
+    fn extract(&self, carrier: &dyn Extractor) -> SpanContext {
+        Self::extract_context(carrier).unwrap_or_default()
+    }
+
+    fn inject(&self, _context: SpanContext, _carrier: &mut dyn Injector) {
+        todo!();
```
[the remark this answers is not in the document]
duncanista: Yeah, this ones i'm not going to particularly adress in this PR, since we don't care about injection yet, maybe I mark them as that they're going to panic since they are not implemented?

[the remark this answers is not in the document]
astuyve: Nope this is totally fine

PR #405, opened 2024-10-04
feat(bottlecap): create Inferred Spans baseline + infer API Gateway HTTP spans

duncanista:
# What?

# Motivation

# Notes


bottlecap/src/config/mod.rs
```diff
@@ -188,6 +188,7 @@ pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
 }
 
 #[allow(clippy::module_name_repetitions)]
+#[derive(Debug, Clone)]
```
alexgallotta, line 191: Do we need to add clone trait here?
I thought the configs were immutable and passed as ref/Arc everywhere

bottlecap/src/lifecycle/invocation/triggers/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use datadog_trace_protobuf::pb::Span;
+use serde::{ser::SerializeMap, Serializer};
+use serde_json::Value;
+
+pub mod api_gateway_http_event;
+
+pub trait Trigger: Sized {
+    fn new(payload: Value) -> Option<Self>;
+    fn is_match(payload: &Value) -> bool;
```
alexgallotta, line 11: Would it be more convenient to have a general matching function returning a type instead of is_match for every subclass? To avoid doing is_match till something matches and maybe hiding bugs when a match hides running other match logic

bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -0,0 +1,109 @@
...
+                s.duration = duration;
+            }
+
+            s.trace_id = invocation_span.trace_id;
+        }
+    }
+
+    fn generate_span_id() -> u64 {
+        // todo: secure random id with OsRng for SnapStart
+        let mut rng = rand::thread_rng();
```
astuyve, line 99: we should do this before merge, I can take a look if you want

astuyve (approved):
looks good to me except rng and tags like cold start.

bottlecap/src/lifecycle/invocation/triggers/mod.rs
duncanista: I thought so, but I wanted to not have to separate payload-specific logic into its own trait, I'm still exploring the idea, but I think this might change depending on how it would look like when we add more cases.

How would you change this code? My main constraints are: not deserialize on every match, just deserialize once into `serde_json::Value` and then try to find the actual type of the trigger, wdyt?

bottlecap/src/lifecycle/invocation/span_inferrer.rs
duncanista: Ok, should be a simple fix, will look into it tomorrow – what do you mean by `tags like cold start`?

astuyve: You've got a comment in the code about adding tags for cold starts/proactive inits on the span

PR #413, opened 2024-10-15
feat(bottlecap): Add Composite Trace Propagator

duncanista:
# What?

Creates a composite propagator to extract trace context.

This is because we have specific logic on how we handle/resolve trace context in the tracers.

# How?

- New struct `DatadogCompositePropagator` using `DatadogHeaderPropagator` and `TraceContextPropagator`.
- Added configuration options so the propagator can decide how to act.

# Motivation

[SVLS-5769](https://datadoghq.atlassian.net/browse/SVLS-5769) and...

- Maintain feature parity with the tracers.

# Notes
- Unit tests were copied from Python tracer, seen as the source of "true".

[SVLS-5769]: https://datadoghq.atlassian.net/browse/SVLS-5769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/traces/propagation/mod.rs
+++ b/bottlecap/src/traces/propagation/mod.rs
@@ -1,3 +1,202 @@
+use std::{collections::HashMap, sync::Arc};
+
+use crate::{
+    config::{self, trace_propagation_style::TracePropagationStyle},
+    traces::context::SpanContext,
+};
+use carrier::{Extractor, Injector};
+use datadog_trace_protobuf::pb::SpanLink;
+use text_map_propagator::{
+    BAGGAGE_PREFIX, DATADOG_HIGHER_ORDER_TRACE_ID_BITS_KEY, DATADOG_LAST_PARENT_ID_KEY,
+    TRACESTATE_KEY,
+};
+
 pub mod carrier;
 pub mod error;
 pub mod text_map_propagator;
+
+pub trait Propagator {
+    fn extract(&self, carrier: &dyn Extractor) -> Option<SpanContext>;
+    fn inject(&self, context: SpanContext, carrier: &mut dyn Injector);
+}
+
+pub struct DatadogCompositePropagator {
+    propagators: Vec<Box<dyn Propagator + 'static>>,
+    config: Arc<config::Config>,
+}
+
+#[allow(clippy::never_loop)]
+impl Propagator for DatadogCompositePropagator {
+    fn extract(&self, carrier: &dyn Extractor) -> Option<SpanContext> {
+        if self.config.trace_propagation_extract_first {
+            for propagator in &self.propagators {
+                let context = propagator.extract(carrier);
+
+                if self.config.trace_propagation_http_baggage_enabled {
+                    if let Some(mut context) = context {
+                        Self::attach_baggage(&mut context, carrier);
+                        return Some(context);
+                    }
+                }
+
+                return context;
+            }
+        }
+
+        let (contexts, styles) = self.extract_available_contexts(carrier);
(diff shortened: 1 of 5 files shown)


bottlecap/src/traces/propagation/mod.rs
```diff
@@ -1,3 +1,873 @@
...
+            if let Some(context) = propagator.extract(carrier) {
+                contexts.push(context);
+                styles.push(self.config.trace_propagation_style_extract[i]);
+            }
+        }
+
+        (contexts, styles)
+    }
+
+    fn resolve_contexts(
```
alexgallotta, line 106: can this method be outside of the struct implementation?

bottlecap/src/traces/propagation/mod.rs
```diff
@@ -1,3 +1,873 @@
...
+
+            i += 1;
+        }
+
+        primary_context.links = links;
+
+        primary_context
+    }
+
+    fn attach_baggage(context: &mut SpanContext, carrier: &dyn Extractor) {
```
alexgallotta, line 190: can this method be outside of the struct implementation?

bottlecap/src/traces/propagation/mod.rs
```diff
@@ -1,3 +1,873 @@
...
+    }
+
+    fn inject(&self, _context: SpanContext, _carrier: &mut dyn Injector) {
+        todo!()
+    }
+}
+
+impl DatadogCompositePropagator {
+    #[must_use]
+    pub fn new(config: Arc<config::Config>) -> Self {
```
alexgallotta, line 66: If also this method is removed, there is no need of the impl altogether.
It's a bit heavy for a constructor to have all this logic, it could be a builder/function

alexgallotta (commented):
Great job overall!
I would argue that in a lot of places struct/traits could be avoided and make the code easier, more readable, easier to test and shorter though

bottlecap/src/traces/propagation/mod.rs
duncanista: Since this is the only place being used, I'd rather keep it here, there's no other place where this is being used. What would be the benefit of keeping it outside of the struct?

bottlecap/src/traces/propagation/mod.rs
duncanista: Same question as the other comment

bottlecap/src/traces/propagation/mod.rs
duncanista: How would you modify this? I decided to follow this because of the behavior changes on the configuration, and to save the state of the available propagators, at the end of the day, it will be very cheap in computation

alexgallotta: I took some time to modify it and give it an idea of what I thought
https://github.com/DataDog/datadog-lambda-extension/pull/418/files#diff-2fb5a4da2205bb01fbaeabd0078f986e4fd54c0ca51a14a628686d2aff0d41bdR9

I think it's already a great job and there is a lot of business logic I am not familiar with!

My take, from what I've read, is to use abstractions and traits/struct very sparingly and when they are really worth it.
Usually better start with basic function and modules most of the time. It helps to also see things differently when the struct impl is not around, and when refactoring or changing something you don't need to change 10 fields all around the code/function initialization!

That PR is not necessarily to merge or continue working on, just to show how it could look like with a different approach

alexgallotta (approved):

PR #417, opened 2024-10-17
[SVLS-5714] Add lambda network enhanced metrics

shreyamalpani:
### What does this PR do?
This PR introduces three new enhanced lambda metrics. Each of the metrics are emitted once per invocation and represent network usage by the lambda function.

The three new metrics are:

- `aws.lambda.enhanced.rx_bytes` - bytes received by the function
- `aws.lambda.enhanced.tx_bytes` - bytes transmitted by the function
- `aws.lambda.enhanced.total_network` - sum of `aws.lambda.enhanced.rx_bytes` and `aws.lambda.enhanced.tx_bytes`

- Build the lambda extension
- Invoke the function


bottlecap/src/proc/proc.rs
duncanista, line 1: Why are we creating a `proc/proc.rs` if this could live in `proc/mod.rs`?

bottlecap/src/proc/test_data/net/invalid_dev_malformed
duncanista, line 1: All testing fixtures should live in the root `tests/` folder

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -1,21 +1,29 @@
...
+use std::collections::VecDeque;
 use std::env::consts::ARCH;
 use std::sync::{Arc, Mutex};
-use tracing::error;
+use tracing::{debug, error};
 
 pub struct Lambda {
     pub aggregator: Arc<Mutex<Aggregator>>,
     pub config: Arc<crate::config::Config>,
+    enhanced_metric_data: VecDeque<Vec<Box<dyn metric_data::EnhancedMetricData>>>,
```
duncanista, line 16: What's the decision on this type of architecture?

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -1,21 +1,29 @@
 use super::constants::{self, BASE_LAMBDA_INVOCATION_PRICE};
+use super::metric_data;
```
duncanista, line 2: Let's not use `super`, but `crate::...`

bottlecap/src/proc/proc.rs
shreyamalpani: I thought mod.rs only contains the names of the other files based on the other directories, I can move it!

bottlecap/src/metrics/enhanced/lambda.rs
shreyamalpani: Each invocation has a set of offsets/data that needs to be stored at the beginning of the invocation, which we would be storing in the Vec (currently it only holds NetworkData, but in the future would include others like CPUData, etc.)

The queue is there so that when the metrics are being sent after the PlatformReport event, which may be at the start of the subsequent invocation (at which point there would be two of each type of metric data), we only calculate the set of metrics for the invocation that has finished by popping off the Vec object at the front.

alexgallotta: Why are we using a mutable field for a trait? Can't the collection of data be a simple struct or vec?

bottlecap/src/proc/proc.rs
alexgallotta: btw I think that the preferred way to define modules in rust is
`folder/module.rs`
that is searched by default.
The old way, still supported, is to use the mod.rs

bottlecap/src/metrics/enhanced/metric_data.rs
```diff
@@ -0,0 +1,73 @@
...
+use dogstatsd::{
+    aggregator::Aggregator,
+    metric::{Metric, MetricValue},
+};
+use std::sync::MutexGuard;
+use tracing::debug;
+use tracing::error;
+
+#[allow(clippy::module_name_repetitions)]
+pub trait EnhancedMetricData {
```
alexgallotta, line 12: We have a lot of struct/traits/hiearchy's in our code, I think it's increasing the already high number of LOC and serve us no favor

bottlecap/src/metrics/enhanced/lambda.rs
duncanista: I just don't understand why we would all of those files for this use case. This look overly complex for what we actually have in our source code, this could be simplified even more

DarcyRaynerDD: I gave some advice here to Shreya, but I think the basic idea is, there will be many different implementations of these metrics, that will all have different behaviour, eg. different intervals they are sampled, different times they should be compared to previous values, and store different data. To me, that makes sense as a Trait with multiple different structs implementing it. That way, we don't need all the state/logic bundled together into a megastruct or enum. Now, I would  still suggest as a change here, which would be not to use a Dequeue and just have a regular vec/array, but to store all internal data/state for a metric type inside the struct itself, including if it needs to queue state.

duncanista: I had a chat with Shreya today, I showed the current `invocation/context.rs` in the Universal Instrumentation branch which contains a 5-context buffer which has relevant data for metrics. This way we can have some of this data living there, maybe we can have a broader chat with the team around how we handle this, I do agree we could have a trait which multiple structs could implement!

PR #424, opened 2024-10-22
[SVLS-5714] Add lambda network enhanced metrics

shreyamalpani:
### What does this PR do?
This PR introduces three new enhanced lambda metrics. Each of the metrics are emitted once per invocation and represent network usage by the lambda function.

The three new metrics are:

- `aws.lambda.enhanced.rx_bytes` - bytes received by the function
- `aws.lambda.enhanced.tx_bytes` - bytes transmitted by the function
- `aws.lambda.enhanced.total_network` - sum of `aws.lambda.enhanced.rx_bytes` and `aws.lambda.enhanced.tx_bytes`

- Build the lambda extension
- Invoke the function

### Additional Notes
- Will implement the structure to store the offsets for the different types of metrics together in the next PR


bottlecap/src/proc/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct NetworkData {
+    pub rx_bytes: f64,
+    pub tx_bytes: f64,
+}
+
+pub fn get_network_data() -> Result<NetworkData, Box<dyn std::error::Error>> {
+    get_network_data_helper(PROC_NET_DEV_PATH)
+}
```
duncanista, line 19: Is this method really needed?

bottlecap/src/proc/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+
+fn get_network_data_helper(path: &str) -> Result<NetworkData, Box<dyn std::error::Error>> {
+    let file = File::open(path).map_err(|e| Box::new(e) as Box<dyn Error>)?;
```
duncanista, line 22: Why do we need to box the error?

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -105,6 +107,65 @@ impl Lambda {
...
+            match proc::get_network_data() {
+                Ok(data) => {
+                    Self::generate_network_enhanced_metrics(offset, data, &mut aggr);
+                }
+                Err(_e) => {
+                    debug!("Could not emit network enhanced metrics");
+                }
+            }
+        } else {
+            debug!("Could not emit network enhanced metrics");
```
duncanista, line 165: Maybe we change the wording? This is not accurate, it's more like no data, so we didn't send anything

bottlecap/src/proc/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+        let rx_bytes: f64 = match values.next().and_then(|s| s.parse().ok()) {
+            Some(value) => value,
+            None => continue,
+        };
+        let tx_bytes: f64 = match values.nth(7).and_then(|s| s.parse().ok()) {
+            Some(value) => value,
+            None => continue,
+        };
+
+        return Ok(NetworkData { rx_bytes, tx_bytes });
```
duncanista, line 45: I'm not sure I understand the logic, as I'm not familiar with it, could you make it more readable, or add some code comments in order to help the reader understand why you are doing this? Or document what this method does? 

Someone with no context would struggle with this code

bottlecap/src/proc/mod.rs
shreyamalpani: This was following the pattern from the agent because we use the inner method for unit testing

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -105,6 +107,65 @@ impl Lambda {
         }
     }
 
+    pub(crate) fn generate_network_enhanced_metrics(
+        network_offset: NetworkData,
+        network_data: NetworkData,
```
duncanista, line 112: I wonder if we should rename them to denote what they are, the offset is great, but not sure if `network_data` is descriptive enough

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -105,6 +107,65 @@ impl Lambda {
...
+            error!("failed to insert tx_bytes metric: {}", e);
+        }
+
+        let metric = Metric::new(
+            constants::TOTAL_NETWORK_METRIC.into(),
+            MetricValue::distribution(total_network),
+            None,
+        );
+        if let Err(e) = aggr.insert(metric) {
+            error!("failed to insert total_network metric: {}", e);
```
duncanista, line 143: Uppercase first

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -375,6 +375,9 @@ async fn extension_loop_active(
                     request_id, deadline_ms, invoked_function_arn
                 );
                 lambda_enhanced_metrics.increment_invocation_metric();
+                let mut p = invocation_processor.lock().await;
+                p.on_invoke_event(request_id);
```
duncanista, line 379: Wonder since this is always invoked first, if we might need to change the constructor for the context buffer, simplifying our logic. As I'm not sure if this would guarantee us to always have this operation to exist before any other hook from the telemetry API, we might need some experiment to check!

bottlecap/src/proc/mod.rs
duncanista: Rename the helper then to `get_network_data_from_path`

duncanista:
Amazing work, this is taking a much much better form! 🎉

bottlecap/src/proc/mod.rs
shreyamalpani: You're right, it isn't needed! Changing it to io::Error

bottlecap/src/proc/mod.rs
duncanista: Actually, why would we even need to have the `get_network_data`, can't we just use the `get_network_data_from_path`?

bottlecap/src/metrics/enhanced/lambda.rs
shreyamalpani: True, would "Could not generate network enhanced metrics" or "Could not find data to generate network enhanced metrics" sound better

duncanista: The second looks better!

bottlecap/src/proc/mod.rs
shreyamalpani: I guess it isn't necessary, but we use the get_network_data function in two different files, so I'm not sure if we should be importing and passing in that constant in both

duncanista: Fair point, I'll leave the decision to you

bottlecap/src/metrics/enhanced/lambda.rs
shreyamalpani: Would `network_data_offset`/`network_data_start` and then `network_data_end` be more clear

bottlecap/src/bin/bottlecap/main.rs
shreyamalpani: Had a discussion about this with Jordan, will do in a future PR!

bottlecap/src/proc/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+    let file = File::open(path)?;
+    let reader = io::BufReader::new(file);
+
+    for line in reader.lines() {
+        let line = line?;
+        let mut values = line.split_whitespace();
+
+        // Check for the line containing lambda network data by interface name
+        if let Some(interface_name) = values.next() {
+            if interface_name.starts_with(LAMDBA_NETWORK_INTERFACE) {
```
duncanista, line 30: I think you can simplify it more by chaining this to something like
```suggestion
       if values.next().map_or(false, |interface_name| 
            interface_name.starts_with(LAMDBA_NETWORK_INTERFACE)) {
}
```

bottlecap/src/proc/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+                // Read the value for bytes received if present, otherwise break and return error
+                let rx_bytes: f64 = match values.next().and_then(|s| s.parse().ok()) {
+                    Some(value) => value,
+                    None => break,
+                };
+                // Read the value for bytes transmitted if present, otherwise break and return error
+                let tx_bytes: f64 = match values.nth(7).and_then(|s| s.parse().ok()) {
```
duncanista, line 37:
```suggestion
                let tx_bytes: f64 = match values.nth(7).and_then(|s| s.parse().ok()) {
```
Can you explain the magic number `7` in comment?

bottlecap/src/proc/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+                    Some(value) => value,
+                    None => break,
+                };
```
duncanista, line 40: Wonder if we can simplify this to something like 
```suggestion
                let rx_bytes = values.next().and_then(|s| s.parse().ok());
                let tx_bytes = values.nth(7).and_then(|s| s.parse().ok());
                // Read the value for bytes transmitted if present, otherwise break and return error
                match (rx_bytes, tx_bytes) {
                    (Some(rx_v), Some(tx_v)) => Ok(...)
                    (_, _) => Err(...)
                }
```

duncanista (approved):
Left some non-blocking comments! Amazing work

bottlecap/src/proc/mod.rs
shreyamalpani: Thank you for the suggestions!

PR #428, opened 2024-10-24
fix(bottlecap): add trigger tags to invocation span

duncanista:
# What?

Trigger tags now will be showed on the invocation span.

# How

Uses the `get_tags` method in the `SpanInferrer` when an inferred span is present, allowing us to later use to add them to the invocation span.

# Notes

Minor refactored removing unnecessary method for a reference of our inferred span.


bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -142,15 +146,17 @@ impl SpanInferrer {
         rng.gen()
     }
 
-    /// Returns a reference to the inner `self.inferred_span`
+    /// Returns a clone of the carrier associated with the inferred span
     ///
     #[must_use]
-    pub fn get_inferred_span(&self) -> &Option<Span> {
-        &self.inferred_span
+    pub fn get_carrier(&self) -> Option<HashMap<String, String>> {
```
alexgallotta, line 152: do we need to use getter here instead of .field (maybe with crate pub)?

duncanista: Yes, because we are cloning it instead of passing a reference

alexgallotta: you can still call span_inferrer.carrier.clone() thought, right?

duncanista: I could make it a public method, but it's recommended to have a getter method when you are doing cloning on non-primitives, all clones are considered heavy in this case

PR #430, opened 2024-10-28
Generate CPU Enhanced Metrics

shreyamalpani:
### What does this PR do?

This PR introduces eight new enhanced lambda metrics related to CPU usage. Each of the metrics are emitted once per invocation.

Three of the metrics represent CPU time spent running the lambda function:
- `aws.lambda.enhanced.cpu_system_time` - time spent by the CPU running in kernel mode
- `aws.lambda.enhanced.cpu_user_time` - time spent by the CPU running in user mode
- `aws.lambda.enhanced.cpu_total_time` - sum of `aws.lambda.enhanced.cpu_system_time` and `aws.lambda.enhanced.cpu_user_time`

The other five metrics represent CPU utilization by the lambda function:
- `aws.lambda.enhanced.cpu_total_utilization_pct` - total CPU utilization of the function, as a normalized percent (e.g. 35%)
- `aws.lambda.enhanced.cpu_total_utilization` - total CPU utilization of the function, as normalized cores (e.g. 1.2 cores)
- `aws.lambda.enhanced.num_cores` - total number of cores available in the environment (e.g. 4 cores)
- `aws.lambda.enhanced.cpu_min_utilization` - CPU utilization on the least utilized core, as a normalized percent (e.g. 10%)
- `aws.lambda.enhanced.cpu_max_utilization` - CPU utilization on the most utilized core, as a normalized percent (e.g. 80%)

#### Additional Notes
- The CPU time metrics are sent after the `PlatformReport` event (similar to how the network metrics are being sent) to match the metric values reported by Lambda Insights 
- The CPU utilization metrics are not based off Lambda Insights and were new metrics added to the go agent. These metrics are sent after the `PlatformRuntimeDone` event (similar to how these metrics are being sent in the go agent) to avoid accounting for additional idle time after this event. 
- The CPU utilization metrics are based on this metric format: [CPU Utilization Metric Representation](https://docs.google.com/document/d/1wpoaYlVM8ntHTsYTcYHFr3pCxq5l2_ynpp_Pg92LZqA/edit#heading=h.rfc7e5wnt2p)

### Testing

- [Build the lambda extension](https://datadoghq.atlassian.net/wiki/spaces/SLS/pages/2723611380/Lambda+Extension#Manual-Testing)
- Invoke the function
- [Compare with lambda insight metrics](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Lambda-Insights-view-metrics.html) (for the CPU time metrics)
- Can verify the CPU utilization metrics by comparing the numbers of an invocation using the go agent vs. an invocation using the lambda extension to check that they are around the same range for that function


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -76,12 +77,21 @@ impl Processor {
...
     }
 
-    /// Given a `request_id`, add the enhanced metric offsets to the context buffer.
+    /// Given a `request_id`, creates the context and adds the enhanced metric offsets to the context buffer.
     ///
-    pub fn on_invoke_event(&mut self, request_id: String) {
-        let network_offset: Option<NetworkData> = proc::get_network_data().ok();
-        self.context_buffer
-            .add_network_offset(&request_id, network_offset);
+    pub fn on_invoke_event(&mut self, request_id: String, collect_enhanced_data: bool) {
```
duncanista, line 82: Why pass a config which can be inferred from the constructor?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -462,6 +463,11 @@ async fn extension_loop_active(
                                             stats_flusher.manual_flush()
                                         );
                                     }
+
+                                    if let Some(offsets) = enhanced_metric_data {
+                                        lambda_enhanced_metrics.set_cpu_utilization_enhanced_metrics(offsets.cpu_offset, offsets.uptime_offset);
+                                    }
+
                                     break;
```
duncanista, line 471: Why all the way here? It looks unnecessary to do this, what's the difference between doing it here, and all the way where its created?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -476,11 +482,13 @@ async fn extension_loop_active(
...
-                                        lambda_enhanced_metrics.set_network_enhanced_metrics(network_offset);
+                                    let (post_runtime_duration_ms, enhanced_metric_data) = p.on_platform_report(&request_id, metrics.duration_ms);
+                                    if let Some(duration) = post_runtime_duration_ms {
+                                        lambda_enhanced_metrics.set_post_runtime_duration_metric(duration);
+                                    }
+                                    if let Some(offsets) = enhanced_metric_data {
+                                        lambda_enhanced_metrics.set_network_enhanced_metrics(offsets.network_offset);
+                                        lambda_enhanced_metrics.set_cpu_time_enhanced_metrics(offsets.cpu_offset);
                                     }
                                     drop(p);
```
duncanista, line 493: Not sure I understand why are we not doing the some as before?

bottlecap/src/proc/clock.rs
```diff
@@ -0,0 +1,21 @@
+use libc;
+use std::io;
```
duncanista, line 2: What's the justification on this crate? Why `libc` clock and not the standard clock?

bottlecap/src/proc/mod.rs
```diff
@@ -58,6 +60,122 @@ fn get_network_data_from_path(path: &str) -> Result<NetworkData, io::Error> {
...
+
+pub fn get_uptime() -> Result<f64, io::Error> {
+    get_uptime_from_path(PROC_UPTIME_PATH)
+}
+
+fn get_uptime_from_path(path: &str) -> Result<f64, io::Error> {
+    let file = File::open(path)?;
+    let reader = io::BufReader::new(file);
+
+    if let Some(Ok(line)) = reader.lines().next() {
```
duncanista, line 155: not sure we want to have a `Some(Ok(...))` is there a better way to handle this?

duncanista (commented):
Left some comments

bottlecap/src/bin/bottlecap/main.rs
shreyamalpani: When these metrics are sent after the PlatformReport event it includes a lot of idle time after the actual invocation, so the cpu utilization ratio becomes very low relative to the idle time. I tested this out with a function, and for the first invocation it was showing around 10% but for all the subsequent invocations it reported varied numbers sometimes less than 1% because of the idle time between invocations. When I tried out sending after the PlatformRuntimeDone event, it consistently reported around 20%, which is similar to the numbers in the go agent as well.

bottlecap/src/bin/bottlecap/main.rs
shreyamalpani: I realized that we were returning None for both `post_runtime_duration_ms` and `enhanced_metric_data` if we could not calculate `post_runtime_duration_ms` because of `context.runtime_duration_ms` being 0. Changed it to handle so that even if we can't calculate `post_runtime_duration_ms`, we still try to send enhanced metrics.

bottlecap/src/proc/mod.rs
shreyamalpani: Yes I'll change it to be more similar the other methods

bottlecap/src/proc/mod.rs
```diff
@@ -58,6 +60,122 @@ fn get_network_data_from_path(path: &str) -> Result<NetworkData, io::Error> {
...
+        individual_cpu_idle_times: HashMap::new(),
+    };
+
+    // SC_CLK_TCK is the system clock frequency in ticks per second
+    // We'll use this to convert CPU times from user HZ to milliseconds
+    let clktck = clock::get_clk_tck()? as f64;
+
+    for line in reader.lines() {
+        let line = line?;
+        let mut values = line.split_whitespace();
```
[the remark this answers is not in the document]
shreyamalpani: I get an error that the temporary value of `line?` gets dropped while borrowing and to use a let binding

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -150,18 +150,28 @@ impl ContextBuffer {
...
-            self.insert(Context::new(request_id.clone(), 0.0, 0.0, 0, network_data));
+            self.insert(Context::new(
+                request_id.clone(),
+                0.0,
+                0.0,
+                0,
+                enhanced_metric_data,
+            ));
         }
     }
```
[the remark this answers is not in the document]
shreyamalpani: For sure, will open that PR next

bottlecap/Cargo.toml
```diff
@@ -19,6 +19,7 @@ dogstatsd = { git = "https://github.com/DataDog/libdatadog", branch = "main" }
 figment = { version = "0.10", default-features = false, features = ["yaml", "env"] }
 hyper = { version = "0.14", default-features = false, features = ["server"] }
 lazy_static = { version = "1.5", default-features = false }
+libc = "0.2"
```
alexgallotta, line 22: Can we use the same format of other lines and specify the default-features=false? Or the minimum we need?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -462,6 +463,11 @@ async fn extension_loop_active(
                                             stats_flusher.manual_flush()
                                         );
                                     }
+
+                                    if let Some(offsets) = enhanced_metric_data {
```
alexgallotta, line 467: This seems to be done after the flushing, a few lines above. It seems that potentially we will miss flushing these metrics.
Is it intended?

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -337,19 +350,40 @@ mod tests {
...
+            total_user_time_ms: 100.0,
+            total_system_time_ms: 53.0,
+            total_idle_time_ms: 20.0,
+            individual_cpu_idle_times: individual_cpu_idle_times,
+        });
+
+        let uptime_offset = Some(50.0);
+
+        let enhanced_metric_data = Some(EnhancedMetricData {
+            network_offset: network_offset,
```
alexgallotta, line 366: no need to repeat the names,
```
network_offset: network_offset,
```
can become
```
network_offset,
```

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -425,8 +425,9 @@ async fn extension_loop_active(
                                     ..
                                 } => {
                                     let mut p = invocation_processor.lock().await;
+                                    let mut enhanced_metric_data = None;
                                     if let Some(metrics) = metrics {
-                                        p.on_platform_runtime_done(
+                                        enhanced_metric_data = p.on_platform_runtime_done(
                                             &request_id,
                                             metrics.duration_ms,
                                             config.clone(),
```
alexgallotta, line 433: all these stuff like configs and tags provider could be set once at initialization of the invocation_processor, rather than cloned continuously

shreyamalpani: True, I'll be doing some refactoring in my next PR, so I can look into that then

bottlecap/src/proc/mod.rs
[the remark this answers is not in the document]
alexgallotta: As a tip for next time, usually this blocks of code could be written in a functional way. Asking chatgpt, the refactor is instantaneous:
```
let cpu_data = reader.lines()
    .filter_map(|line| line.ok())
    .filter_map(|line| {
        let mut values = line.split_whitespace();
        values.next().map(|label| (label, values))
    })
    .fold(CPUData {
        total_user_time_ms: 0.0,
        total_system_time_ms: 0.0,
        total_idle_time_ms: 0.0,
        individual_cpu_idle_times: HashMap::new(),
    }, |mut cpu_data, (label, mut values)| {
        match label {
            "cpu" => {
                let user: Option<f64> = values.next().and_then(|s| s.parse().ok());
                values.next(); // skip "nice"
                let system: Option<f64> = values.next().and_then(|s| s.parse().ok());
                let idle: Option<f64> = values.next().and_then(|s| s.parse().ok());

                if let (Some(user_val), Some(system_val), Some(idle_val)) = (user, system, idle) {
                    cpu_data.total_user_time_ms = (1000.0 * user_val) / clktck;
                    cpu_data.total_system_time_ms = (1000.0 * system_val) / clktck;
                    cpu_data.total_idle_time_ms = (1000.0 * idle_val) / clktck;
                }
            }
            label if label.starts_with("cpu") => {
                let idle: Option<f64> = values.nth(3).and_then(|s| s.parse().ok());

                if let Some(idle_val) = idle {
                    cpu_data.individual_cpu_idle_times.insert(label.to_string(), (1000.0 * idle_val) / clktck);
                }
            }
            _ => {}
        }
        cpu_data
    });
```

the pros are:
- usually more legible (one filter/action at time)
- more robust (fn once ensure things are used just once)
- it's a 0 cost abstraction
- less indentation

bottlecap/src/proc/clock.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#[allow(clippy::cast_sign_loss)]
+#[cfg(not(target_os = "windows"))]
+pub fn get_clk_tck() -> Result<u64, io::Error> {
+    let clk_tck = unsafe { libc::sysconf(libc::_SC_CLK_TCK) };
```
alexgallotta, line 6: does this change over time during an invocation? Or can we read it only once at the start?

duncanista, line 6: I'm kinda not sure if this code is going to cause problems, we are only checking if it's -1, there's no comments here for unexpected/other behavior. I'm blocking the PR so we can talk about this in person.

duncanista (changes requested):
Blocking so we don't merge this and can talk more about it in person, would like to see if there are other alternatives we could try.

Let's ask Rust experts what we can do around the clock code and thoroughly test it with an alpine container too!

bottlecap/src/proc/mod.rs
```diff
@@ -58,6 +60,123 @@ fn get_network_data_from_path(path: &str) -> Result<NetworkData, io::Error> {
...
+                let user: Option<f64> = values.next().and_then(|s| s.parse().ok());
+                values.next(); // skip "nice"
+                let system: Option<f64> = values.next().and_then(|s| s.parse().ok());
+                let idle: Option<f64> = values.next().and_then(|s| s.parse().ok());
+
+                match (user, system, idle) {
+                    (Some(user_val), Some(system_val), Some(idle_val)) => {
+                        cpu_data.total_user_time_ms = (1000.0 * user_val) / clktck;
+                        cpu_data.total_system_time_ms = (1000.0 * system_val) / clktck;
+                        cpu_data.total_idle_time_ms = (1000.0 * idle_val) / clktck;
```
duncanista, line 106: Also here

bottlecap/src/proc/clock.rs
```diff
@@ -0,0 +1,21 @@
+use nix::unistd::sysconf;
+use nix::unistd::SysconfVar;
```
duncanista, line 2: merge import

bottlecap/src/proc/clock.rs
```diff
@@ -0,0 +1,21 @@
...
+            "Could not find system clock ticks per second",
+        )),
+    }
+}
+
+#[cfg(target_os = "windows")]
+pub fn get_clk_tck() -> Result<u64, io::Error> {
+    // Windows does not have this concept
+    Ok(1)
+}
```
duncanista, line 21: Do we still need the same type of constructor here now that you're not using the `libc` crate?

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -166,6 +166,177 @@ impl Lambda {
...
+        }
+
+        // Maximally utilized CPU is the one with the least time spent in the idle process
+        let cpu_max_utilization = ((uptime - min_idle_time) / uptime) * 100.0;
+        // Minimally utilized CPU is the one with the most time spent in the idle process
+        let cpu_min_utilization = ((uptime - max_idle_time) / uptime) * 100.0;
+
+        let cpu_total_utilization_decimal =
+            ((uptime * num_cores) - total_idle_time) / (uptime * num_cores);
+        let cpu_total_utilization_pct = cpu_total_utilization_decimal * 100.0;
```
duncanista, line 261: Where does the `100.0` come from

bottlecap/src/proc/clock.rs
shreyamalpani: The nix package is a safer wrapper to the libc crate, so we would still need a similar structure

duncanista (approved):

bottlecap/src/proc/clock.rs
duncanista: Cool, just want to make sure if we still needed the windows os, the clippy cast sign loss allow and etc

PR #433, opened 2024-11-04
Refactor context creation

shreyamalpani:
### What does this PR do?
Creates the context for a request_id on receiving the Invoke event, and removes this logic from other events since the Invoke event is always received first.

### DIFF
--- a/bottlecap/src/lifecycle/invocation/context.rs
+++ b/bottlecap/src/lifecycle/invocation/context.rs
@@ -255,15 +255,6 @@ mod tests {
 
         buffer.add_init_duration(&request_id, 100.0);
         assert_eq!(buffer.get(&request_id).unwrap().init_duration_ms, 100.0);
-
-        // Add init duration to a context that doesn't exist
-        let unexistent_request_id = String::from("unexistent");
-        buffer.add_init_duration(&unexistent_request_id, 200.0);
-        assert_eq!(buffer.size(), 2);
-        assert_eq!(
-            buffer.get(&unexistent_request_id).unwrap().init_duration_ms,
-            200.0
-        );
     }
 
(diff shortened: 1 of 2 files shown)


bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -93,8 +93,14 @@ impl ContextBuffer {
             .find(|context| context.request_id == *request_id)
     }
 
-    /// Adds the init duration to a `Context` in the buffer. If the `Context` is not found, a new
-    /// `Context` is created and added to the buffer.
+    /// Creates a new `Context` and adds it to the buffer.
+    ///
+    #[allow(clippy::ptr_arg)]
```
duncanista, line 98: Why do we need this?

shreyamalpani: I could change it to &str instead of &String, but I wasn't sure if it is better to keep it consistent with the other functions using &String

```
pub fn create_context(&mut self, request_id: &str) {
        self.insert(Context::new(request_id.to_owned(), 0.0, 0.0, 0, None));
 }
```

duncanista: I just don't think it's worth passing a reference if we are gonna clone inside

duncanista: It doesn't have to be consistent because in the other ones you're searching, and in this one you're keeping the value.

duncanista (approved):

bottlecap/src/lifecycle/invocation/context.rs
shreyamalpani: I see, thanks!

PR #443, opened 2024-11-11
feat(bottlecap): generate tmp enhanced metrics

shreyamalpani:
### What does this PR do?
This PR introduces three new enhanced lambda metrics. These metrics are emitted once per invocation and represent space used and space available in the `/tmp` directory.

The three new metrics are:
- `aws.lambda.enhanced.tmp_max` - max available bytes in `/tmp`
- `aws.lambda.enhanced.tmp_used` - bytes used or reserved in `/tmp`
- `aws.lambda.enhanced.tmp_free` - bytes available in `/tmp`

- Build the lambda extension
- Invoke the function

### Additional Notes
- `tmp_used` reflects a peak over the course of the request. Suppose 2000 bytes are written to `/tmp`, then deleted, then 3000 bytes are written, then deleted. The metric will report 3000.

### DIFF
--- a/bottlecap/Cargo.toml
+++ b/bottlecap/Cargo.toml
@@ -20,7 +20,7 @@ figment = { version = "0.10", default-features = false, features = ["yaml", "env
 hyper = { version = "0.14", default-features = false, features = ["server"] }
 lazy_static = { version = "1.5", default-features = false }
 log = { version = "0.4", default-features = false }
-nix = { version = "0.26", default-features = false, features = ["feature"] }
+nix = { version = "0.26", default-features = false, features = ["feature", "fs"] }
 protobuf = { version = "3.5", default-features = false }
 regex = { version = "1.10", default-features = false }
 reqwest = { version = "0.12", features = ["json", "http2", "rustls-tls"], default-features = false }
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -62,8 +62,8 @@ use std::{
     sync::{Arc, Mutex},
 };
 use telemetry::listener::TelemetryListenerConfig;
+use tokio::sync::mpsc::Sender;
 use tokio::sync::Mutex as TokioMutex;
-use tokio::sync::{mpsc::Sender, watch};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error};
 use tracing_subscriber::EnvFilter;
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -60,11 +60,11 @@ use std::{
     os::unix::process::CommandExt,
     path::Path,
     process::Command,
-    sync::{mpsc, Arc, Mutex},
+    sync::{Arc, Mutex},
 };
 use telemetry::listener::TelemetryListenerConfig;
-use tokio::sync::mpsc::Sender;
 use tokio::sync::Mutex as TokioMutex;
+use tokio::sync::{mpsc::Sender, watch};
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error};
 use tracing_subscriber::EnvFilter;
@@ -375,7 +375,7 @@ async fn extension_loop_active(
                 lambda_enhanced_metrics.increment_invocation_metric();
 
                 // Start a channel for monitoring tmp enhanced data
-                let (tmp_chan_tx, tmp_chan_rx) = mpsc::channel::<bool>();
+                let (tmp_chan_tx, tmp_chan_rx) = watch::channel(());
                 lambda_enhanced_metrics.set_tmp_enhanced_metrics(tmp_chan_rx);
 
                 let mut p = invocation_processor.lock().await;
@@ -457,9 +457,9 @@ async fn extension_loop_active(
                                     // set cpu utilization metrics here to avoid accounting for extra idle time
                                     if let Some(mut data) = enhanced_metric_data {
                                         lambda_enhanced_metrics.set_cpu_utilization_enhanced_metrics(data.cpu_offset, data.uptime_offset);
-                                        // Drop tmp_chan as a signal to stop monitoring tmp
+                                        // Send signal to stop monitoring tmp
                                         if let Some(tmp_chan) = data.tmp_chan.take() {
-                                            _ = tmp_chan.send(false);
+                                            _ = tmp_chan.send(());
                                             drop(tmp_chan);
                                         }
                                     }
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -379,7 +379,7 @@ async fn extension_loop_active(
                 lambda_enhanced_metrics.set_tmp_enhanced_metrics(tmp_chan_rx);
 
                 let mut p = invocation_processor.lock().await;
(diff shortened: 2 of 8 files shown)


bottlecap/Cargo.toml
```diff
@@ -20,7 +20,7 @@ figment = { version = "0.10", default-features = false, features = ["yaml", "env
 hyper = { version = "0.14", default-features = false, features = ["server"] }
 lazy_static = { version = "1.5", default-features = false }
 log = { version = "0.4", default-features = false }
-nix = { version = "0.26", default-features = false, features = ["feature"] }
+nix = { version = "0.26", default-features = false, features = ["feature", "fs"] }
```
duncanista, line 23: happy to see that this crate allows us to do more!

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -375,8 +373,13 @@ async fn extension_loop_active(
                     request_id, deadline_ms, invoked_function_arn
                 );
                 lambda_enhanced_metrics.increment_invocation_metric();
+
+                // Start a channel for monitoring tmp enhanced data
+                let (tmp_chan_tx, tmp_chan_rx) = watch::channel(());
```
duncanista, line 378: Why a `watch` channel?

bottlecap/src/metrics/enhanced/statfs.rs
```diff
@@ -0,0 +1,20 @@
+#![allow(clippy::module_name_repetitions)]
+
+use nix::sys::statfs::statfs;
+use std::io;
+use std::path::Path;
+
+#[cfg(not(target_os = "windows"))]
+pub fn statfs_info(path: &str) -> Result<(f64, f64, f64), io::Error> {
```
duncanista, line 8: Add a comment of what this does

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -81,7 +81,7 @@ impl Processor {
 
     /// Given a `request_id`, creates the context and adds the enhanced metric offsets to the context buffer.
     ///
-    pub fn on_invoke_event(&mut self, request_id: String) {
+    pub fn on_invoke_event(&mut self, request_id: String, tmp_chan: Option<watch::Sender<()>>) {
```
duncanista, line 84: Why an `Option` if we are clearly creating it?

bottlecap/src/bin/bottlecap/main.rs
duncanista: Do we need to bound this channel?

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -343,6 +349,86 @@ impl Lambda {
...
+                Ok(stats) => stats,
+                Err(err) => {
+                    debug!("Could not emit tmp enhanced metrics. {:?}", err);
+                    return;
+                }
+            };
+            let tmp_max = bsize * blocks;
+            let mut tmp_used = bsize * (blocks - bavail);
+
+            let mut interval = interval(Duration::from_millis(10));
```
duncanista, line 405: Why 10ms?

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -343,6 +349,86 @@ impl Lambda {
...
+            loop {
+                tokio::select! {
```
duncanista, line 407: Not sure I understand why are we calculating this every 10ms, why not at the start/end as we have been doing for the other metrics?

shreyamalpani: The previous metrics were cumulative in a sense, so we wanted to take the difference between the start and end to get the values for the invocation. For tmp (and the following fd and thread metrics), we can only read the current state, so we need to keep checking periodically to capture the maximum value for the metric over the invocation

bottlecap/src/bin/bottlecap/main.rs
shreyamalpani: Used a watch channel because it holds a single value and we only care about receiving one value - the stop signal

bottlecap/src/metrics/enhanced/lambda.rs
shreyamalpani: This was the interval used in the Go agent so left it the same, a less frequent ticker may not capture the tmp data accurately

bottlecap/src/metrics/enhanced/lambda.rs
duncanista: Would a very fast API (<5ms) be blocked (has to wait for this to finish) by these metrics?

shreyamalpani: This shouldn't block the main thread because we aren't awaiting `set_tmp_enhanced_metrics` to complete and instead using the channel/signal to determine when to generate the metrics, and in that case it would send the metrics with the initialized values of `tmp_max` and `tmp_used`

bottlecap/src/bin/bottlecap/main.rs
duncanista: Amazing

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -90,13 +90,20 @@ impl Processor {
     pub fn on_invoke_event(&mut self, request_id: String) {
         self.context_buffer.create_context(request_id.clone());
         if self.enhanced_metrics_enabled {
+            // Collect offsets for network and cpu metrics
             let network_offset: Option<NetworkData> = proc::get_network_data().ok();
             let cpu_offset: Option<CPUData> = proc::get_cpu_data().ok();
             let uptime_offset: Option<f64> = proc::get_uptime().ok();
+
+            // Start a channel for monitoring tmp enhanced data
+            let (tmp_chan_tx, tmp_chan_rx) = watch::channel(());
```
duncanista, line 99:
```suggestion
            let (tmp_chan_tx, _) = watch::channel(());
```

Do we use `rx`?

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -90,13 +90,20 @@ impl Processor {
...
+            self.enhanced_metrics.set_tmp_enhanced_metrics(tmp_chan_rx);
+
             let enhanced_metric_offsets = Some(EnhancedMetricData {
                 network_offset,
                 cpu_offset,
                 uptime_offset,
+                tmp_chan: tmp_chan_tx,
```
duncanista, line 106: maybe let's name it `tmp_chan_tx`?

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -343,6 +349,86 @@ impl Lambda {
...
+                    biased;
+                    // When the stop signal is received, generate final metrics
+                    _ = send_metrics.changed() => {
+                        let mut aggr: std::sync::MutexGuard<Aggregator> =
+                            aggr.lock().expect("lock poisoned");
+                        Self::generate_tmp_enhanced_metrics(tmp_max, tmp_used, &mut aggr);
+                        return;
```
duncanista, line 414: Does this return break the loop?

duncanista (approved):
Left some comments, but approving in the meantime!

Looks good!

bottlecap/src/metrics/enhanced/lambda.rs
shreyamalpani: Yup!

PR #446, opened 2024-11-11
chore(bottlecap): move `EnhancedMetrics` to live in `InvocationProcessor`

duncanista:
# What?

Moves where enhanced metrics are being called to be aggregated.

# Why?

Reduce the footprint in `main.rs`, and to have more control over the whole invocation lifecycle.

Also unblocks #444


shreyamalpani (commented):
LGTM!

shreyamalpani (approved):

PR #449, opened 2024-11-12
feat(bottlecap): add Step Functions trace extraction

duncanista:
# What?

Allows Step Function invokes to Lambda to appear in the same trace

Screenshot 2024-11-12 at 4 24 43 PM
Screenshot 2024-11-12 at 4 25 03 PM

# How

Generating trace context data through deterministic methods

# Notes

- Step Functions don't need an inferred span, so we remove it in case a generated_span_context exists


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -23,7 +23,13 @@ use crate::{
     telemetry::events::{ReportMetrics, Status},
     traces::{
         context::SpanContext,
-        propagation::{DatadogCompositePropagator, Propagator},
+        propagation::{
+            text_map_propagator::{
+                DatadogHeaderPropagator, DATADOG_PARENT_ID_KEY, DATADOG_SPAN_ID_KEY,
+                DATADOG_TRACE_ID_KEY,
+            },
+            DatadogCompositePropagator, Propagator,
```
purple4reina, line 31: At point in the invocation is this happening? Is this extracting context from the inbound event? Or is it getting it from the start invocation request headers?

Just asking cuz it doesn't look like we have support for w3c headers. Is that correct?

duncanista: I'm not sure which line you're talking about, you left the comment in the import

duncanista: > Just asking cuz it doesn't look like we have support for w3c headers. Is that correct?
We do

purple4reina: Ahhh, didn't realize this is just the import. Thanks for answering the question!

bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    fn get_arn(&self, _region: &str) -> String {
+        if let Some(sm) = &self.state_machine {
+            return sm.id.clone();
+        }
+
+        String::new()
+    }
+
+    fn get_carrier(&self) -> HashMap<String, String> {
```
duncanista, line 102: No carrier, we generate the span context. The decision behind this is that the customer might be using W3C or other style, and we cannot just generate a carrier based on that, its easier to just return a `SpanContext`

bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
```diff
@@ -0,0 +1,366 @@
...
+                "1914fe7789eb32be".to_string(),
+            )]),
+            links: vec![],
+        };
+
+        assert_eq!(span_context, expected);
+    }
+
+    #[test]
+    fn test_generate_parent_id() {
```
duncanista, line 326: These tests are the ones from the Go Agent, so we can see that data is generated as expected.

agocs (approved):
LGTM

PR #450, opened 2024-11-12
feat(bottlecap): add Cold Start Span + Tags

duncanista:
# What?

- Creates an `aws.lambda.cold_start` span on Cold Starts (or proactive inits), adds the correct tags to the `aws.lambda` span for cold starts.
- Adds tags for cold start to enhanced metrics

image
image

# How?

- Calculate cold start span given the telemetry API events.
- Calculate the cold start / proactive init based on the invoke.
- Adding a set of dynamic value tags for enhanced metrics, so we can append the cold start or proactive init tags.

# Notes
- `ContextBuffer` now doesn't remove on `on_platform_report`, instead, we held all invocations and await for the 6th one to remove some space. This is so we can calculate properly the tags on cold start.


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -255,7 +347,7 @@ impl Processor {
         // Set the report log metrics
         self.enhanced_metrics.set_report_log_metrics(&metrics);
 
-        if let Some(context) = self.context_buffer.remove(request_id) {
+        if let Some(context) = self.context_buffer.get(request_id) {
```
duncanista, line 349: Doing a `get` since I wanna make sure that only when the buffer is empty, we can calculate the init tags correctly

alexgallotta (approved):

PR #452, opened 2024-11-13
feat(bottlecap): add aws trace header for java and sqs

alexgallotta:
image

### DIFF
--- a/bottlecap/src/lifecycle/invocation/span_inferrer.rs
+++ b/bottlecap/src/lifecycle/invocation/span_inferrer.rs
@@ -17,7 +17,7 @@ use crate::lifecycle::invocation::{
         lambda_function_url_event::LambdaFunctionUrlEvent,
         s3_event::S3Record,
         sns_event::{SnsEntity, SnsRecord},
-        sqs_event::SqsRecord,
+        sqs_event::{SqsRecord, extract_trace_context_from_aws_trace_header},
         step_function_event::StepFunctionEvent,
         Trigger, FUNCTION_TRIGGER_EVENT_SOURCE_ARN_TAG,
     },
--- a/bottlecap/src/lifecycle/invocation/span_inferrer.rs
+++ b/bottlecap/src/lifecycle/invocation/span_inferrer.rs
@@ -35,6 +35,8 @@ pub struct SpanInferrer {
     is_async_span: bool,
     // Carrier to extract the span context from
     carrier: Option<HashMap<String, String>>,
+    // context taken from `AWSTraceHeader` when java->sqs->java
+    aws_header_context: Option<SpanContext>,
     // Generated Span Context from Step Functions
     generated_span_context: Option<SpanContext>,
     // Tags generated from the trigger
@@ -50,6 +52,7 @@ impl SpanInferrer {
             wrapped_inferred_span: None,
             is_async_span: false,
             carrier: None,
+            aws_header_context: None,
             generated_span_context: None,
             trigger_tags: None,
         }
@@ -95,6 +98,11 @@ impl SpanInferrer {
         } else if SqsRecord::is_match(payload_value) {
             if let Some(t) = SqsRecord::new(payload_value.clone()) {
                 t.enrich_span(&mut inferred_span, &self.service_mapping);
+
+                self.aws_header_context = extract_trace_context_from_aws_trace_header(
+                    t.attributes.aws_trace_header.clone(),
+                )
+                .ok();
 
                 // Check for SNS event wrapped in the SQS body
                 if let Ok(sns_entity) = serde_json::from_str::<SnsEntity>(&t.body) {
@@ -297,6 +305,10 @@ impl SpanInferrer {
         // Step Functions `SpanContext` is deterministically generated
         if let Some(sc) = &self.generated_span_context {
             return Some(sc.clone());
+        }
+
+        if self.aws_header_context.clone().is_some() {
+            return self.aws_header_context.clone();
         }
 
         if let Some(sc) = self.carrier.as_ref().and_then(|c| propagator.extract(c)) {
--- a/bottlecap/src/lifecycle/invocation/span_inferrer.rs
+++ b/bottlecap/src/lifecycle/invocation/span_inferrer.rs
@@ -35,9 +35,7 @@ pub struct SpanInferrer {
     is_async_span: bool,
     // Carrier to extract the span context from
     carrier: Option<HashMap<String, String>>,
-    // context taken from `AWSTraceHeader` when java->sqs->java
-    aws_header_context: Option<SpanContext>,
-    // Generated Span Context from Step Functions
+    // Generated Span Context from Step Functions or context taken from `AWSTraceHeader` when java->sqs->java
     generated_span_context: Option<SpanContext>,
     // Tags generated from the trigger
     trigger_tags: Option<HashMap<String, String>>,
@@ -52,7 +50,6 @@ impl SpanInferrer {
             wrapped_inferred_span: None,
             is_async_span: false,
             carrier: None,
-            aws_header_context: None,
             generated_span_context: None,
             trigger_tags: None,
         }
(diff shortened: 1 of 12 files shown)


bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -52,6 +54,7 @@ impl SpanInferrer {
             wrapped_inferred_span: None,
             is_async_span: false,
             carrier: None,
+            aws_header_context: None,
```
duncanista, line 57: Instead of creating a new field, can't you you use the same logic StepFunctions is doing?

bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
```diff
@@ -198,6 +202,81 @@ impl Trigger for SqsRecord {
...
+    );
+    Ok(SpanContext {
+        trace_id,
+        span_id: parent_id,
+        sampling: Some(Sampling {
+            priority: Some(sampling_priority),
+            mechanism: None,
+        }),
+        origin: None,
+        tags: HashMap::new(),
```
duncanista, line 275: How are you handling 128 bits here?

bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
```diff
@@ -198,6 +202,81 @@ impl Trigger for SqsRecord {
...
+            }
+        } else if part.starts_with("Sampled=") && sampled.is_empty() {
+            sampled = part[8..].to_string();
+        }
+
+        if !trace_id.is_empty() && !parent_id.is_empty() && !sampled.is_empty() {
+            break;
+        }
+        start_part = end_part + 1;
+    }
```
duncanista, line 247: Can't we just do a simpler algo?
```Root=1-64cc2edd-112fbf1701d1355973a11d57;Parent=7d5a9776024b2d42;Sampled=0```
Split by `;`
- `Root=1-64cc2edd-112fbf1701d1355973a11d57`
- `Parent=7d5a9776024b2d42`
- `Sampled=0`

Then split by `=`?

File: bottlecap/src/lifecycle/invocation/span_inferrer.rs (hunk 1 above)
Comment (alexgallotta) [reply bottlecap/src/lifecycle/invocation/span_inferrer.rs:57]: ok I will reuse the same var

File: bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs (hunk 3 above)
Comment (alexgallotta) [reply bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs:247]: I chatgpt-translated the algorithm used in go.
I am not sure if there are side effects of using an easier split. I would argue a regexp would be much better and faster, but can you confirm that the format is a `;` concatenations of those 3 headers?

File: bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs (hunk 2 above)
Comment (alexgallotta) [reply bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs:271]: I didn't see any code handling 128 bits for the aws header. I think the datadog-specific format doesn't require 128 bits.
@purple4reina do you have context on this?

File: bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs (hunk 3 above)
Comment (alexgallotta) [reply bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs:244]: the logic seems to be to parse all the `;` separated segments until all 3 variables are set. I did not want to spend too much time thinking on side effects

File: bottlecap/src/lifecycle/invocation/span_inferrer.rs (hunk 1 above)
Comment (alexgallotta) [reply bottlecap/src/lifecycle/invocation/span_inferrer.rs]: I had to add a boolean since step functions are ignoring inferred spans.
I would honestly prefer the 2 span approach since it's more explicit and easier to understand.
But I will go with the boolean approach for now

File: bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs (hunk 2 above)
Comment (agocs) [reply bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs:271]: The upstream tracer may include the upper 64 bits of a trace ID. e.g. https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsInterceptor.java#L44

We have a backwards compatibility layer that allows flame graphs to connect 128 bit trace chunks to 64 bit trace chunks given that the lower 64 bits match, but the right thing to do (whether now or in a fast-follow item) is to support 128 bit trace IDs.

File: bottlecap/src/lifecycle/invocation/span_inferrer.rs [hunk 4]
```diff
@@ -76,6 +76,8 @@ impl SpanInferrer {
             ..Default::default()
         };
 
+        let mut is_step_function = false;
```
duncanista, line 79: Can't we just downcast into the StepFunctions event instead of having a boolean? That way at the very bottom you can just check if the trigger is StepFunction

bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
```diff
@@ -196,6 +196,81 @@ impl Trigger for SqsRecord {
...
+
+    fn is_async(&self) -> bool {
+        true
+    }
+}
+
+// extractTraceContextfromAWSTraceHeader extracts trace context from the
+// AWSTraceHeader directly. Unlike the other carriers in this file, it should
+// not be passed to the tracer.Propagator, instead extracting context directly.
+pub(crate) fn extract_trace_context_from_aws_trace_header(
```
duncanista, line 208: Why a crate method and not just a method for the struct? We are not using it anywhere else

alexgallotta: I prefer the flexibility and not adding extra restrictions that aren't needed. It makes testing and changing code simpler.
 I also think a lot of the structs and traits should go away, so would rather avoid adding code and complexity to keep doing that

bottlecap/src/lifecycle/invocation/span_inferrer.rs
alexgallotta: Not sure what you mean

bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
alexgallotta: Thanks!

bottlecap/src/lifecycle/invocation/span_inferrer.rs
duncanista: instead of having a boolean, where you have this thing, just try downcasting the boxed trigger into a stepfunctions event, if it succeeds, then its guaranteed to be a stepfunction event

bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
```diff
@@ -196,9 +196,87 @@ impl Trigger for SqsRecord {
...
+
+    let sampling_priority = i8::from(sampled == "1");
+
+    let mut trace_context = HashMap::new();
+    trace_context.insert("trace_id".to_string(), trace_id.to_string());
+    trace_context.insert("parent_id".to_string(), parent_id.to_string());
+    trace_context.insert(
+        "sampling_priority".to_string(),
+        sampling_priority.to_string(),
+    );
```
duncanista, line 262: What is this being used for?

bottlecap/src/lifecycle/invocation/span_inferrer.rs
alexgallotta: can you give me an example? I don't think it's doable or considered a good approach in rust

bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
alexgallotta: oh that's dead code from before, ty

PR #453, opened 2024-11-13
feat(bottlecap): generate file descriptor and threads enhanced metrics

shreyamalpani:
### What does this PR do?
This PR introduces four new enhanced lambda metrics. These metrics are emitted once per invocation and represent file descriptor and threads data across all running processes.

The two file descriptor metrics are:
- `aws.lambda.enhanced.fd_max` - maximum number of file descriptors that can be opened by the function
- `aws.lambda.enhanced.fd_use` - maximum number of file descriptors open at once over the duration of the function

The two threads metrics are:
- `aws.lambda.enhanced.threads_max` - maximum number of threads that can be used by the function
- `aws.lambda.enhanced.threads_use` - maximum number of threads in use at once over the duration of the function

- Build the lambda extension
- Invoke the function

### Additional Notes
- `fd_max` and `threads_max` both default to send 1024, even if we are not able to read this data, because that is the AWS Lambda limit of max open files and max processes
- This implementation of `fd_use` and `threads_use` considers all running processes, whereas Lambda Insights only considers the process of the lambda itself, so these metrics will always be higher than the numbers reported by Lambda Insights
- `fd_use` reflects a peak over the course of the request. Suppose 200 file descriptors are opened, then deleted, then 300 are opened, then deleted. The metric will report 300. 
- `threads_use` also reflects a peak over the course of the request. Suppose 20 threads are created, then closed, then 30 are created, then closed. The metric will report 30.


bottlecap/src/proc/mod.rs
```diff
@@ -180,6 +211,135 @@ fn get_uptime_from_path(path: &str) -> Result<f64, io::Error> {
     ))
 }
 
+#[must_use]
+pub fn get_fd_max_data(pids: &[i32]) -> f64 {
+    get_fd_max_data_from_path(PROC_PATH, pids)
+}
+
+fn get_fd_max_data_from_path(path: &str, pids: &[i32]) -> f64 {
+    let mut fd_max: f64 = 1024.0; // Default to hard limit set by AWS Lambda
```
duncanista, line 220: maybe move the `1024` to be a constant?

bottlecap/tests/proc/13/.gitkeep
duncanista, line 1: Do we need this `.gitkeep` files?

bottlecap/src/proc/mod.rs
```diff
@@ -180,6 +211,135 @@ fn get_uptime_from_path(path: &str) -> Result<f64, io::Error> {
...
+            ));
+        };
+
+        for dir_entry in files.flatten() {
+            if let Ok(file_type) = dir_entry.file_type() {
+                if file_type.is_dir() {
+                    threads_use += 1;
+                }
+            }
+        }
```
duncanista, line 337: Can this be reduced into a map operation?

bottlecap/src/proc/mod.rs
alexgallotta: also, why float? The fd number should always be an int, right?

bottlecap/src/proc/mod.rs
```diff
@@ -180,6 +211,135 @@ fn get_uptime_from_path(path: &str) -> Result<f64, io::Error> {
...
+                    break;
+                }
+
+                // Skip past the first three items ("Max open files") to get the soft limit
+                let fd_max_pid_str = values[3];
+                let Ok(fd_max_pid) = fd_max_pid_str.parse() else {
+                    debug!("File descriptor max data not found in file {}", limits_path);
+                    break;
+                };
+                fd_max = fd_max.min(fd_max_pid);
```
alexgallotta, line 245: not sure I understand here: the fd_max is the minimum between current and fd_max_pid? Shouldn't it be the max?

bottlecap/src/proc/mod.rs
```diff
@@ -180,6 +211,135 @@ fn get_uptime_from_path(path: &str) -> Result<f64, io::Error> {
...
+}
+
+pub fn get_fd_use_data(pids: &[i32]) -> Result<f64, io::Error> {
+    get_fd_use_data_from_path(PROC_PATH, pids)
+}
+
+fn get_fd_use_data_from_path(path: &str, pids: &[i32]) -> Result<f64, io::Error> {
+    let mut fd_use = 0;
+
+    for pid in pids {
```
alexgallotta, line 261: why in some loops we have the reference and in others not?
It is better to have consistency

bottlecap/src/proc/mod.rs
```diff
@@ -180,6 +211,135 @@ fn get_uptime_from_path(path: &str) -> Result<f64, io::Error> {
...
+    }
+
+    fd_max
+}
+
+pub fn get_fd_use_data(pids: &[i32]) -> Result<f64, io::Error> {
+    get_fd_use_data_from_path(PROC_PATH, pids)
+}
+
+fn get_fd_use_data_from_path(path: &str, pids: &[i32]) -> Result<f64, io::Error> {
```
alexgallotta, line 258: Are pids i32 enough?
I think some years ago it was changed to be long at i64 bits, not sure which one lambda uses

bottlecap/src/proc/mod.rs
```diff
@@ -192,6 +352,21 @@ mod tests {
         safe_path.to_str().unwrap().to_string()
     }
 
+    #[test]
+    #[allow(clippy::float_cmp)]
```
alexgallotta, line 356: instead of float_cmp, you can use a comparison with < than delta. It should be safer and avoid this exception

bottlecap/src/proc/mod.rs
```diff
@@ -180,6 +211,135 @@ fn get_uptime_from_path(path: &str) -> Result<f64, io::Error> {
...
+                let values: Vec<&str> = line.split_whitespace().collect();
+
+                // Line should have 6 items: Max open files [soft limit] [hard limit] [units]
+                if values.len() < 6 {
+                    debug!("File descriptor max data not found in file {}", limits_path);
+                    break;
+                }
+
+                // Skip past the first three items ("Max open files") to get the soft limit
+                let fd_max_pid_str = values[3];
```
alexgallotta, line 240: Regexp are usually a good choice when parsing strings like this.
They are very efficient, requires less code and they are more robust than manual parsing. Chat gpt does a great job generating regexp given a string example

alexgallotta (commented):
Good work!
Left some comments on style and suggestions, but the only thing to double check is the i32 bits for pids I think

bottlecap/src/proc/mod.rs
shreyamalpani: True it is always an int, I made it a float to be consistent with all the other metrics

bottlecap/tests/proc/13/.gitkeep
shreyamalpani: I think we need them for git to recognize the empty folders

bottlecap/src/proc/mod.rs
shreyamalpani: My thought on this was to get the min of all the limits to find the most constraining limit, I'll add a comment about that I see how its confusing

bottlecap/src/proc/mod.rs
shreyamalpani: True, I think i32 might be sufficient for lambda, but I'll make it i64 just to be safe

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -429,6 +429,115 @@ impl Lambda {
...
+
+            let mut interval = interval(Duration::from_millis(2));
+            loop {
+                tokio::select! {
+                    biased;
+                    // When the stop signal is received, generate final metrics
+                    _ = send_metrics.changed() => {
+                        let mut aggr: std::sync::MutexGuard<Aggregator> =
+                            aggr.lock().expect("lock poisoned");
+                        Self::generate_fd_enhanced_metrics(fd_max, fd_use, &mut aggr);
```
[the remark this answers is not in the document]
shreyamalpani: Right, I had it in the struct because it performs similar logic of inserting the metrics as some of the other methods that use self, should I move all of the `generate_*_metrics` functions out of Lambda?

bottlecap/src/proc/mod.rs
```diff
@@ -180,27 +212,154 @@ fn get_uptime_from_path(path: &str) -> Result<f64, io::Error> {
...
+    Ok(threads_use as f64)
+}
+
 #[cfg(test)]
 #[allow(clippy::unwrap_used)]
 mod tests {
     use super::*;
     use std::path::PathBuf;
 
+    const DELTA: f64 = 1e-6;
```
duncanista, line 333: Isn't this just a precision measure? Maybe DELTA is confusing to the reader

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -429,6 +429,115 @@ impl Lambda {
...
+
+            // Set fd_max and initial value for fd_use to -1
+            let fd_max = proc::get_fd_max_data(&pids);
+            let mut fd_use = -1_f64;
+
+            // Set threads_max and initial value for threads_use to -1
+            let threads_max = proc::get_threads_max_data(&pids);
+            let mut threads_use = -1_f64;
+
+            let mut interval = interval(Duration::from_millis(1));
```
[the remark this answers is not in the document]
shreyamalpani: It's to catch the fd count more accurately in case a large number of files are opened and closed very quickly, 2ms also gets pretty close

bottlecap/src/proc/mod.rs
shreyamalpani: Hmm should I switch it back to how it was earlier?

duncanista: No, I mean, rename the variable to something like `PRECISION` its much clearer – sorry, just a nit

duncanista (approved):

PR #455, opened 2024-11-15
feat(bottlecap): support service mapping and `peer.service` tag

duncanista:
# What?

Supports `DD_SERVICE_MAPPING` as explained in our [public documentation](https://docs.datadoghq.com/serverless/aws_lambda/configuration/?tab=datadogcli#dd_service_mapping).

Also added `peer.service` tags.

# How?

Created an interface to resolve service name, its easier to read than following a builder pattern. Which is cleaner, but less readable.


alexgallotta (approved):
It looks good! Not sure what'st the purpose of `ServiceNameResolver` though, it could have been all methods.
Also, instead of wrapping extra structs and not really using polymorphism,  the span_inferrer can be in the same module of the trigger and avoid having a lot of public modules

PR #464, opened 2024-11-21
feat(bottlecap): add `runtime` tag

duncanista:
# What?

Adds the `runtime` tag to the invocation span and metric tags.

Screenshot 2024-11-21 at 10 24 06 AM

Screenshot 2024-11-21 at 10 24 36 AM

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -19,8 +19,12 @@ use crate::{
         span_inferrer::SpanInferrer, tag_span_from_value,
     },
     metrics::enhanced::lambda::{EnhancedMetricData, Lambda as EnhancedMetrics},
-    proc::{self, CPUData, NetworkData},
-    tags::provider,
+    proc::{
+        self,
+        constants::{ETC_PATH, PROC_PATH},
+        CPUData, NetworkData,
+    },
+    tags::{lambda::tags::resolve_runtime_from_proc, provider},
     telemetry::events::{ReportMetrics, Status},
     traces::{
         context::SpanContext,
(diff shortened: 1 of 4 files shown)


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -174,6 +185,13 @@ impl Processor {
                 proactive_initialization.to_string(),
             );
         }
+
+        if let Some(runtime) = &self.runtime {
+            self.span
+                .meta
+                .insert(String::from("runtime"), runtime.to_string());
+        }
```
duncanista, line 193: Not sure if we actually need this, it looks like previous versions have the `runtime` tag in spans, but not sure if this is done in backend or tracer.

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -166,6 +172,11 @@ impl Processor {
             } else {
                 cold_start = true;
             }
+
+            // Resolve runtime only once
+            let runtime = resolve_runtime_from_proc(PROC_PATH, ETC_PATH);
+            self.enhanced_metrics.set_runtime_tag(&runtime);
```
duncanista, line 178: This is definitely needed tho

bottlecap/src/tags/lambda/tags.rs
```diff
@@ -145,7 +145,7 @@ fn tags_from_env(
 }
 
 #[allow(dead_code)] // keeping this logic for when async runtime resolution will be supported
```
alexgallotta, line 147: we can remove this now, right?

duncanista: yeah

bottlecap/src/lifecycle/invocation/processor.rs
duncanista: We do need it

PR #467, opened 2024-11-22
feat(bottlecap): handle OOM exceptions

duncanista:
# What?

Handles OOM exceptions, sending enhanced metric as expected.

Screenshot 2024-11-21 at 4 23 14 PM

# Notes

Metrics look iffy, but that's OK, because when the sandbox get's signal-killed, there's no way for us to tell that we got OOM as of now.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -452,6 +452,8 @@ async fn extension_loop_active(
                                     // pass the invocation deadline to
                                     // flush tasks here, so they can
                                     // retry if we have more time
+                                    //
+                                    // We always want to flush on a timeout
                                     if flush_control.should_flush_end() || status == Status::Timeout {
                                         tokio::join!(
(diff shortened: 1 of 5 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -400,7 +400,12 @@ async fn extension_loop_active(
                     match event {
                         Event::Metric(event) => {
                             debug!("Metric event: {:?}", event);
-                        }
+                        },
+                        Event::OutOfMemory => {
+                            let mut p = invocation_processor.lock().await;
+                            p.on_out_of_memory_error();
+                            drop(p);
```
alexgallotta, line 407: do we need this explicit drop?
They are all over the place and I think they are useless since it is done automatically at the end of the block

duncanista: Maybe not on the ones where we have it at the end, we can improve this later

alexgallotta (approved):

PR #470, opened 2024-11-22
fix(bottlecap): fallback when otel set

duncanista:
# What?

Fallsback to Go when any OTEL env var required for Serverless is set

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -30,6 +30,10 @@ pub struct FallbackConfig {
     serverless_appsec_enabled: bool,
     appsec_enabled: bool,
     profiling_enabled: bool,
+    // otel
+    trace_otel_enabled: bool,
+    otlp_config_receiver_protocols_http_endpoint: Option<String>,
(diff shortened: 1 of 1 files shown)


alexgallotta (approved):

bottlecap/src/config/mod.rs
```diff
@@ -255,6 +272,22 @@ pub mod tests {
         });
     }
 
+    #[test]
+    fn test_fallback_on_otel() {
+        figment::Jail::expect_with(|jail| {
+            jail.clear_env();
+            jail.set_env("DD_EXTENSION_VERSION", "next");
+            jail.set_env(
+                "DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_HTTP_ENDPOINT",
```
purple4reina, line 281: There's also a complementary `DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_GRPC_ENDPOINT` we should probably listen for as well.

duncanista: Its in the config file as fallback too

PR #478, opened 2024-11-25
priority of decrypt: kms > arn > use dd_api_key

alexgallotta:
Current logic was ignoring kms/secret arn when dd_api was set, but from the docs and goagent behavior it should be the other way around 
https://docs.datadoghq.com/serverless/libraries_integrations/cli/

> Datadog API Key encrypted using KMS. Sets the DD_KMS_API_KEY environment variable on your Lambda function configuration. Note: DD_API_KEY is ignored when DD_KMS_API_KEY is set.

> 	The ARN of the secret storing the Datadog API key in AWS Secrets Manager. Sets the DD_API_KEY_SECRET_ARN on your Lambda function configuration. Notes: DD_API_KEY_SECRET_ARN is ignored when DD_KMS_API_KEY is set. Add the secretsmanager:GetSecretValue permission to the Lambda execution role.

Also remove the 32 char hex validation

### DIFF
--- a/bottlecap/src/secrets/decrypt.rs
+++ b/bottlecap/src/secrets/decrypt.rs
@@ -13,35 +13,49 @@ use tracing::debug;
 use tracing::error;
 
 pub async fn resolve_secrets(config: Arc<Config>, aws_config: &AwsConfig) -> Option<String> {
-    if !config.api_key_secret_arn.is_empty() || !config.kms_api_key.is_empty() {
-        let before_decrypt = Instant::now();
+    let api_key_candidate =
+        if !config.api_key_secret_arn.is_empty() || !config.kms_api_key.is_empty() {
+            let before_decrypt = Instant::now();
 
-        let client = match Client::builder().use_rustls_tls().build() {
-            Ok(client) => client,
-            Err(err) => {
-                error!("Error creating reqwest client: {}", err);
-                return None;
+            let client = match Client::builder().use_rustls_tls().build() {
+                Ok(client) => client,
+                Err(err) => {
+                    error!("Error creating reqwest client: {}", err);
+                    return None;
+                }
+            };
+
+            let decrypted_key = if config.kms_api_key.is_empty() {
+                decrypt_aws_sm(&client, config.api_key_secret_arn.clone(), aws_config).await
+            } else {
+                decrypt_aws_kms(&client, config.kms_api_key.clone(), aws_config).await
+            };
+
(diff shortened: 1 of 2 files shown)


duncanista:
Great catch

bottlecap/src/secrets/decrypt.rs
```diff
@@ -27,10 +24,10 @@ pub async fn resolve_secrets(config: Arc<Config>, aws_config: &AwsConfig) -> Opt
             }
         };
 
-        let decrypted_key = if config.api_key_secret_arn.is_empty() {
-            decrypt_aws_kms(&client, config.kms_api_key.clone(), aws_config).await
+        let decrypted_key = if config.kms_api_key.is_empty() {
+            decrypt_aws_kms(&client, config.api_key_secret_arn.clone(), aws_config).await
```
duncanista, line 28: you're passing the secret arn to KMS

bottlecap/src/secrets/decrypt.rs
```diff
@@ -27,10 +24,10 @@ pub async fn resolve_secrets(config: Arc<Config>, aws_config: &AwsConfig) -> Opt
...
         } else {
-            decrypt_aws_sm(&client, config.api_key_secret_arn.clone(), aws_config).await
+            decrypt_aws_sm(&client, config.kms_api_key.clone(), aws_config).await
```
duncanista, line 30: you're passing KMS to secrets manager

bottlecap/src/secrets/decrypt.rs
```diff
@@ -43,7 +40,7 @@ pub async fn resolve_secrets(config: Arc<Config>, aws_config: &AwsConfig) -> Opt
             }
         }
     } else {
-        return None;
+        Some(config.api_key.clone())
```
duncanista, line 43: Will this resolve to None since the API key wouldn't be set either?

duncanista (commented):
One more thing 🍎

bottlecap/src/secrets/decrypt.rs
alexgallotta: Not sure what you mean.
Why shouldn't it be set? And how would it resolve to None?
I moved the cleanup code inside, so maybe it's less confusing

duncanista: If it's empty, we should send None

duncanista (approved):

PR #482, opened 2024-11-26
Feat: support adaptive flush strategy again

astuyve:
I removed this initially as I believed it either wasn't necessary or didn't help much (I think? I had just had a baby)
We've since gotten some feedback that, although users can set `DD_SERVERLESS_FLUSH_STRATEGY=periodically,10000` to flush every 10s regardless of the number of invocations, they prefer the automatic adaptive strategy we had earlier.

So I'm adding this back

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -11,7 +11,7 @@
 
 use bottlecap::{
     base_url,
-    config::{self, AwsConfig, Config},
+    config::{self, flush_strategy::FlushStrategy, AwsConfig, Config},
     event_bus::bus::EventBus,
     events::Event,
     lifecycle::{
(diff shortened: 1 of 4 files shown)


bottlecap/src/lifecycle/flush_control.rs
```diff
@@ -1,11 +1,18 @@
 use crate::config::flush_strategy::FlushStrategy;
+use std::time;
 use tokio::time::Interval;
+use tracing::debug;
+
+use super::invocation_times::InvocationTimes;
```
duncanista, line 6: use crate instead of super

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -504,6 +504,10 @@ async fn extension_loop_active(
                     if !flush_control.should_flush_end() {
                         break;
                     }
+
+                    if matches!(flush_control.flush_strategy, FlushStrategy::Periodically(_)) {
+                        break;
+                    }
```
duncanista, line 510: Not sure I understand this

duncanista (approved):

alexgallotta (approved):

bottlecap/src/bin/bottlecap/main.rs
astuyve: woops, needed to remove the above. We were repurposing `should_flush_end()` to effectively mean `is the flush strategy periodically), so I'm just hoisting that up here.

This code was effectively:
```
if !!matches!(flush_control.flush_strategy, Periodically)
```

Issue #493, opened 2024-12-09
DD v67 doesn't support DD_URL and DD_LOGS_CONFIG_LOGS_DD_URL


tommysitehost:
Hey ya, it doesn't seem like v67 respect `DD_LOGS_CONFIG_LOGS_DD_URL` and `DD_URL` env vars. Switching it to `DD_EXTENSION_VERSION=compatibility` fixes it. 

Will the next-generation lambda extension support  `DD_LOGS_CONFIG_LOGS_DD_URL` and `DD_URL` in the future?

Additionally, setting `DD_LOG_LEVEL=trace` doesn't generate a lot of trace logs on v67 compared to previous version.

duncanista:
Hey everyone,

I've added a PR to fallback directly in the meantime, so we don't send the data to the wrong intake URL.
In the meantime, I'm gonna merge the fix to prevent this, while we wait to add this feature!

tommysitehost:
@duncanista Hey ya, will we get another tag before xmas break for this merged PR?

duncanista:
Hey. @tommysitehost,

Originally, I was hoping to get a new release before holidays, but as the code freeze started at Datadog, I couldn't get a working build before that, sorry about it, but rest assured that our priority is releasing as soon as the holidays code freeze lifts. We have a running build and are hoping to release the new version very soon! Sorry about that!

natepage:
Hi @duncanista, not sure if you're the right person to ask, but do you know if this change would allow PCI compliance instrumentation?

We do have existing APM working with the PCI compliant endpoint, I have tried to instrument a Lambda function using the extension v66 and v67 but it doesn't seem to be working.
It could be related to not being able to have custom intake URLs, but would like to check if there could be something else preventing it from working.

Thank you 😄

duncanista:
Hey @natepage,

Yes, right now what this would do is fallback onto the Go agent until we get the PCI compliance instrumentation ready for the Next-Gen Lambda Extension.

We don't have this feature ready yet, but we will definitely keep this issue open so we can add it, rest assured that it is on our radar.

Will keep updated as soon as we get this going.

We expect to release tomorrow if everything goes well!

duncanista:
This should fallback now in the latest [v68](https://github.com/DataDog/datadog-lambda-extension/releases/tag/v68).

I'm gonna keep this issue, and rename it so we eventually add support for this!

tommysitehost:
[...]

PR #495, opened 2024-12-10
fix: fallback on intake urls

duncanista:
# What?

Prevents a bug where data is being set to the wrong intake, by fallbacking to the main agent while we support the feature in Rust.

# How?

Adding a fallback check when the configuration settings are present for `DD_URL`, APM, and Logs intake urls

# Motivation

Configured intake URLs are not supported yet for the Rust extension
#493

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -42,7 +42,7 @@ pub struct FallbackConfig {
     url: Option<String>,
     dd_url: Option<String>,
     logs_config_logs_dd_url: Option<String>,
-    // APM, as opposed to logs, do not use the `apm_config` prefix for env vars
+    // APM, as opposed to logs, does not use the `apm_config` prefix for env vars
     apm_dd_url: Option<String>,
 }
 
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -42,7 +42,8 @@ pub struct FallbackConfig {
     url: Option<String>,
     dd_url: Option<String>,
     logs_config_logs_dd_url: Option<String>,
-    apm_config_apm_dd_url: Option<String>,
+    // APM, as opposed to logs, do not use the `apm_config` prefix for env vars
+    apm_dd_url: Option<String>,
 }
 
 /// `FallbackYamlConfig` is a struct that represents fields in `datadog.yaml` not yet supported in the extension yet.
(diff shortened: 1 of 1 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -38,10 +38,23 @@ pub struct FallbackConfig {
     trace_otel_enabled: bool,
     otlp_config_receiver_protocols_http_endpoint: Option<String>,
     otlp_config_receiver_protocols_grpc_endpoint: Option<String>,
-    // YAML otel, we don't care about the content
-    otlp_config: Option<Value>,
+    // intake urls
+    url: Option<String>,
+    dd_url: Option<String>,
+    logs_config_logs_dd_url: Option<String>,
+    apm_config_apm_dd_url: Option<String>,
```
apiarian-datadog, line 45: does this also be cover the `DD_APM_DD_URL` environment variable spelling?

duncanista: That's not supported in the agent: https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml#L555-L582

So i'm not adding that one, as it could be directly used by the tracers, which we don't want to interfere with

apiarian-datadog: as seen here: https://datadoghq.atlassian.net/wiki/spaces/SLS/pages/2722136681/Serverless+Disaster+Recovery+Runbook#Extension

apiarian-datadog: the thing you linked is related to fips, what about this: https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml#L1096-L1101 ?

duncanista: That's under `apm_config`, as you can see here: https://github.com/DataDog/datadog-agent/blob/8a450cc476d33f6374164f1f00d92cdf65309558/pkg/config/config_template.yaml#L1047-L1103

apiarian-datadog: ah! does that mean that the tracer tells the agent where it should be sending the traces based on the `DD_APM_DD_URL` parameter?

duncanista: No, that's where the tracers will send the traces to

apiarian-datadog: got it, thanks! looks like i had a confused mental model of the architecture

duncanista: Actually, fixed it, env vars are indeed not following their own standard in the agent 😕

alexgallotta (approved):

PR #497, opened 2024-12-12
set dogstatsd timeout

alexgallotta:


bottlecap/Cargo.toml
duncanista, line 1: If [DataDog/libdatadog#800](https://github.com/DataDog/libdatadog/pull/800) gets merged, let's update this?

duncanista (approved):
LGTM – update fmt and license

PR #500, opened 2024-12-17
Decode error message and error type

shreyamalpani:
### What does this PR do?
Decodes error.msg and error.type headers as was previously done for the error.stack header, because of https://github.com/DataDog/dd-trace-dotnet/pull/6438 encoding error.msg and error.type.

### Motivation
[\[SLES-2001\] .NET Datadog Tracer throws exception from AWS Lambda Function](https://datadoghq.atlassian.net/browse/SLES-2001)
[\[SVLS-6041\] Having non-ascii characters in a stack trace causes an exception within the tracer.](https://datadoghq.atlassian.net/browse/SVLS-6041)

### Testing
Tested with a lambda that throws an exception with an error message containing a non-ASCII character.
image

[SLES-2001]: https://datadoghq.atlassian.net/browse/SLES-2001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SVLS-6041]: https://datadoghq.atlassian.net/browse/SVLS-6041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -596,8 +596,8 @@ impl Processor {
             }
 
             if let Some(s) = stack {
-                let decoded_stack = base64_to_string(s).unwrap_or_else(|_| {
-                    debug!("Error stack header may not be encoded, setting as is");
+                let decoded_stack = base64_to_string(s).unwrap_or_else(|e| {
+                    debug!("Failed to decode error stack: {e}");
                     s.to_string()
                 });
 
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -574,13 +574,10 @@ impl Processor {
             self.span.error = 1;
 
             if let Some(m) = message {
-                let decoded_message = match base64_to_string(m) {
-                    Ok(decoded) => decoded,
-                    Err(e) => {
-                        debug!("Failed to decode error message: {e}");
-                        m.to_string()
-                    }
-                };
+                let decoded_message = base64_to_string(m).unwrap_or_else(|_| {
+                    debug!("Error message header may not be encoded, setting as is");
+                    m.to_string()
+                });
 
                 self.span
                     .meta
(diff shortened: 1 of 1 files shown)


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -574,15 +574,31 @@ impl Processor {
             self.span.error = 1;
 
             if let Some(m) = message {
+                let decoded_message = match base64_to_string(m) {
+                    Ok(decoded) => decoded,
+                    Err(e) => {
+                        debug!("Failed to decode error message: {e}");
+                        m.to_string()
+                    }
+                };
```
duncanista, line 583: I'm pretty sure there's a cleaner way to do this

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -574,15 +574,31 @@ impl Processor {
             self.span.error = 1;
 
             if let Some(m) = message {
+                let decoded_message = match base64_to_string(m) {
+                    Ok(decoded) => decoded,
+                    Err(e) => {
+                        debug!("Failed to decode error message: {e}");
```
duncanista, line 580: Message is too ambiguous, since you're actually setting the error at the end, so it's kinda misleading and user might think that it failed, but the error message is actually there

bottlecap/src/lifecycle/invocation/processor.rs
duncanista, line 1: Needs unit testing

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -574,25 +574,32 @@ impl Processor {
             self.span.error = 1;
 
             if let Some(m) = message {
+                let decoded_message = base64_to_string(m).unwrap_or_else(|_| {
+                    debug!("Error message header may not be encoded, setting as is");
```
duncanista, line 578: Much more cleaner, great!

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -574,25 +574,32 @@ impl Processor {
...
             if let Some(s) = stack {
-                let decoded_stack = match base64_to_string(s) {
-                    Ok(decoded) => decoded,
-                    Err(e) => {
-                        debug!("Failed to decode error stack: {e}");
-                        s.to_string()
-                    }
-                };
+                let decoded_stack = base64_to_string(s).unwrap_or_else(|_| {
+                    debug!("Error stack header may not be encoded, setting as is");
```
duncanista, line 600: This always comes encoded, so the previous error message was OK

duncanista (approved):
Let's keep the previous message for the error stack, overall LGTM

PR #503, opened 2024-12-26
S3 Downstream Span Pointers

nhulston:
### What does this PR do?
_This is my first time writing Rust code, so focus on my syntax and other Rust concepts I may not understand yet._

Adds span pointers to spans for Lambdas triggered by S3 `putObject`, `copyObject`, and `completeMultipartUpload` events. This change will affect the downstream case for Universal Instrumentation Lambda runtimes (Java, .NET, Golang).

Span pointers are similar to [Span Links](https://docs.datadoghq.com/tracing/trace_collection/span_links/), but for cases when it is impossible to pass the Trace ID and Span ID between the spans that need to be linked.

When the calculated hashes for the upstream and downstream lambdas match, the Datadog backend will automatically link the two traces together.

Upstream case (requires tracer to be instrumented): Screenshot 2025-01-08 at 4 28 59 PM
Downstream case (this code):  
Screenshot 2025-01-08 at 4 29 51 PM

When clicking on the linked span, a new tab opens linking to the opposite Lambda function.

Mostly manual testing, but I also added unit tests. There are also functional tests in https://github.com/DataDog/serverless-self-monitoring/pull/347

To verify my changes, run [this Lambda function](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/nhulston-java-s3-span-pointers-dev-main?tab=testing) (only on the serverless AWS account) with the event payload
```json
{
  "shouldPutObject": false,
  "shouldCopyObject": false,
  "shouldMultipartUpload": false
}
```
and change one of the bools to true. Check Datadog to ensure that the spans are properly linked.

I also added unit tests:
```
cargo test test_generate_span_pointer_hash
cargo test test_get_span_pointers
cargo test test_attach_span_pointers_to_span
```

### Additional Notes
- This handles the 'downstream' case, for Lambdas that are triggered S3 requests. I will have another PR in each tracer to handle the 'upstream' case. [Java PR here](https://github.com/DataDog/dd-trace-java/pull/8075)
- S3 and Dynamo are supported in Python+Node. I will later open another PR in this repo to add support for Dynamo.


bottlecap/src/traces/trace_processor.rs
```diff
@@ -48,6 +51,33 @@ impl TraceChunkProcessor for ChunkProcessor {
...
+            if span.name == "aws.lambda" {
+                if let Some(span_pointers) = &self.span_pointers {
+                    if !span_pointers.is_empty() {
+                        let span_links: Vec<Value> = span_pointers
+                            .iter()
+                            .map(|sp| {
+                                json!({
+                                    "attributes": {
+                                        "link.kind": "span-pointer",
+                                        "ptr.dir": "u",
```
apiarian-datadog, line 63: they're always upstream?

bottlecap/src/span_pointers/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod helpers;
```
duncanista, line 1: I'd sugges to move `span_pointers` inside `traces`, so `traces/span_pointers/mod.rs` or `traces/span_pointers.rs`

you don't need the `helper.rs`, that's not how rust file structure is recommended.

bottlecap/src/span_pointers/helpers.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub kind: String,
+}
+
+#[must_use]
+pub fn generate_span_pointer_hash(components: &[&str]) -> String {
+    let data_to_hash = components.join("|");
+    let mut hasher = Sha256::new();
+    hasher.update(data_to_hash.as_bytes());
+    let result = hasher.finalize();
+    hex::encode(result)[..32].to_string()
```
duncanista, line 15: magic number?

bottlecap/src/span_pointers/helpers.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct SpanPointer {
+    pub hash: String,
+    pub kind: String,
+}
+
+#[must_use]
+pub fn generate_span_pointer_hash(components: &[&str]) -> String {
+    let data_to_hash = components.join("|");
+    let mut hasher = Sha256::new();
+    hasher.update(data_to_hash.as_bytes());
```
duncanista, line 13: Can't you just create the hash directly from the array join?

bottlecap/src/lifecycle/invocation/triggers/mod.rs
```diff
@@ -55,6 +56,11 @@ pub trait Trigger: ServiceNameResolver {
             .unwrap_or(&fallback.to_string())
             .to_string()
     }
+
+    /// Default implementation for adding span pointers
+    fn get_span_pointers(&self) -> Option<Vec<SpanPointer>> {
+        None
+    }
 }
```
duncanista, line 64: If Span Pointers are only available for S3 and DynamoDB, I'd suggest not creating a new method to the interface, as this is probably only gonna be used for those two.

Instead, create more methods directly in the implementation, an example can be seen in `step_functions_event.rs`

bottlecap/src/lifecycle/invocation/triggers/s3_event.rs
```diff
@@ -121,6 +122,25 @@ impl Trigger for S3Record {
     fn is_async(&self) -> bool {
         true
     }
+
+    fn get_span_pointers(&self) -> Option<Vec<SpanPointer>> {
```
duncanista, line 126: Why is the definition a vector?

bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -39,6 +40,8 @@ pub struct SpanInferrer {
     generated_span_context: Option<SpanContext>,
     // Tags generated from the trigger
     trigger_tags: Option<HashMap<String, String>>,
+    // Span pointers from S3 or DynamoDB streams
+    pub span_pointers: Vec<SpanPointer>,
```
duncanista, line 44: Make it an optional, you're already doing `Some()` wherever this is being used at the end, it doesn't make sense to have it here as a non-optional.

bottlecap/src/traces/trace_processor.rs
```diff
@@ -48,6 +51,33 @@ impl TraceChunkProcessor for ChunkProcessor {
                 }
             }
 
+            if span.name == "aws.lambda" {
+                if let Some(span_pointers) = &self.span_pointers {
+                    if !span_pointers.is_empty() {
+                        let span_links: Vec<Value> = span_pointers
+                            .iter()
```
duncanista, line 58: Too many nested ids, we need to find a better way for this.

bottlecap/src/traces/trace_processor.rs
```diff
@@ -48,6 +51,33 @@ impl TraceChunkProcessor for ChunkProcessor {
...
+                                })
+                            })
+                            .collect();
+
+                        if let Ok(span_links_json) = serde_json::to_string(&span_links) {
+                            span.meta
+                                .insert("_dd.span_links".to_string(), span_links_json);
+                        }
+                    }
+                }
```
duncanista, line 78: Can't this be a whole function?

bottlecap/src/traces/trace_processor.rs
```diff
@@ -48,6 +51,33 @@ impl TraceChunkProcessor for ChunkProcessor {
...
+                                    "attributes": {
+                                        "link.kind": "span-pointer",
+                                        "ptr.dir": "u",
+                                        "ptr.hash": sp.hash,
+                                        "ptr.kind": sp.kind,
+                                    },
+                                    "span_id": "0",
+                                    "trace_id": "0"
+                                })
+                            })
```
duncanista, line 70: Why this specific format? I'm not sure we want to have a just a json lying around here like this

duncanista (commented):
Left a couple comments

bottlecap/src/traces/trace_processor.rs
```diff
@@ -48,6 +51,33 @@ impl TraceChunkProcessor for ChunkProcessor {
...
+                            .collect();
+
+                        if let Ok(span_links_json) = serde_json::to_string(&span_links) {
+                            span.meta
+                                .insert("_dd.span_links".to_string(), span_links_json);
```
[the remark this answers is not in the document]
nhulston: I believe it just adds it to the array without overwriting it, but let me test and circle back!

bottlecap/src/traces/trace_processor.rs
```diff
@@ -48,6 +51,33 @@ impl TraceChunkProcessor for ChunkProcessor {
                 }
             }
 
+            if span.name == "aws.lambda" {
```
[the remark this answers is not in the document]
nhulston: If we add span pointers to every (or multiple) spans, the Datadog UI will display that number of span links rather than just 1.

We could add it to any span in the trace; it doesn't have to be the aws.lambda span. It could also be the S3 inferred span as well. But in Python/Node, we just add the span pointers the main aws.lambda span so I was just following that precedent.

It's definitely not optimal that we have to manually check the span name though. In Node we just get the lambda span from the `wrappedCurrentSpan` instance variable.
https://github.com/DataDog/datadog-lambda-js/blob/c726f30ba0d1dfa43b3b1a0fe025f4a80d14ad2a/src/trace/listener.ts#L204-L208

bottlecap/src/traces/trace_processor.rs
nhulston: Yes, this code only runs when the Lambda was triggered by an S3 stream.

All the cases where `ptr.dir` is down will be handled by the tracers

bottlecap/src/traces/trace_processor.rs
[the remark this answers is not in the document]
nhulston: Adding the span pointers to the S3 inferred span makes sense. But once we get to dynamo, there are different types of DynamoDB inferred spans. So I think it's better to just keep it simple and add the span pointers to the aws.lambda span since we know that it will always be there.

bottlecap/src/lifecycle/invocation/triggers/s3_event.rs
```diff
@@ -121,6 +122,25 @@ impl Trigger for S3Record {
...
+        let bucket_name = &self.s3.bucket.name;
+        let key = &self.s3.object.key;
+        let e_tag = self.s3.object.e_tag.trim_matches('"');
```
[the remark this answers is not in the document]
nhulston: When making a request (on the tracer side), the AWS SDK gives the raw eTag.
But on a S3 stream (on Bottlecap), the AWS SDK wraps the eTag in quotes. So we have to strip the quotes to make sure the hash calculation is identical on both sides

bottlecap/src/lifecycle/invocation/triggers/s3_event.rs
nhulston: Most of the time there will only be 1 span pointer. But the user can configure their stream settings so that the Lambda is triggered only after several S3 updates. In that case, there will be multiple S3 updates and therefore span pointers

bottlecap/src/lifecycle/invocation/triggers/s3_event.rs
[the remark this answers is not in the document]
duncanista: Let's document this in a comment or something

bottlecap/src/lifecycle/invocation/span_inferrer.rs
nhulston: Nice catch, thanks

bottlecap/src/traces/trace_processor.rs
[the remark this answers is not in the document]
apiarian-datadog: can we add it to the top level span, whatever that is, rather than the span named with a magical "aws.lambda" string?

bottlecap/src/lifecycle/invocation/triggers/mod.rs
nhulston: I agree, thanks for the suggestion

bottlecap/src/traces/trace_processor.rs
nhulston: Yes, I'll extract it to the spanpointers.rs file

bottlecap/src/traces/trace_processor.rs
nhulston: This is the format the Datadog frontend expects in order to be able to 'link' two traces together. I'm not sure if there's any other way to do it; do you have any suggestions?

bottlecap/src/span_pointers/helpers.rs
nhulston: For S3, yes we can; the as_bytes() here doesn't make a difference. But in Dynamo, the key names can be numbers, strings, or binary. I'll be reusing this function for the DynamoDB implementation

bottlecap/src/traces/spanpointers.rs
duncanista, line 1: `span_pointers.rs`

bottlecap/src/span_pointers/helpers.rs
duncanista: I mean, but you can still reduce this to a one liner, no? no matter if you keep using the `as_bytes()`

bottlecap/src/traces/trace_processor.rs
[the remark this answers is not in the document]
nhulston: The aws.lambda is the main span, and then Bottlecap creates the inferred S3 span

bottlecap/src/span_pointers/helpers.rs
nhulston: oh yes we can

bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -178,6 +182,9 @@ impl SpanInferrer {
         } else if S3Record::is_match(payload_value) {
             if let Some(t) = S3Record::new(payload_value.clone()) {
                 t.enrich_span(&mut inferred_span, &self.service_mapping);
+                if let Some(pointers) = t.get_span_pointers() {
+                    self.span_pointers = Some(pointers);
+                }
```
duncanista, line 187:
```suggestion
                self.span_pointers = t.get_span_pointers();
```

bottlecap/src/traces/trace_processor.rs
```diff
@@ -48,6 +50,12 @@ impl TraceChunkProcessor for ChunkProcessor {
...
+                if let Some(span_pointers) = &self.span_pointers {
+                    attach_span_pointers_to_meta(&mut span.meta, span_pointers);
+                }
```
duncanista, line 56: Can't you do the check inside of the method?

bottlecap/src/traces/span_pointers.rs
```diff
@@ -0,0 +1,179 @@
...
+        .map(|sp| {
+            json!({
+                "attributes": {
+                    "link.kind": "span-pointer",
+                    "ptr.dir": "u",
+                    "ptr.hash": sp.hash,
+                    "ptr.kind": sp.kind,
+                },
+                "span_id": "0",
+                "trace_id": "0"
```
duncanista, line 46: Maybe add a comment about why we are setting this to 0?

duncanista, line 46: Make sure to use the `trace-protobuf` implementation of `libdatadog` for the `SpanLink` structure instead of the JSON you have here

bottlecap/src/traces/trace_processor.rs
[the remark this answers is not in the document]
nhulston: Okay, I've changed it so we get the span directly from the index; no need to loop through and check the span names

bottlecap/src/traces/span_pointers.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    let span_links: Vec<SpanLink> = span_pointers
+        .iter()
+        .map(|sp| {
+            let mut attributes = HashMap::new();
+            attributes.insert("link.kind".to_string(), "span-pointer".to_string());
+            attributes.insert("ptr.dir".to_string(), "u".to_string());
+            attributes.insert("ptr.hash".to_string(), sp.hash.clone());
+            attributes.insert("ptr.kind".to_string(), sp.kind.clone());
```
duncanista, line 46: HashMap::from([...])

duncanista: You can add it directly in the `SpanLink` with this constructor

bottlecap/src/traces/span_pointers.rs
```diff
@@ -0,0 +1,198 @@
...
+
+            SpanLink {
+                // We set trace_id and span_id to 0 since they're unknown; the frontend
+                // uses `ptr.hash` instead to find the opposite link if it exists.
+                trace_id: 0,
+                span_id: 0,
+                attributes,
```
duncanista, line 53: Move the attributes to the top/bottom, and then explain that not only trace_id or span_id, but all trace propagation fields

bottlecap/src/traces/trace_processor.rs
```diff
@@ -57,6 +59,10 @@ impl TraceChunkProcessor for ChunkProcessor {
                 .insert("_dd.origin".to_string(), "lambda".to_string());
             obfuscate_span(span, &self.obfuscation_config);
         }
+
+        if let Some(span) = chunk.spans.get_mut(root_span_index) {
+            attach_span_pointers_to_meta(&mut span.meta, &self.span_pointers);
```
duncanista, line 64: Nice

bottlecap/src/traces/span_pointers.rs
```diff
@@ -0,0 +1,198 @@
...
+pub fn generate_span_pointer_hash(components: &[&str]) -> String {
+    let mut hasher = Sha256::new();
+    hasher.update(components.join("|").as_bytes());
+    let result = hasher.finalize();
+    hex::encode(result)[..SPAN_POINTER_HASH_LENGTH].to_string()
+}
+
+pub fn attach_span_pointers_to_meta<S: ::std::hash::BuildHasher>(
+    meta: &mut HashMap<String, String, S>,
+    span_pointers_option: &Option<Vec<SpanPointer>>,
```
duncanista, line 29: don't call it `option`, the typing are there for a reason

bottlecap/src/traces/span_pointers.rs
```diff
@@ -0,0 +1,198 @@
...
+) {
+    let Some(span_pointers) = span_pointers_option else {
+        return;
+    };
+
+    if span_pointers.is_empty() {
+        return;
+    }
```
duncanista, line 37:
```suggestion
    let Some(span_pointers) = span_pointers_option.as_ref().filter(|sp| !sp.is_empty()) else {
            return;
    };
```

bottlecap/src/traces/span_pointers.rs
```diff
@@ -0,0 +1,193 @@
...
+                    ("ptr.hash".to_string(), sp.hash.clone()),
+                    ("ptr.kind".to_string(), sp.kind.clone()),
+                ]),
+            }
+        })
+        .collect();
+
+    if let Ok(span_links_json) = serde_json::to_string(&span_links) {
+        meta.insert("_dd.span_links".to_string(), span_links_json);
+    }
```
duncanista, line 58:
```suggestion
    serde_json::to_string(&span_links)
            .map(|json| meta.insert("_dd.span_links".to_string(), json));
```

duncanista (approved):

bottlecap/src/traces/trace_processor.rs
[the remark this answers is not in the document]
nhulston: I stand corrected; this replaced existing span links. Thanks for catching this! I just added logic in the latest 2 commits to handle&test this case

bottlecap/src/traces/span_pointers.rs
```diff
@@ -0,0 +1,257 @@
...
+                    ("ptr.hash".to_string(), sp.hash.clone()),
+                    ("ptr.kind".to_string(), sp.kind.clone()),
+                ]),
+            }
+        })
+        .collect();
+
+    let mut all_span_links = meta
+        .get("_dd.span_links")
+        .and_then(|existing| serde_json::from_str::<Vec<SpanLink>>(existing).ok())
```
duncanista, line 58: why casting them and not just appending to them? What happens if you cannot cast?

wouldn't it be easier to just append to whatever is inside `_dd.span_links`? can't you just instead of replacing whatever is inside the json just add the new ones? not sure why we are casting the existing data here

nhulston: AFAIK we can't append without casting because `meta.get("_dd.span_links")` is a string.

One option could be to parse the string manually -- remove the last `]`, insert a comma if required, insert the new span link string, and readd the `]`

wdyt?

duncanista: What if you parsed an array of strings and then appended your already serialized string?

bottlecap/src/traces/span_pointers.rs
```diff
@@ -0,0 +1,257 @@
...
+    span_pointers: &Option<Vec<SpanPointer>>,
+) {
+    let Some(span_pointers) = span_pointers.as_ref().filter(|sp| !sp.is_empty()) else {
+        return;
+    };
+
+    let new_span_links: Vec<SpanLink> = span_pointers
+        .iter()
+        .map(|sp| {
+            SpanLink {
```
apiarian-datadog, line 38: do these `SpanLink` structs have an implied or expected constraint that the trace and span ids are not zero? the python version did, so we had to actually create a subclass of the span link class to correctly store span pointers. also, just to check, does the serialization code do any deduplication on the trace and span id, or does it just ship all the links that it finds?

bottlecap/LICENSE-3rdparty.yml
```diff
@@ -40,7 +40,19 @@ third_party_libraries:
...
+      Permission to use, copy, modify, and/or distribute this software for
+      any purpose with or without fee is hereby granted.
+
+      THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+      WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+      MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+      ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+      WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+      AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT
+      OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
```
purple4reina, line 55: This change seems important enough to me to warrant its own PR.

bottlecap/src/traces/span_pointers.rs
nhulston: okay Jordan and I discussed this and this seems like the best way to do it. It's definitely not optimal, but parsing into the `SpanLink` struct rather than a `Value` does make the serde operation faster

bottlecap/LICENSE-3rdparty.yml
nhulston: cc @duncanista 
He said he knows the issue and it's fine to keep it in this PR

bottlecap/src/traces/span_pointers.rs
nhulston: Good idea, thanks for bringing this up. It does work in both cases, when trace and span ids are and are not 0. I'll update the unit test to have 0 for trace+span id

PR #505, opened 2025-01-07
Enable tmp, fd, and threads enhanced metrics

shreyamalpani:
Based on self-monitoring tests, this should fix the issue of the process-based metrics adding a lot of post-runtime duration overhead.

[Full investigation doc](https://docs.google.com/document/d/1nn-dcJwxtvH6Fs_Dnj_QNm9c_KrSvWCv4s0TyJ7E4k8/edit?usp=sharing)

### DIFF
--- a/bottlecap/src/metrics/enhanced/lambda.rs
+++ b/bottlecap/src/metrics/enhanced/lambda.rs
@@ -450,12 +450,7 @@ impl Lambda {
         }
     }
 
-    #[allow(unreachable_code)]
-    #[allow(unused_variables)]
-    #[allow(unused_mut)]
     pub fn set_tmp_enhanced_metrics(&self, mut send_metrics: Receiver<()>) {
-        // Temporarily disabled
-        return;
         if !self.config.enhanced_metrics {
             return;
         }
@@ -502,9 +497,11 @@ impl Lambda {
         });
     }
 
-    pub fn generate_fd_enhanced_metrics(
+    pub fn generate_process_enhanced_metrics(
         fd_max: f64,
         fd_use: f64,
+        threads_max: f64,
+        threads_use: f64,
         aggr: &mut std::sync::MutexGuard<Aggregator>,
         tags: Option<SortedTags>,
     ) {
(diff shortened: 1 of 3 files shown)


bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -502,9 +497,11 @@ impl Lambda {
         });
     }
 
-    pub fn generate_fd_enhanced_metrics(
+    pub fn generate_process_enhanced_metrics(
```
duncanista, line 500:
```suggestion
    pub fn generate_process_metrics(
```

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -573,15 +558,25 @@ impl Lambda {
...
+            let mut pids = proc::get_pid_list();
 
-            // Set fd_max and initial value for fd_use to -1
+            // Set fd_max and initial value for fd_use
             let fd_max = proc::get_fd_max_data(&pids);
-            let mut fd_use = -1_f64;
+            let mut fd_use = if let Ok(fd_use) = proc::get_fd_use_data(&pids) {
+                fd_use
+            } else {
+                debug!("Could not get file descriptor use.");
```
duncanista, line 568: More descriptive? Which purpose?
```suggestion
                debug!("Could not get file descriptor use.");
```

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -573,15 +558,25 @@ impl Lambda {
...
+            };
 
-            // Set threads_max and initial value for threads_use to -1
+            // Set threads_max and initial value for threads_use
             let threads_max = proc::get_threads_max_data(&pids);
-            let mut threads_use = -1_f64;
+            let mut threads_use = if let Ok(threads_use) = proc::get_threads_use_data(&pids) {
+                threads_use
+            } else {
+                debug!("Could not get threads use.");
```
duncanista, line 577: Same as above
```suggestion
            let mut threads_use = if let Ok(threads_use) = proc::get_threads_use_data(&pids) {
                threads_use
            } else {
                debug!("Could not get threads use.");
```

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -573,15 +558,25 @@ impl Lambda {
...
+                -1_f64
+            };
```
duncanista, line 580: Could this be done as an `ok_or_else(|...|)` or something similar?

I also don't know if the debug log can be added there in the else part, lmk

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -573,15 +558,25 @@ impl Lambda {
...
 
             let mut interval = interval(Duration::from_millis(1));
```
duncanista, line 581:
```suggestion
            let mut interval = interval(Duration::from_millis(1));
```

Turn the number into a constant, and let's see if increasing it benefits us? So we don't fire up every ms

duncanista (approved):

PR #512, opened 2025-01-13
fix: update placeholder value for datadog-agent in gitlab pipeline

rogeriocamorim:
Upgrade the lambda extension upgrade lot of security issue.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,7 +10,7 @@ variables:
     value: main
   AGENT_VERSION:
     description: "Latest release version of the datadog-agent to tag the build with."
-    value: "7.55.1"
+    value: "7.61.0"
   LAYER_SUFFIX:
     description: "Suffix to be appended to the layer name (default empty)."
     value: ""


.gitlab-ci.yml
duncanista, line 1: Hey, the current Agent is set as `7.61.0`, this is just a placeholder value for our pipeline release, but it's not being actually used, as we update this on a constant basis on every release. You can check the release [here](https://github.com/DataDog/datadog-lambda-extension/releases/tag/v68) – but rest assured that the latest version was built with `7.60.1`, we'll soon release a new version with `7.61.0` when the time comes! 

Thanks for the change tho!

duncanista (approved):

.gitlab-ci.yml
rogeriocamorim: Gotcha is more because we use that and the security vulnerability fixed here (https://github.com/DataDog/datadog-agent/pull/32380) is something that we need to adjust Asap on our system.

duncanista: Sounds good, thanks for explaining that!

Will make sure we release a new layer with more bug fixes and this soon!

PR #517, opened 2025-01-16
feat: [SVLS-5992] support DD_URL and DD_DD_URL for metrics intake

apiarian-datadog:


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -284,13 +285,27 @@ async fn extension_loop_active(
...
         .expect("failed to create aggregator"),
     ));
-    let mut metrics_flusher = MetricsFlusher::new(
-        resolved_api_key.clone(),
-        Arc::clone(&metrics_aggr),
-        build_fqdn_metrics(config.site.clone()),
-        config.https_proxy.clone(),
-        Duration::from_secs(config.flush_timeout),
-    );
+    let metrics_intake_url_prefix = MetricsIntakeUrlPrefix::new(
```
apiarian-datadog, line 288: this feels like it should really be part of the `config` but that doesn't play well with our config parsing process. still, maybe we should be computing this stuff and handling sites and overrides and such in the calling function?

bottlecap/src/config/mod.rs
```diff
@@ -39,8 +39,6 @@ pub struct FallbackConfig {
     otlp_config_receiver_protocols_http_endpoint: Option<String>,
     otlp_config_receiver_protocols_grpc_endpoint: Option<String>,
     // intake urls
-    url: Option<String>,
-    dd_url: Option<String>,
```
apiarian-datadog, line 43: is this the right way to move configs from fallback into a fully supported state?

bottlecap/src/bin/bottlecap/main.rs
apiarian-datadog: is this an opportunity to split the `config_for_parsing` from the `config_for_using` structs? c.c. @duncanista @alexgallotta

bottlecap/src/config/mod.rs
```diff
@@ -107,6 +105,11 @@ pub struct Config {
     pub trace_propagation_style_extract: Vec<TracePropagationStyle>,
     pub trace_propagation_extract_first: bool,
     pub trace_propagation_http_baggage_enabled: bool,
+    // Intake URLs
+    // These two are for metrics intake. The official config is `DD_DD_URL` but `DD_URL` can also
+    // be used for backwards compatibility. `DD_DD_URL` takes precedence.
+    pub url: Option<String>,
```
astuyve, line 111: The next-gen extension is already a breaking change, can we just drop this and add it to a breaking change note in the readme?

duncanista: @astuyve what about customers using old configuration? we just fallback? what's the politics behind this on how we manage a different configuration?

bottlecap/src/bin/bottlecap/main.rs
alexgallotta: yep, let's keep it all close in the config/mod.rs

alexgallotta: At this stage, the code should have a validated and working immutable config struct, ready to be used around/
The only exception would be for config that changes dynamically, but that must be address with a prod/consumer or other async stuff

bottlecap/src/config/mod.rs
alexgallotta: If it's somehow ok to remove that deprecated config, I would say to go for it.
Worst case, the few who will complain will have a quick fix

bottlecap/src/bin/bottlecap/main.rs
apiarian-datadog: well, i've pushed it up into `main()`. will followup on a thread there.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -163,6 +164,41 @@ fn build_function_arn(account_id: &str, region: &str, function_name: &str) -> St
 async fn main() -> Result<()> {
     let (aws_config, config) = load_configs();
 
+    let site = Site::new(config.site.clone()).map_err(|e| {
```
apiarian-datadog, line 167: @duncanista, @alexgallotta : how would y'all want me to handle this config parsing block? should i push it into `load_configs()` and change that to return `aws_config, config, and metrics_intake_url_prefix`? or something else?

astuyve:
Closing in favor of #585

PR #520, opened 2025-01-22
fix: Honor noproxy and skip proxying if ddsite is in the noproxy list

### LINKED ISSUE #514: Lambda Extension not properly honoring NO_PROXY env var
We are trying to configure a Lambda in a VPC with no egress which has all the Datadog VPC Endpoints set up with private dns. For all non-Datadog traffic, we must use an outbound proxy with a strict allowlist. 

Our Lambda is set up with the following env vars:

```
DD_API_KEY_SECRET_ARN="arn:to:api:key"
DD_LOG_LEVEL="debug"
DD_PROXY_NO_PROXY_NONEXACT_MATCH="true"


HTTP_PROXY="http://internal-proxy.example.internal:10000"
HTTPS_PROXY="http://internal-proxy.example.internal:10000"
NO_PROXY="127.0.0.1,localhost,172.16.0.0/12,us-east-1.amazonaws.com,datadoghq.com"
```

When the lambda starts it succesfully fetches it's API key through the VPC Endpoint (not the proxy). However, when the lambda tries to send traces and metrics, we see the following error logs:

```
DD_EXTENSION | ERROR | Failed to send logs to datadog: error sending request for url (https://http-intake.logs.datadoghq.com/api/v2/logs)
DD_EXTENSION | DEBUG | 500: Failed to push to API: reqwest::Error { kind: Request, url: "https://api.datadoghq.com/api/beta/sketches", source: hyper_util::client::legacy::Error(Connect, "unsuccessful tunnel") }
```

We confirm on our HTTP proxy that it received CONNECT requests for various datadohq.com endpoints and rejected them all. 

The `reqwest` and `hyper_util` in the DEBUG log make me think that the Rust client in the extension is not honoring or parsing the `NO_PROXY` env var correctly? But the fact that it got its API key correctly means the AWS SDK Client is honoring that env var?

This is on Extension v67

Any ideas?

astuyve:
Fixes #514 by checking if NO_PROXY contains DD_SITE domain, and the the user is likely relying on privatelink endpoints + private DNS isntead of an HTTP proxy for telemetry data.

This is kind of a hack in 2 ways:
1. `reqwest` has some support for NO_PROXY but `hyper-proxy` doesn't. I could use the built in version, but either way I'd need to build it myself
2. We simply pass a proxy string to libdatadog for APM + statsd, we don't actually pass config anywhere. So we'd have to add that.

This solution works for the customer reporting it, as their NO_PROXY string includes `datadoghq.com`.

But I wonder if there's a use case for a customer who would use privatelink for, say metrics only - but then use an HTTP proxy for logs (perhaps to inspect logs for PII or some other form of compliance).

In that case our entire shared http client architecture (complete with shared threadpools) would need to change. I think we should leave that for the future at this time, and help our friends in #514 move on.

Screenshots of my config:
image

VPC endpoints as per our [docs](https://docs.datadoghq.com/agent/guide/private-link/?tab=connectfromsameregion):
image

Logs/metrics/traces:
image
image
image

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -529,6 +529,33 @@ pub mod tests {
     }
 
     #[test]
+    fn test_proxy_config() {
+        figment::Jail::expect_with(|jail| {
+            jail.clear_env();
+            jail.set_env("DD_PROXY_HTTPS", "my-proxy:3128");
+            let config = get_config(Path::new("")).expect("should parse config");
+            assert_eq!(config.https_proxy, Some("my-proxy:3128".to_string()));
+            Ok(())
+        });
+    }
+
+    #[test]
+    fn test_noproxy_config() {
+        figment::Jail::expect_with(|jail| {
+            jail.clear_env();
+            jail.set_env("DD_SITE", "datadoghq.eu");
+            jail.set_env("DD_PROXY_HTTPS", "my-proxy:3128");
+            jail.set_env(
+                "NO_PROXY",
+                "127.0.0.1,localhost,172.16.0.0/12,us-east-1.amazonaws.com,datadoghq.eu",
+            );
+            let config = get_config(Path::new("")).expect("should parse noproxy");
+            assert_eq!(config.https_proxy, None);
+            Ok(())
+        });
+    }
+
+    #[test]
     fn test_parse_flush_strategy_end() {
         figment::Jail::expect_with(|jail| {
             jail.clear_env();
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -261,11 +261,12 @@ pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
     if config.site.is_empty() {
         config.site = "datadoghq.com".to_string();
     }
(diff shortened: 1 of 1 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -260,7 +261,12 @@ pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
     if config.site.is_empty() {
         config.site = "datadoghq.com".to_string();
     }
-
+    if let Ok(no_proxy_string) = std::env::var("NO_PROXY") {
```
duncanista, line 264:
```suggestion
    if std::env::var("NO_PROXY").map_or(false, |no_proxy| no_proxy.contains(&config.site)) {
        debug!("NO_PROXY contains DD_SITE, disabling proxy");
        config.https_proxy = None;
    }
```

duncanista, line 264: Should we add a `todo` comment?

duncanista (approved):

bottlecap/src/config/mod.rs
astuyve: For potentially parsing the entire string?

astuyve: yeah could oneline it

PR #528, opened 2025-01-27
fix: licenses

astuyve:
Updates the 3rdparty license files

### DIFF
--- a/bottlecap/LICENSE-3rdparty.yml
+++ b/bottlecap/LICENSE-3rdparty.yml
@@ -21481,7 +21481,46 @@ third_party_libraries:
 
       END OF TERMS AND CONDITIONS
   - license: Unicode-3.0
-    text: NOT FOUND
+    text: |
+      UNICODE LICENSE V3
+
+      COPYRIGHT AND PERMISSION NOTICE
+
+      Copyright © 1991-2023 Unicode, Inc.
+
+      NOTICE TO USER: Carefully read the following legal agreement. BY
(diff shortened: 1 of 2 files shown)


bottlecap/LICENSE-3rdparty.yml
```diff
@@ -21481,46 +21481,7 @@ third_party_libraries:
...
-      OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS,
-      WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
-      ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THE DATA
-      FILES OR SOFTWARE.
-
-      Except as contained in this notice, the name of a copyright holder shall
-      not be used in advertising or otherwise to promote the sale, use or other
-      dealings in these Data Files or Software without prior written
-      authorization of the copyright holder.
+    text: NOT FOUND
```
duncanista, line 21484: Removing a license 🤔 

This bug has been happening for a while, sometimes it gets fixed randomly

duncanista (approved):

bottlecap/LICENSE-3rdparty.yml
astuyve: The bug is that our CI is installing cargo-bundle-licenses via an unqualified installation command, so it was pulling the latest version, in this case v3.

My local machine had v2.x, so there were some conflicts. Our cache key is perhaps a big ambiguous, but in this case it's passing and the fix was just updating `cargo-bundle-licenses` on my machine.

PR #529, opened 2025-01-28
feat: use new dsd

astuyve:
Bumps our dep on libdatadog, new metrics intake/site things are a breaking change but since we don't rely on a version, we gotta impl it.


duncanista (approved):

duncanista:
Issues with testing – but approving overall

alexgallotta (approved):

PR #532, opened 2025-01-29
feat: Support snapstart's vended credentials

astuyve:
Lambda SnapStart doesn't provide AWS access credentials to prevent them from being cached during the snapshot creation process.
We have two options, we can delay resolving credentials into the invoke phase (probably don't wanna do that for performance reasons where users rely on the additional vCPU power during init), or we can use the credential API to grab creds during init and use them.

Since we never re-use credentials, this works fine for us. It should only incur overhead during snapshot creation, which is a control plane operation anyway.

Screenshot of the decrypted secrets w/ snapshot (long since expired):
image
(can't wait for a bot to try these old expired creds)

### DIFF
--- a/bottlecap/src/secrets/decrypt.rs
+++ b/bottlecap/src/secrets/decrypt.rs
@@ -53,9 +53,9 @@ pub async fn resolve_secrets(config: Arc<Config>, aws_config: &mut AwsConfig) ->
             }
 
             let decrypted_key = if config.kms_api_key.is_empty() {
-                decrypt_aws_sm(&client, config.api_key_secret_arn.clone(), &aws_config).await
+                decrypt_aws_sm(&client, config.api_key_secret_arn.clone(), aws_config).await
             } else {
-                decrypt_aws_kms(&client, config.kms_api_key.clone(), &aws_config).await
+                decrypt_aws_kms(&client, config.kms_api_key.clone(), aws_config).await
             };
 
             debug!("Decrypt took {}ms", before_decrypt.elapsed().as_millis());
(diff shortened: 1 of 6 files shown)


bottlecap/src/secrets/decrypt.rs
```diff
@@ -24,11 +24,39 @@ pub async fn resolve_secrets(config: Arc<Config>, aws_config: &AwsConfig) -> Opt
                     return None;
                 }
             };
+            let mut cloned_config = aws_config.clone();
```
duncanista, line 27: `cloned_config` is too confusing, this method has a `config` and `aws_config`, yet this `cloned_config` is a clone of `aws_config`

Do we even need the `cloned_config`?

Can't we then just pass a clone below to `decrypt_aws_*` instead of cloning and then using it in this if?

astuyve: I debated making the aws_config mutable, maybe that's a better option instead of cloning it. I believe we fully consume the aws_config anyway in this case, it's not used by other functions.

bottlecap/src/telemetry/events.rs
```diff
@@ -125,6 +125,30 @@ pub enum TelemetryRecord {
...
+
+    #[serde(rename = "platform.restoreReport", rename_all = "camelCase")]
+    PlatformRestoreReport {
+        /// Status of the invocation
+        status: Status,
+        /// When unsuccessful, the `error_type` describes what kind of error occurred
+        error_type: Option<String>,
+    },
+    #[serde(rename = "platform.restoreRuntimeDone", rename_all = "camelCase")]
+    PlatformRestoreRuntimeDone {
```
duncanista, line 146: Maybe we could provide some snapstart visibility eventually? seems easy enough

duncanista (approved):

duncanista:
Looks good! Fix clippy and :shipit:

bottlecap/src/telemetry/events.rs
astuyve: I think we could report the metric, yeah

duncanista: Even create a span

astuyve: Why? It happens up to 2 weeks before an invocation occurs

PR #534, opened 2025-01-29
feat: add an Aggregator for Traces and Stats

duncanista:
# What?

Adds aggregators for Traces and Stats.

# Motivation

Large trace payloads are being dropped, because they're rejected by the backend.
Also [SVLS-6250](https://datadoghq.atlassian.net/browse/SVLS-6250)

# Test

- Added unit tests
- Tested in an AWS Lambda with a heavy amount of incoming traces

[SVLS-6250]: https://datadoghq.atlassian.net/browse/SVLS-6250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/traces/stats_aggregator.rs
+++ b/bottlecap/src/traces/stats_aggregator.rs
@@ -1,3 +1,4 @@
+use datadog_trace_protobuf::pb::ClientStatsPayload;
 use prost::Message;
 use std::collections::VecDeque;
 
@@ -17,19 +18,13 @@ use std::collections::VecDeque;
 const MAX_CONTENT_SIZE_BYTES: usize = 3 * 1024 * 1024; // ~3MB
 
 #[allow(clippy::module_name_repetitions)]
-pub struct StatsAggregator<T>
-where
-    T: Message,
-{
-    queue: VecDeque<T>,
+pub struct StatsAggregator {
+    queue: VecDeque<ClientStatsPayload>,
     max_content_size_bytes: usize,
-    buffer: Vec<T>,
+    buffer: Vec<ClientStatsPayload>,
 }
 
-impl<T> Default for StatsAggregator<T>
-where
-    T: Message,
-{
+impl Default for StatsAggregator {
     fn default() -> Self {
         StatsAggregator {
             queue: VecDeque::new(),
@@ -39,10 +34,7 @@ where
     }
 }
 
-impl<T> StatsAggregator<T>
-where
-    T: Message,
-{
+impl StatsAggregator {
     #[allow(dead_code)]
     #[allow(clippy::must_use_candidate)]
     pub fn new(max_content_size_bytes: usize) -> Self {
@@ -53,11 +45,11 @@ where
         }
     }
 
-    pub fn add(&mut self, message: T) {
+    pub fn add(&mut self, message: ClientStatsPayload) {
         self.queue.push_back(message);
     }
 
-    pub fn get_batch(&mut self) -> Vec<T> {
+    pub fn get_batch(&mut self) -> Vec<ClientStatsPayload> {
         let mut batch_size = 0;
 
         // Fill the batch
--- a/bottlecap/src/traces/stats_aggregator.rs
+++ b/bottlecap/src/traces/stats_aggregator.rs
@@ -45,8 +45,8 @@ impl StatsAggregator {
         }
     }
 
-    pub fn add(&mut self, message: ClientStatsPayload) {
-        self.queue.push_back(message);
+    pub fn add(&mut self, payload: ClientStatsPayload) {
+        self.queue.push_back(payload);
     }
 
     pub fn get_batch(&mut self) -> Vec<ClientStatsPayload> {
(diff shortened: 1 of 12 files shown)


bottlecap/src/traces/stats_aggregator.rs
```diff
@@ -0,0 +1,82 @@
...
+/// <https://github.com/DataDog/datadog-agent/blob/996dd54337908a6511948fabd2a41420ba919a8b/pkg/trace/writer/stats.go#L33-L35>
+// const MAX_ENTRY_SIZE_BYTES: usize = 375;
+
+/// Maximum content size per payload in compressed bytes,
+///
+/// <https://github.com/DataDog/datadog-agent/blob/996dd54337908a6511948fabd2a41420ba919a8b/pkg/trace/writer/stats.go#L35-L41>
+const MAX_CONTENT_SIZE_BYTES: usize = 3 * 1024 * 1024; // ~3MB
+
+#[allow(clippy::module_name_repetitions)]
+pub struct StatsAggregator<T>
```
duncanista, line 20: Had to make it a <T> because using the `ClientStatsPayload` type directly wouldn't allow me to use the `encoded_len` method from `pb::Message`

duncanista: Seems commit `588585e` fixes this

bottlecap/src/traces/stats_flusher.rs
```diff
@@ -4,55 +4,73 @@
...
-        tokio::spawn(async move {
-            while let Some(stats_payload) = rx.recv().await {
-                let mut buffer = buffer_producer.lock().await;
-                buffer.push(stats_payload);
-            }
-        });
+        let endpoint = Endpoint {
+            url: hyper::Uri::from_str(&stats_url).expect("can't make URI from stats url, exiting"),
+            api_key: Some(api_key.clone().into()),
+            timeout_ms: Endpoint::DEFAULT_TIMEOUT,
```
alexgallotta, line 52: Can you use the config timeout here? So we are consistent with the timeouts

bottlecap/src/traces/trace_agent.rs
```diff
@@ -47,31 +47,36 @@ pub enum ApiVersion {
...
 impl TraceAgent {
     #[must_use]
-    pub async fn new(
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
         config: Arc<config::Config>,
+        trace_aggregator: Arc<Mutex<trace_aggregator::TraceAggregator>>,
         trace_processor: Arc<dyn trace_processor::TraceProcessor + Send + Sync>,
-        trace_flusher: Arc<dyn trace_flusher::TraceFlusher + Send + Sync>,
+        stats_aggregator: Arc<Mutex<stats_aggregator::StatsAggregator>>,
```
alexgallotta, line 55: these types are becoming a pain to handle. Can we start using type aliases?

bottlecap/src/traces/trace_aggregator.rs
```diff
@@ -0,0 +1,169 @@
...
+pub const MAX_CONTENT_SIZE_BYTES: usize = (32 * 1_024 * 1_024) / 10;
+
+#[allow(clippy::module_name_repetitions)]
+pub struct TraceAggregator {
+    queue: VecDeque<SendData>,
+    max_content_size_bytes: usize,
+    buffer: Vec<SendData>,
+}
+
+impl Default for TraceAggregator {
```
alexgallotta, line 17: all this file is duplicate of trace stats aggregator, maybe dedup?

duncanista: Not really, this one has a different type, and they have different ways of handling how the data is treated.

Trace payload the length in the struct, whereas Stats Payload is a proto where we use method from the trait Message to get the length of the payload.

bottlecap/src/traces/trace_agent.rs
duncanista: Maybe? We'd have to define them at the top for all of them, maybe not in this PR.

bottlecap/src/traces/stats_flusher.rs
duncanista: I'm just moving what was there before, maybe I could update it in another PR too to not add unrelated changes to aggregation!

alexgallotta: oh I see, ok I am changing all of them in my pr, don't worry about that

bottlecap/src/traces/stats_aggregator.rs
```diff
@@ -0,0 +1,167 @@
...
+            max_content_size_bytes: MAX_CONTENT_SIZE_BYTES,
+            buffer: Vec::with_capacity(MAX_CONTENT_SIZE_BYTES),
+        }
+    }
+}
+
+impl StatsAggregator {
+    #[allow(dead_code)]
+    #[allow(clippy::must_use_candidate)]
+    pub fn new(max_content_size_bytes: usize) -> Self {
```
astuyve, line 40: Could just re-use the constant here right? Or are you simply parameterizing it for the test?

bottlecap/src/traces/stats_aggregator.rs
```diff
@@ -0,0 +1,167 @@
+use datadog_trace_protobuf::pb::ClientStatsPayload;
+use prost::Message;
```
astuyve, line 2: We're using prost just to measure the size? Is there a way we can do that with whatever protobuf lib we're already using?

astuyve: ah sick, awesome ty

bottlecap/src/traces/stats_aggregator.rs
astuyve: param is good then

PR #535, opened 2025-01-30
aj/fix fd max metric

astuyve:
the early return here while mapping through the PID list means if a PID dies, we immediately stop reporting metrics from the rest.

This change just trace-level logs that and moves on.

In the future, we should rescan the pid list occasionally

before:
image

after:
image


bottlecap/src/tags/lambda/tags.rs
```diff
@@ -49,7 +49,7 @@ const FUNCTION_TAGS_KEY: &str = "_dd.tags.function";
 // TODO(astuyve) decide what to do with the version
 const EXTENSION_VERSION_KEY: &str = "dd_extension_version";
 // TODO(duncanista) figure out a better way to not hardcode this
-const EXTENSION_VERSION: &str = "68-next";
+const EXTENSION_VERSION: &str = "69-next";
```
alexgallotta, line 52: is it a good idea to change the version in this PR?

alexgallotta (approved):

bottlecap/src/tags/lambda/tags.rs
astuyve: whoops, removing

PR #539, opened 2025-02-04
feat: Extract sampling priority from tracer and apply to new lambda span

nhulston:
### What does this PR do?

**How sampling works normally (outside of Lambda):** 
The tracer decides whether or not to sample a trace based on the `DD_TRACE_SAMPLING_RULES` (new) or `DD_TRACE_SAMPLING_RATE` (deprecated) env vars. Then, the tracer sets the `_sampling_priority_v1` metric to the sampling priority on some or all spans. If the priority is <= 0, it should be dropped; otherwise it should be sampled.

Then the agent gets the sampling priority from this metric and decides whether to drop or sample the trace based on this sampling priority and other factors, e.g. if an error occurred or other special rules.

**How sampling works in Lambda**
Since the serverless agent drops the lambda span received from the tracer and creates a new lambda span, we instead get the sampling priority from the request headers of the `/lambda/end-invocation` request. 
- Currently the Go serverless agent sets the chunk priority directly
- This PR sets the sampling priority on the newly created aws.lambda span, and then lets `libdatadog` set the chunk priority in `normalize`
https://github.com/DataDog/libdatadog/blob/c6ad4ff82852ed6d39fd618a602020b6ba623ed6/trace-normalization/src/normalizer.rs#L87-L99 

The complex sampling logic exists in the main `datadog-agent` and in the backend, but not in the serverless Go agent. Therefore, in Lambda (Python, Node, and serverless Go agent), we historically just send all traces to the backend, and let it decide whether to sample/drop traces.

This allows env variables like `DD_TRACE_SAMPLING_RULES` and `DD_TRACE_SAMPLING_RATE` to work with Bottlecap for universally instrumented runtimes.

Manual testing. Prior to this PR, sampling rules worked in Node+Python, but not in Java/.NET/Golang with Bottlecap.

After these changes, sampling rules work in: Java+Golang now, and Node+Python still work as expected. As for .NET, this is one fix, but it turns out the .NET tracer does not send the correct sampling priority header on the `/lambda/end-invocation` request. That needs to be fixed.

I also added unit tests:
```shell
cargo test test_update_span_context_with_sampling_priority
cargo test test_update_span_context_with_invalid_priority
cargo test test_update_span_context_no_sampling_priority
```

### Additional Notes
In the future, we could implement the complex sampling logic that exists in the main agent to take some load off the backend (this would be done in libdatadog). This consists of logic like drop traces with negative priority but don't drop traces with errors, etc. This logic would also have to be implemented in Node+Python too.

There are many types of sampling rules, here's the list in order by priority. This PR covers the first cases 1-4 and 6 but not 5:
1. remote sampling rules
2. local sampling rules
3. remote global sampling rate
4. local global sampling rate
5. sampling rates from the agent (max traces per second)
6. if nothing else, rate defaults to 100% (keep all traces)


bottlecap/src/traces/trace_processor.rs
```diff
@@ -35,10 +35,19 @@ struct ChunkProcessor {
...
     fn process(&mut self, chunk: &mut pb::TraceChunk, root_span_index: usize) {
+        if let Some(priority_i8) = self.sampling_priority {
+            chunk.priority = priority_i8.into();
+            if priority_i8 <= 0 {
+                chunk.dropped_trace = true;
+                return;
+            }
+        }
+
         chunk
```
duncanista, line 51: This feels like something very generic which we should not do at this level, but instead on the tracing libraries of libdatadog when you're collecting a chunk

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -537,7 +541,12 @@ impl Processor {
                 parent_id = header.parse::<u64>().unwrap_or(0);
             }
 
-            // TODO: sampling priority extraction
+            if let Some(priority_str) = headers.get(DATADOG_SAMPLING_PRIORITY_KEY) {
```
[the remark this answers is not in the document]
nhulston: We discussed this but just posting here for other reviewers:

I'm copying logic from the Go agent. On end invocation, the Go agent similarly calls InjectContext (equivalent to `update_span_context_from_headers` in Rust):
https://github.com/DataDog/datadog-agent/blob/23a03806e9998192fbe5fcd1d38e193acf8d7713/pkg/serverless/daemon/routes.go#L139-L144

which extracts the sampling priority from the headers of the `/lambda/end-invocation` request from the tracer:
https://github.com/DataDog/datadog-agent/blob/5cfbd2648275aa960ab91796a48aa21ca81dec52/pkg/serverless/invocationlifecycle/trace.go#L239-L242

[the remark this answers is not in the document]
nhulston: As for the `on_invocation_start`, this is implemented in: https://github.com/DataDog/datadog-lambda-extension/blob/40b0e064f2371eb867165205a5c713ca9390fb12/bottlecap/src/lifecycle/listener.rs#L109-L112
This handles the case where sampling rules are defined as traces/second on the agent level, and the agent communicates sampling priority back to the tracer.
This likely doesn't work in Bottlecap yet (TODO test), and could be implemented in a future PR

bottlecap/src/traces/trace_processor.rs
nhulston: So we should actually just be setting the priority here, but not marking the chunk as dropped. That's because if the tracer gives a negative priority but an error occurs during execution, the trace should still get sampled.
Simplifying to:
```
        if let Some(priority_i8) = self.sampling_priority {
            chunk.priority = priority_i8.into();
        }
```

This still works as expected; the backend now decides whether to sample or drop the trace; all we do is set the priority on the chunk

bottlecap/src/lifecycle/invocation/processor.rs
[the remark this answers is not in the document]
nhulston: Okay as discussed in person/Slack:

we just need to set the sampling priority under the `_sampling_priority_v1` metric on any span in the trace (we're doing it on the newly created lambda span since we know that's always there)

then libdatadog will set the chunk priority: https://github.com/DataDog/libdatadog/blob/c6ad4ff82852ed6d39fd618a602020b6ba623ed6/trace-normalization/src/normalizer.rs#L87-L99

then the backend will the chunk priority to sample or drop the trace

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -537,7 +538,14 @@ impl Processor {
...
+                if let Ok(priority) = priority_str.parse::<f64>() {
+                    self.span
+                        .metrics
+                        .insert(TAG_SAMPLING_PRIORITY.to_string(), priority);
+                    debug!("Extracted sampling priority from tracer: {priority}");
```
duncanista, line 546: Do we really need the debug log? We expect to always get this sampling priority, right?

nhulston: yeah probably not

duncanista (approved):
LGTM – make sure to manually test in a Lambda with an inferred span, double hop inferred span, and cold start. Check in APM that those traces also get dropped.

nhulston:
Tested and works with:
- Cold start
- Child span
- Inferred span with no trace context

Sampling priority is not propagated through single/double hop trace propagation cases; we'd need to implement that in a separate PR. The Go agent logic that handles that is in this file: https://github.com/DataDog/datadog-agent/blob/4b5c8b9270fe4626702db6d66298a060176251d0/pkg/serverless/trace/propagation/extractor.go

They're also dropped in APM

PR #542, opened 2025-02-05
fix: do not allocate memory for buffer

duncanista:
# What?

Removes allocating memory for the buffer batch

# Why?

Since the buffer is set as:
- `SendData` for traces
- `ClientStatsPayload` for stats

I'm allocating the buffer with the max content size in bytes, so it's around 3 million allocations of the object, instead of actual bytes.

### DIFF
--- a/bottlecap/src/traces/stats_aggregator.rs
+++ b/bottlecap/src/traces/stats_aggregator.rs
@@ -29,7 +29,7 @@ impl Default for StatsAggregator {
         StatsAggregator {
             queue: VecDeque::new(),
             max_content_size_bytes: MAX_CONTENT_SIZE_BYTES,
-            buffer: Vec::with_capacity(MAX_CONTENT_SIZE_BYTES),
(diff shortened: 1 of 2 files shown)


duncanista:
https://github.com/DataDog/datadog-lambda-extension/blob/1fd17871a39f44aac9dbfd6928eff23584b6370e/bottlecap/src/logs/aggregator.rs#L10-L22

This one is OK because I'm allocating actual bytes

alexgallotta (approved):

alexgallotta:
Looks good!
Btw the fact that there are 3 places doing the same exact logic, with 2 of them suffering of this bug and the other doesn't is quite bad. It makes everything way more tricky than it should, someone not familiar with the code could have fixed only one of the buffers, let it run on 512mb without seeing the issue

That is the main reason for https://github.com/DataDog/datadog-lambda-extension/compare/main...alexgallotta/refactor

PR #546, opened 2025-02-06
feat: get tags from top level span from tracer

duncanista:
# What?

Provides a proposal to get the tags from the incoming top-level span and attach them to the `aws.lambda` span generated in this project.

# Motivation

[SVLS-4645](https://datadoghq.atlassian.net/browse/SVLS-4645)

- Top level span gets custom tags directly set on the span
- Missing metrics are set on top level span

[SVLS-4645]: https://datadoghq.atlassian.net/browse/SVLS-4645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bottlecap/src/traces/trace_processor.rs
```diff
@@ -101,6 +106,10 @@ fn filter_span_from_lambda_library_or_runtime(span: &Span) -> bool {
         }
     }
     if span.resource == INVOCATION_SPAN_RESOURCE {
+        let mut guard = context_buffer.lock().expect("lock poisoned");
+        if let Some(request_id) = span.meta.get("request_id") {
```
duncanista, line 110: We assume the top level span is including the `request_id`, allowing us to match the data in the context buffer.

This is done in Ruby and Go.

.NET would have to send the top-level span first, as that one gets ditched and never arrives, should be a quick fix, but might need to get access to context so we can tag the current span with the `request_id`.

Not sure how Java handles this.

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -20,13 +22,33 @@ impl Context {
...
+        }
+    }
+
+    #[must_use]
+    pub fn from_request_id(request_id: &str) -> Self {
+        Context::new(request_id.to_owned(), 0f64, 0f64, 0, None, None)
+    }
+}
+
+impl Default for Context {
```
astuyve, line 43: do we only need this for tests? if so can we wrap it in a test macro? Defaults can cause subtle bugs

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -319,6 +329,7 @@ impl Processor {
                 _ = offsets.process_chan_tx.send(());
             }
         }
+        drop(context_buffer);
```
astuyve, line 332: are we sure we need the manual drop here? Doesn't the closure on 307 help us by 581? Is the drop necessary? If we add new code after, we have to re-lock right?

bottlecap/src/lifecycle/invocation/context.rs
duncanista: We do use it for logs processor, where we have default empty at the beginning all the time

astuyve (approved):

bottlecap/src/lifecycle/invocation/processor.rs
duncanista: Yeah, so since we only await at the very end, I decided to lock in the last place we modify the `context_buffer`, in reality, I'm dropping to be 100% sure that nothing strange happens. I'll add another comment to note that we drop because we await later on 332

alexgallotta (commented):
Instead of passing around the context buffer under a lock all over the place,
if only the request_id is needed, can't it be just cloned as string and passed?

If anything else is needed also, the trace_agent already has the `ServerlessTraceProcessor` which has the context buffer too.
It seems a bit convoluted to pass it around everywhere like that

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -20,13 +22,33 @@ impl Context {
...
+    fn default() -> Self {
+        Context {
+            request_id: String::new(),
```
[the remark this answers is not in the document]
duncanista: Yup, updated

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -51,7 +51,7 @@ const TAG_SAMPLING_PRIORITY: &str = "_sampling_priority_v1";
 
 pub struct Processor {
     // Buffer containing context of the previous 5 invocations
-    pub context_buffer: ContextBuffer,
+    pub context_buffer: Arc<Mutex<ContextBuffer>>,
```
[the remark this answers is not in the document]
duncanista: I can remove the `pub`, and I'd like to remove the lock altogether, yet I cannot find a way to get the desired outcome without the lock.

The lifecycle processor is just another component which has access to the context buffer, which the trace processor should have access too. They both need access because if the tracer sends by any chance a top level span, we need to make sure to add it to the current context of the given request id, that way whenever there's a platform runtime done, we can attach the tracer top level span information onto the one we create. WDYT?

bottlecap/src/lifecycle/invocation/processor.rs
duncanista: Instead, since I don't need to assign it to anything, I think I can refactor this by moving the first method to the bottom close to where we actually use the `get(...)` method and then use a scope to directly drop the lock 🤔

PR #550, opened 2025-02-10
Fallback on gov regions

nhulston:
For FIPs compliance, we're falling back to the Go agent when the Lambda is in a govcloud region.

Tested manually, we do fall back as expected on gov regions:
```
{
  "DD_EXTENSION_FALLBACK_REASON": "gov_region"
}
```

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -7,7 +7,7 @@ pub mod trace_propagation_style;
 use std::collections::HashMap;
 use std::path::Path;
 use std::time::Instant;
-use std::vec;
+use std::{env, vec};
 
 use figment::providers::{Format, Yaml};
 use figment::{providers::Env, Figment};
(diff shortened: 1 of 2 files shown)


duncanista (changes requested):
Pass `aws_config` existing instance.

nhulston:
> Pass `aws_config` existing instance.

We have to make a copy of whatever we pass; since we're only using region, I'm just passing the region only to reduce overhead

bottlecap/src/config/mod.rs
```diff
@@ -177,17 +177,26 @@ fn fallback(figment: &Figment, yaml_figment: &Figment) -> Result<(), ConfigError
...
     // Customer explicitly opted out of the Next Gen extension
-    let opted_out = match config.extension_version.as_deref() {
+    let is_compatibility = match config.extension_version.as_deref() {
         Some("compatibility") => true,
         // We want customers using the `next` to not be affected
         _ => false,
     };
 
+    // Check for us-gov region
+    let opted_out = is_compatibility || region.starts_with("us-gov-");
```
duncanista, line 187: You don't need a full clone just to do a `starts_with`, a reference is enough

nhulston: If I don't, clippy will fail

```
error[E0382]: use of partially moved value: `aws_config`
   --> src/bin/bottlecap/main.rs:222:6
    |
214 |     let config = match config::get_config(Path::new(&lambda_directory), aws_config.region) {
    |                                                                         ----------------- value partially moved here
...
222 |     (aws_config, config)
    |      ^^^^^^^^^^ value used here after partial move
    |
    = note: partial move occurs because `aws_config.region` has type `std::string::String`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0382`.
error: could not compile `bottlecap` (bin "bottlecap") due to 1 previous error

```

bottlecap/src/config/mod.rs
```diff
@@ -177,17 +177,26 @@ fn fallback(figment: &Figment, yaml_figment: &Figment) -> Result<(), ConfigError
...
+        log_fallback_reason(if is_compatibility {
+            "extension_version"
+        } else {
+            "gov_region"
+        });
+        return Err(ConfigError::UnsupportedField(if is_compatibility {
+            "extension_version".to_string()
+        } else {
+            "gov_region".to_string()
+        }));
```
duncanista, line 199: govcloud should be its own section, just like we do below for appsec or otel, update this code to reflect that

duncanista (approved):

PR #553, opened 2025-02-14
[dotnet] disable some features by default

lucaspimentel:
~Stacked on #623. Merged that PR first.~ ✅

By default, disable a few unused features in `dd-trace-dotnet` to improve startup times and total execution times:

env var name | Description
-------------|------------
`DD_TRACE_STARTUP_LOGS=0` | skip generating the startup log entry (which writes to the file system)
`DD_CIVISIBILITY_ENABLED=0` | explicitly disable CI Visibility to skip it's auto-detection logic


hghotra (approved):
✅

PR #558, opened 2025-02-18
feat: add zstd and compress

astuyve:
Adds zstd and compresses log payloads. Improves perf for people sending large/many logs:
image

I did not test a representative corpus of data so my intention is to test it in self monitoring, but I assume most logs are duplicates and will compress nicely. Went from 44mb on the network to 14mb, and dropped post runtime duration substantially, so that seems worthwhile.


bottlecap/src/logs/flusher.rs
```diff
@@ -59,14 +63,18 @@ impl Flusher {
 
     async fn send(client: reqwest::Client, api_key: String, fqdn: String, data: Vec<u8>) {
         let url = format!("{fqdn}/api/v2/logs");
+        let mut encoder = Encoder::new(Vec::new(), 0).unwrap();
```
duncanista, line 66: This would create an encoder every time we are sending data, is it possible to reuse it somehow?

bottlecap/src/logs/flusher.rs
```diff
@@ -59,14 +63,18 @@ impl Flusher {
...
+        let body = encoder.finish().unwrap();
 
         if !data.is_empty() {
             let resp: Result<reqwest::Response, reqwest::Error> = client
                 .post(&url)
                 .header("DD-API-KEY", api_key)
                 .header("DD-PROTOCOL", "agent-json")
                 .header("Content-Type", "application/json")
-                .body(data)
+                .header("Content-Encoding", "zstd")
```
alexgallotta, line 76: should we add an option to enable/disable decompression?
It might be needed for audit or other user side application

bottlecap/src/logs/flusher.rs
astuyve: I believe the encoder is finalized to get the compressed bytes and can't be reused

duncanista (approved):

PR #563, opened 2025-02-21
aj/new loop

astuyve:
This removes about ~30ms of latency for customers with very busy functions by no longer blocking on the `platform.runtimeDone` event, which is provided to us by Lambda via the Telemetry API. The call has a minimum 25ms buffer time, which is why the delay is present.
image

- Splits the loop into 3 distinct branches:
    - Flushing at the end of a invocation, used for first few invocations
    - Flushing at the beginning of an invocation, used for flush cycle for busy periodic functions
    - Skip flushing, used for non-flush cycle
- we still flush with the race timeout in order to handle long running functions where customers want telemetry data while the function is running. This fixes an issue where we would break until the end for long running functions after flushing once.
- includes a fix to read all of the logs flush response


bottlecap/src/logs/flusher.rs
```diff
@@ -101,18 +102,21 @@ impl Flusher {
...
         let time = Instant::now();
         let resp = req.send().await;
         let elapsed = time.elapsed();
+        debug!("Sent logs to datadog in {}ms", elapsed.as_millis());
 
         match resp {
             Ok(resp) => {
-                if resp.status() != 202 {
+                let status = resp.status();
+                _ = resp.text().await;
```
astuyve, line 114: Fixes an issue where keepalive can't be used if we don't fully read/exhaust the response buffer

bottlecap/src/logs/flusher.rs
```diff
@@ -101,18 +102,21 @@ impl Flusher {
             .body(body)
     }
 
-    async fn send(req: reqwest::RequestBuilder) -> Result<(), Box<dyn Error>> {
+    async fn send(req: reqwest::RequestBuilder) -> Result<(), Box<dyn Error + Send>> {
         let time = Instant::now();
         let resp = req.send().await;
         let elapsed = time.elapsed();
+        debug!("Sent logs to datadog in {}ms", elapsed.as_millis());
```
duncanista, line 109: Do we want to have this on every request for debug?

bottlecap/src/logger.rs
```diff
@@ -22,7 +22,12 @@ where
     ) -> fmt::Result {
         // Format values from the event's's metadata:
         let metadata = event.metadata();
-        write!(&mut writer, "DD_EXTENSION | {} | ", metadata.level())?;
+        write!(
+            &mut writer,
+            "DD_EXTENSION | {} | {} | ",
+            metadata.name(),
```
duncanista, line 28: This would add the file on every debug log, do we want this?

bottlecap/src/lifecycle/flush_control.rs
```diff
@@ -35,25 +35,17 @@ impl FlushControl {
...
             FlushStrategy::Periodically(_) => false,
             FlushStrategy::Default => {
                 if self.invocation_times.should_adapt_to_periodic(now) {
-                    let should_periodic_flush = self.should_periodic_flush(now, TWENTY_SECONDS);
-                    debug!(
-                        "Adapting over to periodic flush strategy. should_periodic_flush: {}",
-                        should_periodic_flush
-                    );
-                    return should_periodic_flush;
+                    return false;
```
duncanista, line 48: Not sure I understand this

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -486,17 +523,163 @@ async fn extension_loop_active(
...
+            } => {
+                debug!(
+                    "Runtime done for request_id: {:?} with status: {:?} and error: {:?}",
+                    request_id,
+                    status,
+                    error_type.unwrap_or("None".to_string())
+                );
+
+                if let Some(metrics) = metrics {
+                    return Some(RuntimeDoneMeta {
```
duncanista, line 615: Can't we just sent the event back?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -222,11 +241,9 @@ fn load_configs() -> (AwsConfig, Arc<Config>) {
...
 }
 
-fn enable_logging_subsystem(config: &Arc<Config>) {
-    let env_filter = format!(
-        "h2=off,hyper=off,rustls=off,datadog-trace-mini-agent=off,{:?}",
-        config.log_level
-    );
+fn enable_logging_subsystem() {
+    let env_filter =
+        "h2=off,hyper=off,reqwest=off,rustls=off,datadog-trace-mini-agent=off".to_string();
```
duncanista, line 246: might be worth to add `h1` too

bottlecap/src/logs/flusher.rs
astuyve: Nope, accident

bottlecap/src/lifecycle/flush_control.rs
astuyve: nope, accident

bottlecap/src/bin/bottlecap/main.rs
astuyve: yeah that's probably better

bottlecap/src/bin/bottlecap/main.rs
astuyve: h1 is not a library

bottlecap/src/logger.rs
astuyve: nope

bottlecap/src/lifecycle/flush_control.rs
astuyve: previously this method would control whether or not to switch over to flushing periodically _as well as_ determining if we've switched to periodic and it's time to flush. That's partially why the main loop was so confusing.

Instead, we've lifted that logic to the main loop, so this method can now be simplified. It only determines if the user specified they want to flush at end, or if we're on the default strategy – if we've seen enough invocations to flip over to periodic flushing. If the answer is yes, then we should not flush at end.

bottlecap/src/bin/bottlecap/main.rs
duncanista: Got carried over by thinking h2 = http2, h1 = http1

You're right

bottlecap/src/bin/bottlecap/main.rs
duncanista: I don't like it a 100% but we can change it in the future. My idea was still sending the clone of the event as an optional, that way we ensure we don't have to push the conditional back to the loop.

It just feels unnecessary when an object already exist for this which has the same information.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -106,15 +105,35 @@ async fn next_event(client: &reqwest::Client, ext_id: &str) -> Result<NextEventR
...
+        })?;
+
+    let status = response.status();
+    let text = response.text().await.map_err(|e| {
+        debug!("Failed to read response body: {}", e);
+        Error::new(std::io::ErrorKind::InvalidData, e.to_string())
+    })?;
+
+    if !status.is_success() {
+        debug!("HTTP Error {} - Response: {}", status, text);
```
duncanista, line 126: Should we be more explicit with this errors?

To mention something about the Next Event?

astuyve: yeah and they probably should be errors as we can't do anything if we get a failure here

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -330,154 +347,174 @@ async fn extension_loop_active(
...
-            Err(err) => {
-                eprintln!("Error: {err:?}");
-                println!("Exiting");
-                return Err(err);
+        let shutdown;
+        // Don't await! Pin it and check in the tokio loop
+        // Tokio select! drops whatever task does not complete first
+        // but the state machine API in Lambda will give us an invalid error
+        // if we call /next twice instead of waiting
+        //let mut next_response_received = false;
```
duncanista, line 364:
```suggestion
```

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -330,154 +347,174 @@ async fn extension_loop_active(
...
+            // flush
+            flush_all(
+                &logs_flusher,
+                &mut metrics_flusher,
+                &*trace_flusher,
+                &*stats_flusher,
+                &mut flush_interval,
+            )
+            .await;
+            flush_interval.reset();
```
duncanista, line 422: Wonder if its possible to run the flush all and the next_lambda_response in joinset? Just an idea

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -330,154 +347,174 @@ async fn extension_loop_active(
...
-            }
-            Err(err) => {
-                eprintln!("Error: {err:?}");
-                println!("Exiting");
-                return Err(err);
+        let shutdown;
+        // Don't await! Pin it and check in the tokio loop
+        // Tokio select! drops whatever task does not complete first
+        // but the state machine API in Lambda will give us an invalid error
+        // if we call /next twice instead of waiting
```
duncanista, line 363: Maybe this comment goes for the pin below?

bottlecap/src/bin/bottlecap/main.rs
astuyve: No because when `/next` is called the program halts until we receive the next invocation. That can lead to missing/dropped telemetry data.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -330,154 +347,174 @@ async fn extension_loop_active(
...
+            'inner: loop {
+                tokio::select! {
+                biased;
+                    next_response = &mut next_lambda_response => {
+                        // Dear reader this is important, you may be tempted to remove this
+                        // after all, why reset the flush interval if we're not flushing?
+                        // It's because the flush_interval is only for the RACE FLUSH
+                        // For long-running txns. The call to `flush_control.should_flush_end()`
+                        // has its own interval which is not reset here.
+                        flush_interval.reset();
```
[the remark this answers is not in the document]
astuyve: Yes we should

bottlecap/src/bin/bottlecap/main.rs
astuyve: it did

bottlecap/src/bin/bottlecap/main.rs
astuyve: yeah it's a little weirder with the enum because we strip a bunch of the event info away and just return what we need. It's another alloc but it should be stack alloc'd and fast anyway. We can change it later

duncanista (approved):

PR #582, opened 2025-02-28
aj/timestamp metrics

astuyve:
1. Adds aggregated 10second buckets for all metrics.
2. Fixes an issue where flush interval wasn't always respected.
3. adds compressed distribution pbuff support, which is pretty decent as tags compress nicely


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -531,6 +536,10 @@ async fn extension_loop_active(
                 &mut race_flush_interval,
             )
             .await;
+            println!(
```
astuyve, line 539: todo: remove

bottlecap/src/events/mod.rs
```diff
@@ -15,9 +15,21 @@ impl MetricEvent {
...
+    pub timestamp: i64,
+}
+
+impl OomEvent {
+    #[must_use]
+    pub fn new(timestamp: i64) -> OomEvent {
+        OomEvent { timestamp }
+    }
+}
+
```
duncanista, line 29: Why not just make the OutOfMemory event be a `OutOfMemory(i64)`?

astuyve: I naively assumed we'd hang other metadata off of it, but we can make it an enum instead of a struct.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -653,8 +659,14 @@ async fn handle_next_invocation(
...
+            let now = std::time::UNIX_EPOCH
+                .elapsed()
+                .expect("can't poll clock, unrecoverable")
+                .as_secs()
+                .try_into()
+                .unwrap_or_default();
             let mut p = invocation_processor.lock().await;
-            p.on_invoke_event(request_id);
+            p.on_invoke_event(request_id, now);
             drop(p);
```
duncanista, line 670: Can we move this into the function?

duncanista (commented):
Looks pretty good!

Goated PR, there's a bunch of changes, let's bake this in self monitoring for a while once it's merged

bottlecap/src/events/mod.rs
astuyve: Done

bottlecap/src/bin/bottlecap/main.rs
astuyve: Done

duncanista (approved):

PR #585, opened 2025-03-04
Aj/support pci and custom endpoints

### LINKED ISSUE #580: PCI Intake support
Hi there,

We tried to use this next-gen extension to instrument our AWS Lambda applications, setup went well, telemetry is flowing through,
However we are facing an issue when it comes to logs as our Datadog organisation is "PCI compliance enabled".

By the default this extension will attempt to send the logs to `https://http-intake.logs.datadoghq.com/api/v2/logs` which results in 403 in our case with the following response:

```json
{
    "errors": [
        {
            "status": "403",
            "title": "Forbidden",
            "detail": "Api key is either forbidden or blacklisted"
        }
    ]
}
```

After trying to understand what we were doing wrong, read docs, source code, etc. We finally tried to request Datadog API ourselves to:

- first validate our API key, and it's valid
- then send logs to Datadog API ourselves

These experiments allowed us to realised that sending logs to the PCI intake endpoints succeeds with 202 Accepted! Yay! 😄 

```http
POST /api/v2/logs HTTP/1.1
DD-API-KEY: <OUR_DD_API_KEY>
Content-Type: application/json
User-Agent: PostmanRuntime/7.43.0
Accept: */*
Cache-Control: no-cache
Postman-Token: 136eb612-0dcd-4dbf-bceb-0d770a39931b
Host: http-intake-pci.logs.datadoghq.com
Accept-Encoding: gzip, deflate, br
Connection: keep-alive
Content-Length: 8
 
{
}
 
HTTP/1.1 202 Accepted
Date: Fri, 28 Feb 2025 01:02:58 GMT
Content-Type: application/json
Content-Length: 2
Connection: keep-alive
cross-origin-resource-policy: cross-origin
accept-encoding: identity,gzip,x-gzip,deflate,x-deflate,zstd
x-content-type-options: nosniff
strict-transport-security: max-age=31536000; includeSubDomains; preload
 
{}
```

I know that this was kinda discussed already as part of #493 , but this issue was closed with a potential solution using DD_PROXY_HTTPS.
We've tried a few things using the proxy configurations, but nothing ended up being successful.
So maybe we are missing something here and a bit of guidance would allow us to get to a working solution.

Otherwise would it be possible to introduce a `usePciIntake|DD_USE_PCI_INTAKE` configuration in a similar fashion as the Datadog Browser RUM SDK that would tell the logs flusher to use the `http-intake-pci.logs...` URL?

This would be amazing and allow us to instrumente our AWS Lambda applications! 🎉 

Thank you for reading this 😄

astuyve:
Closes #580 

Also adds support for TCP proxies (they use the same config so I extended it for metrics).

Verified this manually (although it's covered by automated tests) because you need a special Datadog Org (and thus API key) to use PCI endpoints. Will share org ID in slack.
Logs:
image
Traces:
image


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -315,12 +314,26 @@ async fn extension_loop_active(
...
+
     let flusher_config = MetricsFlusherConfig {
         api_key: resolved_api_key.clone(),
         aggregator: Arc::clone(&metrics_aggr),
-        metrics_intake_url_prefix: MetricsIntakeUrlPrefix::new(Some(metrics_site), None)
-            .expect("can't parse metrics intake URL from site"),
+        metrics_intake_url_prefix: metrics_intake_url.expect("can't parse site or override"),
         https_proxy: config.https_proxy.clone(),
         timeout: Duration::from_secs(config.flush_timeout),
     };
```
duncanista, line 339: Could you create a `start_metrics_agent` function that covers all of this? just to simplify the `extension_loop`

astuyve: probably

bottlecap/src/config/mod.rs
```diff
@@ -141,6 +139,7 @@ impl Default for Config {
             logs_config_processing_rules: None,
             logs_config_use_compression: true,
             logs_config_compression_level: 6,
+            logs_config_logs_dd_url: String::default(),
```
duncanista, line 142: Why making them empty instead of `Option<String>`?

bottlecap/src/config/mod.rs
```diff
@@ -314,10 +300,27 @@ pub fn get_config(config_directory: &Path, region: &str) -> Result<Config, Confi
             .trace_propagation_style_extract
             .clone_from(&config.trace_propagation_style);
     }
+    if config.logs_config_logs_dd_url.is_empty() {
+        config.logs_config_logs_dd_url = build_fqdn_logs(config.site.clone());
```
duncanista, line 304: Does the Go Agent validate that the URL is correct? As in, validate that its indeed a URL

bottlecap/src/config/mod.rs
duncanista: I see, so you're re-using it at the end because technically, it will always exist, but we want to ensure that we can set a different one if it exist.

astuyve: Exactly.

bottlecap/src/config/mod.rs
apiarian-datadog: back when i looked at it, i don't think it does, at least not for metrics. presumably if the resulting url-shaped-string does not work as a url, there would be a runtime crash.

duncanista: Yup, that's what I'm afraid of

bottlecap/src/bin/bottlecap/main.rs
astuyve: We have to – config handles merging env vars as well as the `datadog.yaml` config file along with their precedence, so removing them from the config means they won't be read at all.

bottlecap/src/config/mod.rs
astuyve: Url-shaped strings can crash for reasons other than formatting errors (incorrectly configured proxies being the most common one), so I'm not sure what more we can do here

PR #587, opened 2025-03-05
Produced bytes enhanced metric

shreyamalpani:
Adds the produced bytes enhanced metric received on PlatformRuntimeDone

### DIFF
--- a/bottlecap/src/metrics/enhanced/lambda.rs
+++ b/bottlecap/src/metrics/enhanced/lambda.rs
@@ -866,7 +866,7 @@ mod tests {
         lambda.set_runtime_done_metrics(
             &RuntimeDoneMetrics {
                 duration_ms: 100.0,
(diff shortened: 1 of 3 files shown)


bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -1,9 +1,12 @@
-use crate::metrics::enhanced::{
-    constants::{self, BASE_LAMBDA_INVOCATION_PRICE},
-    statfs::statfs_info,
-};
 use crate::proc::{self, CPUData, NetworkData};
 use crate::telemetry::events::ReportMetrics;
+use crate::{
```
duncanista, line 3: Might be good to either merge imports, or, just pass the telemetry import to the one above

duncanista (approved):

PR #588, opened 2025-03-05
fix(gitlab): private image builds

duncanista:
# What?

Fixes private image builds

# Notes

Now it will just check the latest version on sandbox for the layer, which is what we normally use for testing RCs

### DIFF
--- a/.gitlab/scripts/build_layer.sh
+++ b/.gitlab/scripts/build_layer.sh
@@ -42,7 +42,6 @@ docker_build() {
 
     cp $EXTENSION_PATH/datadog_extension.zip $TARGET_DIR/datadog_extension-${FILE_SUFFIX}.zip
     unzip $EXTENSION_PATH/datadog_extension.zip -d $TARGET_DIR/datadog_extension-${FILE_SUFFIX}
-    rm -rf ${EXTENSION_PATH}
 }
 
 docker_build $ARCHITECTURE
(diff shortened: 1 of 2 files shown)


.gitlab/scripts/build_layer.sh
```diff
@@ -42,7 +42,6 @@ docker_build() {
 
     cp $EXTENSION_PATH/datadog_extension.zip $TARGET_DIR/datadog_extension-${FILE_SUFFIX}.zip
     unzip $EXTENSION_PATH/datadog_extension.zip -d $TARGET_DIR/datadog_extension-${FILE_SUFFIX}
-    rm -rf ${EXTENSION_PATH}
```
astuyve, line 45: why?

astuyve (approved):

.gitlab/scripts/build_layer.sh
duncanista: previously we didnt remove it, I added this line to only have zips and not dirs in the layers folder, but the private images require this to work properly

.gitlab/scripts/build_private_image.sh
```diff
@@ -29,12 +29,12 @@ else
...
 fi
 
-LAYER_NAME="Datadog-Extension$SUFFIX"
+LAYER_NAME="Datadog-Extension"
 if [ -z "$LAYER_SUFFIX" ]; then
     printf "Building container images tagged without suffix\n"
 else
     printf "Building container images tagged with suffix: ${LAYER_SUFFIX}\n"
-    LAYER_NAME="${LAYER_NAME}-${LAYER_SUFFIX}${SUFFIX}"
+    LAYER_NAME="${LAYER_NAME}-${LAYER_SUFFIX}"
```
apiarian-datadog, line 37: we're still only looking for the version number of the base layer, not the custom pipeline suffix layer. is that okay?

PR #591, opened 2025-03-06
feat(bottlecap): Update Step Functions trace extraction

avedmala:
### What

Updates the creation of trace context from a Step Function execution's context object to...
1. Make use of `State.RetryCount` and `Execution.RedriveCount` for inferable parent ID generation in a backwards compatible manner (related to https://github.com/DataDog/datadog-lambda-python/pull/559)
2. Support the new trace context propagation for multi-level trace merging. These are cases where we're receiving a Step Function event but that Step Function has another parent service which can be either a Lambda or another Step Function (related https://github.com/DataDog/datadog-lambda-python/pull/537)

### Why

Parity of https://github.com/DataDog/datadog-agent/pull/34264
1. Using these new values for parent ID generation will prevent collisions with "retry spans" which are Step Function spans that parent a Lambda. Without using the `State.RetryCount` the other values we use for the hash are identical across tries
2. Multi-level trace merging is the future, it lets us merge an arbitrary number of Lambda and Step Function traces. The previous approach only lets us do a max depth of 2 services, losing the context after that. We're able to always preserve some about the top-most service or the root service to keep the Trace ID intact while using the context object to infer the parent ID.

### Testing
Expanded unit tests to account for new cases

Live [trace](https://dd.datad0g.com/apm/trace/67bf8e940e01810a48f7a6ab796c8fb6?graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=5660659611008214469&timeHint=1741299793519) of the nested step function trace merging (SFN -> SFN -> Lambda)
Screenshot 2025-03-06 at 5 37 33 PM

Live [trace](https://dd.datad0g.com/apm/trace/67ca23940000000005c88b1e34b10f25?graphType=waterfall&panel_tab=flamegraph&shouldShowLegend=true&sort=time&spanID=4604307312880118210&timeHint=1741300628495.9663) of the lambda root trace merging (Lambda -> SFN -> SFN -> Lambda)
Screenshot 2025-03-06 at 5 41 10 PM

### DIFF
--- a/bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
+++ b/bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
@@ -26,7 +26,7 @@ pub struct StepFunctionEvent {
     pub trace_id: Option<String>,
     #[serde(rename = "x-datadog-tags")]
     pub trace_tags: Option<String>,
-    #[serde(rename = "RootExecutionID")]
+    #[serde(rename = "RootExecutionId")]
     pub root_execution_id: Option<String>,
     #[serde(rename = "serverless-version")]
     pub serverless_version: Option<String>,
@@ -131,11 +131,29 @@ impl Trigger for StepFunctionEvent {
 impl StepFunctionEvent {
     #[must_use]
     pub fn get_span_context(&self) -> SpanContext {
-        let (lo_tid, hi_tid) = Self::generate_trace_id(self.execution.id.clone());
-        let tags = HashMap::from([(
-            DATADOG_HIGHER_ORDER_TRACE_ID_BITS_KEY.to_string(),
-            format!("{hi_tid:x}"),
-        )]);
+        let (lo_tid, tags) =
+            if let (Some(trace_id), Some(trace_tags)) = (&self.trace_id, &self.trace_tags) {
+                // Lambda Root
+                let lo_tid = trace_id.parse().expect("Failed to parse trace_id to u64");
+                let tags = Self::extract_trace_tags(trace_tags);
+                (lo_tid, tags)
+            } else if let Some(root_execution_id) = &self.root_execution_id {
+                // Nested
+                let (lo_tid, hi_tid) = Self::generate_trace_id(root_execution_id.clone());
+                let tags = HashMap::from([(
+                    DATADOG_HIGHER_ORDER_TRACE_ID_BITS_KEY.to_string(),
+                    format!("{hi_tid:x}"),
+                )]);
+                (lo_tid, tags)
+            } else {
+                // Normal
+                let (lo_tid, hi_tid) = Self::generate_trace_id(self.execution.id.clone());
+                let tags = HashMap::from([(
+                    DATADOG_HIGHER_ORDER_TRACE_ID_BITS_KEY.to_string(),
+                    format!("{hi_tid:x}"),
+                )]);
+                (lo_tid, tags)
+            };
 
         let parent_id = Self::generate_parent_id(
             self.execution.id.clone(),
@@ -209,6 +227,19 @@ impl StepFunctionEvent {
             result
         }
     }
+
+    /// Extracts the `_dd.p.tid` tag from a comma-separated trace tags string
+    ///
+    fn extract_trace_tags(trace_tags: &str) -> HashMap<String, String> {
+        trace_tags
+            .split(',')
+            .filter_map(|s| s.split_once('='))
+            .find_map(|(k, v)| {
+                (k == DATADOG_HIGHER_ORDER_TRACE_ID_BITS_KEY)
+                    .then(|| (k.to_string(), v.to_string()))
+            })
+            .map_or_else(HashMap::new, |entry| HashMap::from([entry]))
+    }
 }
 
 impl ServiceNameResolver for StepFunctionEvent {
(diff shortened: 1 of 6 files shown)


bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
```diff
@@ -118,16 +131,36 @@ impl Trigger for StepFunctionEvent {
 impl StepFunctionEvent {
     #[must_use]
     pub fn get_span_context(&self) -> SpanContext {
-        let (lo_tid, hi_tid) = Self::generate_trace_id(self.execution.id.clone());
-        let tags = HashMap::from([(
-            DATADOG_HIGHER_ORDER_TRACE_ID_BITS_KEY.to_string(),
-            format!("{hi_tid:x}"),
-        )]);
+        let (lo_tid, tags) =
+            if let (Some(trace_id), Some(trace_tags)) = (&self.trace_id, &self.trace_tags) {
```
avedmala, line 135: We wouldn't need this if-else checking if we defined separate events rather than including all fields in one and making some optional

**But**, this way we're keeping all the complexity in this file, otherwise we'd have to match on all the different types of events in the `span_inferrer`

This logic will need infrequent edits so it made sense to me to pull it down similar to what we do in the python layer

bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
```diff
@@ -14,13 +14,6 @@ use crate::{
     },
 };
 
-#[allow(clippy::module_name_repetitions)]
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
-pub struct LegacyStepFunctionEvent {
```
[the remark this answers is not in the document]
duncanista: Oh good catch, it was being covered, just not through this struct

bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
```diff
@@ -70,8 +79,12 @@ impl Trigger for StepFunctionEvent {
...
         Self: Sized,
     {
-        // Check first if the payload is a Legacy Step Function event
-        let p = payload.get("Payload").unwrap_or(payload);
+        // Check if the payload is a Legacy Step Function event and also a JSONata event
+        let p = payload
+            .get("Payload")
+            .unwrap_or(payload)
+            .get("_datadog")
+            .unwrap_or(payload.get("Payload").unwrap_or(payload));
```
duncanista, line 87: Too confusing, maybe let's make `Payload` a constant?

duncanista: I think this can be simplified

bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
```diff
@@ -56,7 +61,11 @@ impl Trigger for StepFunctionEvent {
     where
         Self: Sized,
     {
-        let p = payload.get("Payload").unwrap_or(&payload);
+        let p = payload
+            .get("Payload")
+            .unwrap_or(&payload)
+            .get("_datadog")
+            .unwrap_or(payload.get("Payload").unwrap_or(&payload));
```
duncanista, line 68: This can be simplified

bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
```diff
@@ -118,16 +131,36 @@ impl Trigger for StepFunctionEvent {
...
     pub fn get_span_context(&self) -> SpanContext {
-        let (lo_tid, hi_tid) = Self::generate_trace_id(self.execution.id.clone());
-        let tags = HashMap::from([(
-            DATADOG_HIGHER_ORDER_TRACE_ID_BITS_KEY.to_string(),
-            format!("{hi_tid:x}"),
-        )]);
+        let (lo_tid, tags) =
+            if let (Some(trace_id), Some(trace_tags)) = (&self.trace_id, &self.trace_tags) {
+                // Lambda Root
+                let lo_tid = trace_id.parse().expect("Failed to parse trace_id to u64");
```
duncanista, line 137: Is there a way to avoid using `expect` here?

bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
```diff
@@ -118,16 +131,36 @@ impl Trigger for StepFunctionEvent {
...
+                )]);
+                (lo_tid, tags)
+            } else {
+                // Normal
+                let (lo_tid, hi_tid) = Self::generate_trace_id(self.execution.id.clone());
+                let tags = HashMap::from([(
+                    DATADOG_HIGHER_ORDER_TRACE_ID_BITS_KEY.to_string(),
+                    format!("{hi_tid:x}"),
+                )]);
+                (lo_tid, tags)
```
duncanista, line 155: This can be simplified into one block where you just check for the ID you're going to pass

duncanista (commented):
Left some comments!

bottlecap/src/lifecycle/invocation/triggers/step_function_event.rs
avedmala: refactored so we use `Self::generate_trace_id` as a fallback if the parse fails

PR #593, opened 2025-03-07
feat: Shutdown blocks for report

astuyve:
In the case of a function running once every few hours, we may not block for the platform.report event.

this PR modifies the loop to wait for this event so we set some enhanced metrics after a single function execution.


duncanista (commented):
Great, thanks for modifying this – glad we caught it before it went out!

duncanista (approved):

PR #604, opened 2025-03-11
aj/use secret region

### LINKED ISSUE #594: "Error decrypting key" when upgrading to recent versions of extension
My team is installing the datadog lambda extension in a container image, and we have noticed an issue that is causing us some trouble when deploying it to AWS Lambda.
We have the `DD_API_KEY_SECRET_ARN` environment variable set to the arn of a secret manager were the datadog key is stored.
Previously we were using version v53 of the extension, and `datadog-lambda == 5.88.0` and everything was working as expected.
However, recently we decided to bump the lambda extension version and the python datadog lambda library as well.

We have done so to v72 and 6.106.0 respectively, but now, after deploying, the following error is captured in Cloudwatch:

```
DD_EXTENSION \| ERROR \| Error decrypting key: {     "Message": "User: arn:aws:sts::<accountId>:assumed-role/core-service-develop-us-east-1-lambdaRole/core-service-develop-core is not authorized to perform: secretsmanager:GetSecretValue on resource: arn:aws:secretsmanager:us-west-2:<accountId>:secret:uptogether/datadog/api_key-4QC7nT because no identity-based policy allows the secretsmanager:GetSecretValue action",     "__type": "AccessDeniedException" }
--
DD_EXTENSION \| ERROR \| Failed to resolve secrets, Datadog extension will be idle
```

The `GetSecretValue` permission is granted (otherwise it would not work in the past as well), but after the update, datadog is not able to capture logs.

## Additional notes
* We have tested the same by using v69 and `datadog-lambda == 6.105.0` but again, the same error is captured.
* I was able to upgrade to v66 and `datadog-lambda == 6.102.0` and the logs were captured as expected, so maybe this is an issue related to the Next Generation of Datadog Lambda Extension?

## Additional question
I would also like to take the opportunity to ask something about the documentation that is not fully clear (not sure if this is the right place, but if not, I would appreciate if someone can point me out to the right place)
The documentation in [Installation steps for Container Images](https://docs.datadoghq.com/serverless/aws_lambda/installation/python/?tab=containerimage) states:

```
Note that the minor version of the datadog-lambda package always matches the layer version. For example, datadog-lambda v0.5.0 matches the content of layer version 5.
```

However, it seems that the layer version is not longer matching the datadog-lambda-python package. (For instance, current version of the extension is 73 and the most recent python version is [v6.106.0](https://github.com/DataDog/datadog-lambda-python/releases/tag/v6.106.0)
Is there a guide or something similar to know which version should be used in each case?
As of now, I have been trying to use versions that were released in similar days, assuming they were meant to go together, but I can not ensure this is true. I would appreciate some guidance here.

astuyve:
Fixes #594

you can verify with [this function](https://us-west-2.console.aws.amazon.com/lambda/home?region=us-west-2#/functions/secrets-manager-dev-hello?subtab=envVars&tab=configure) using a secret in us-east-1, while running in us-west-2.

Metrics appearing:
image

### DIFF
--- a/bottlecap/src/secrets/decrypt.rs
+++ b/bottlecap/src/secrets/decrypt.rs
@@ -53,7 +53,18 @@ pub async fn resolve_secrets(config: Arc<Config>, aws_config: &mut AwsConfig) ->
             }
 
             let decrypted_key = if config.kms_api_key.is_empty() {
-                decrypt_aws_sm(&client, config.api_key_secret_arn.clone(), aws_config).await
+                let secret_region = config
+                    .kms_api_key
+                    .split(":")
+                    .nth(3)
+                    .unwrap_or(&aws_config.region);
+                decrypt_aws_sm(
+                    &client,
+                    config.api_key_secret_arn.clone(),
+                    aws_config,
+                    secret_region.to_string(),
+                )
+                .await
             } else {
                 decrypt_aws_kms(&client, config.kms_api_key.clone(), aws_config).await
             };
@@ -107,6 +118,7 @@ async fn decrypt_aws_kms(
 
     let headers = build_get_secret_signed_headers(
         aws_config,
+        aws_config.region.clone(),
         RequestArgs {
             service: "kms".to_string(),
             body: json_body,
@@ -128,6 +140,7 @@ async fn decrypt_aws_kms(
 
         let headers = build_get_secret_signed_headers(
             aws_config,
+            aws_config.region.clone(),
             RequestArgs {
                 service: "kms".to_string(),
                 body: json_body,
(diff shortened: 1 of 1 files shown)


bottlecap/src/secrets/decrypt.rs
```diff
@@ -53,7 +53,18 @@ pub async fn resolve_secrets(config: Arc<Config>, aws_config: &mut AwsConfig) ->
             }
 
             let decrypted_key = if config.kms_api_key.is_empty() {
-                decrypt_aws_sm(&client, config.api_key_secret_arn.clone(), aws_config).await
+                let secret_region = config
```
duncanista, line 56: Might be good to have a note that secrets might not be from the same region

astuyve: can do that

bottlecap/src/secrets/decrypt.rs
```diff
@@ -53,7 +53,18 @@ pub async fn resolve_secrets(config: Arc<Config>, aws_config: &mut AwsConfig) ->
...
+                    .unwrap_or(&aws_config.region);
+                decrypt_aws_sm(
+                    &client,
+                    config.api_key_secret_arn.clone(),
+                    aws_config,
+                    secret_region.to_string(),
+                )
+                .await
             } else {
                 decrypt_aws_kms(&client, config.kms_api_key.clone(), aws_config).await
```
duncanista, line 69: Would KMS have the same issue? we are sending the `aws_config` which has the region config for the lambda

duncanista (approved):
LGTM – thanks for answering my comments!

PR #607, opened 2025-03-13
feat: Add readiness debug log

astuyve:
Adds a debug message with our init duration for the extension specifically, measured by the time main begins to the time we call `/next`
aids in debugging/disambiguation cases
image

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -202,7 +202,16 @@ async fn main() -> Result<()> {
         .map_err(|e| Error::new(std::io::ErrorKind::InvalidData, e.to_string()))?;
 
     if let Some(resolved_api_key) = resolve_secrets(Arc::clone(&config), &mut aws_config).await {
-        match extension_loop_active(&aws_config, &config, &client, &r, resolved_api_key, start_time).await {
+        match extension_loop_active(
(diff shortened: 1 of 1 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -184,6 +184,7 @@ fn build_function_arn(account_id: &str, region: &str, function_name: &str) -> St
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    let start_time = Instant::now();
```
duncanista, line 187: We also do `Instant::now()` in `load_configs` for `aws_config` sandbox init start, maybe worth refactoring that if it's going to be used here? Either we use this one and make the `aws_config` to have it available, or we use the one the `aws_config` has, do as you prefer, another alternative could be possible!

astuyve: yup good point, combined.

duncanista (approved):
LGTM – wondering if we should merge the log that says "vXX version of the Datadog Extension", we can just add version and time it started in the same one, me thinks

astuyve:
I don't think so, because a lot can go wrong between when that log is emitted and we're done booting – I want to make sure we get the version number even if we fail to start, or fail over, for some reason.

PR #608, opened 2025-03-13
feat: parse env and service as strings or ints

### LINKED ISSUE #606: Failed to parse DD_ENV with integer value
Hi there 👋 

We noticed some failures in the extension logs related to parsing `DD_ENV` when it's set to a number, similar to https://github.com/DataDog/datadog-lambda-extension/issues/491

```
Failed to parse Datadog config: invalid type: found unsigned int `1111111111111`, expected a string for key ""ENV"" in `DD_` environment variable(s)
```

(The number itself isn't actually all `1`. I replaced it since it's an account number I'd prefer not to show)

I believe the fix would be to add a `serde(deserialize_with)` on `env`, similar to `version` here: https://github.com/DataDog/datadog-lambda-extension/blob/v74/bottlecap/src/config/mod.rs#L115-L118

Would there be any interest in having the same parsing logic for `DD_ENV` as `DD_VERSION`? Understandably, it shouldn't really be common for `DD_ENV` to be a number on its own, but for several reasons probably not worth getting into here, we have a customer environment that is kinda "stuck" with that as the case. We're happy to work around it in some way, if this isn't a desirable change on your end.

Thanks!

astuyve:
Fixes #606, also accounts for DD_SERVICE which, in unusual but possible circumstances, could be an int

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -812,14 +812,16 @@ pub mod tests {
     }
 
     #[test]
-    fn parse_number_or_string_dd_version_env() {
+    fn parse_number_or_string_env_vars() {
         figment::Jail::expect_with(|jail| {
             jail.clear_env();
             jail.set_env("DD_VERSION", "123");
             jail.set_env("DD_ENV", "123456890");
+            jail.set_env("DD_SERVICE", "123456");
             let config = get_config(Path::new(""), MOCK_REGION).expect("should parse config");
             assert_eq!(config.version.expect("failed to parse DD_VERSION"), "123");
             assert_eq!(config.env.expect("failed to parse DD_ENV"), "123456890");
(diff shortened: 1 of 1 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -810,12 +812,14 @@ pub mod tests {
     }
 
     #[test]
-    fn parse_dd_version() {
+    fn parse_number_or_string_dd_version_env() {
         figment::Jail::expect_with(|jail| {
             jail.clear_env();
             jail.set_env("DD_VERSION", "123");
+            jail.set_env("DD_ENV", "123456890");
```
duncanista, line 819: add one for service too? maybe change it to `parse_number_or_string_env_vars`

duncanista (approved):

bottlecap/src/config/mod.rs
astuyve: sure

PR #609, opened 2025-03-13
Alexgallotta/svls 5285 lwa proxy tmp

alexgallotta:
Add a proxy that will forward requests, capturing and de serializing the payload to trigger lifecycle events
This would allow tracing support for Lambda Web Adapter


alexgallotta:
wrong branch, this is the active one
https://github.com/DataDog/datadog-lambda-extension/pull/638

PR #610, opened 2025-03-13
aj/support log tags

### LINKED ISSUE #515: ddtags is not honored in v68 of Next Generation version
The `ddtags` comes across as a normal log message, but they are not parsed as tags in Datadog logs. Setting back to old version via `DD_EXTENSION_VERSION=compatibility` does parse correctly as expected.

astuyve:
Fixes #515 

You can reproduce by creating a new lambda function, enabling `JSON` logs in Lambda, and then logging something like this:
```
  console.log({ddtags: 'aj:test,aj_type:added_tags', message: {input: event, more_items: "are working"}});
```
  
You should be able to search for those tags in the app:
image


duncanista (approved):
This looks great, I'd like to see the performance implications of trying to parse every message into a json object, I'd expect serde to do a great job here already tho

PR #615, opened 2025-03-18
refactor tracing context extraction

alexgallotta:
- remove some redundant classes
- split datadog and text map context extraction in 2 different files


duncanista:
Closing because this logic is now owned in [dd-trace-rs](https://github.com/DataDog/dd-trace-rs)

PR #616, opened 2025-03-18
do not exit the loop if no context is found

alexgallotta:


duncanista (approved):
Add a description to this PR

PR #619, opened 2025-03-24
DynamoDB Span Pointers

nhulston:
### What does this PR do?

Adds span pointers to spans for Lambdas triggered by DynamoDB streams. This change will affect the downstream case for Universal Instrumentation Lambda runtimes (Java, .NET, Golang).

Span pointers are similar to [Span Links](https://docs.datadoghq.com/tracing/trace_collection/span_links/), but for cases when it is impossible to pass the Trace ID and Span ID between the spans that need to be linked.

When the calculated hashes for the upstream and downstream lambdas match, the Datadog frontend will automatically link the two traces together.

Upstream case (requires tracer to be instrumented): 
Screenshot 2025-03-24 at 1 46 12 PM

Downstream case (this code):
Screenshot 2025-03-24 at 1 46 56 PM

When clicking on the linked span, a new tab opens linking to the opposite Lambda function.

Mostly manual testing, but I also added unit tests. There are also functional tests in https://github.com/DataDog/serverless-e2e-tests -- DynamoDB is TODO but will be implemented soon.


bottlecap/src/lifecycle/invocation/triggers/dynamodb_event.rs
```diff
@@ -143,6 +170,58 @@ impl ServiceNameResolver for DynamoDbRecord {
...
+
+            let (k1, attr1) = keys[0];
+            // If unable to get string value, just return None
+            let v1 = attr1.to_string()?;
+
+            let (k2, attr2) = keys[1];
+            let v2 = attr2.to_string()?;
+
+            (k1.clone(), v1, k2.clone(), v2)
+        };
```
duncanista, line 207:
```suggestion
       let (primary_key1, value1, primary_key2, value2) = match self.dynamodb.keys.len() {
    1 => {
        let (key, attr_value) = self.dynamodb.keys.iter()
            .next()
            .expect("No DynamoDB keys found");

        (key.clone(), attr_value.to_string()?, String::new(), String::new())
    },
    _ => {
        let mut sorted_keys: Vec<_> = self.dynamodb.keys.iter()
            .collect::<Vec<_>>()
            .into_iter()
            .sorted_by_key(|k| k.0.clone())
            .collect();

        let (k1, attr1) = sorted_keys[0];
        let (k2, attr2) = sorted_keys[1];

        (k1.clone(), attr1.to_string()?, k2.clone(), attr2.to_string()?)
    }
};
```

duncanista (approved):
LGTM – left a comment

PR #622, opened 2025-03-25
Add DSM and Profiling endpoints

astuyve:
- **feat: Support DSM proxy endpoint**
- **feat: profiling support**
- **feat: add additional tags**


bottlecap/src/traces/trace_agent.rs
```diff
@@ -204,6 +211,24 @@ impl TraceAgent {
...
+            (&Method::POST | &Method::PUT, DSM_AGENT_PATH) => {
+                match Self::handle_dsm_proxy(config, req).await {
+                    Ok(result) => Ok(result),
+                    Err(err) => log_and_create_http_response(
+                        &format!("DSM endpoint error: {err}"),
+                        StatusCode::INTERNAL_SERVER_ERROR,
+                    ),
+                }
+            }
+            (&Method::POST | &Method::PUT, PROFILING_ENDPOINT_PATH) => {
```
astuyve, line 223: may just need post here

bottlecap/src/traces/trace_agent.rs
```diff
@@ -298,6 +325,108 @@ impl TraceAgent {
...
+                && name.as_str().to_lowercase() != "content-length"
+            {
+                if let Ok(header_value) =
+                    reqwest::header::HeaderValue::from_str(value.to_str().unwrap_or_default())
+                {
+                    request_builder = request_builder.header(name.as_str(), header_value);
+                }
+            }
+        }
+        request_builder = request_builder.header("DD-API-KEY", &config.api_key);
```
duncanista, line 365: This should use the resolved API key, not the config one

bottlecap/src/traces/trace_agent.rs
```diff
@@ -298,6 +325,108 @@ impl TraceAgent {
...
+            Err(err) => {
+                return log_and_create_http_response(
+                    &format!("Error reading request body: {err}"),
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                );
+            }
+        };
+
+        let target_url = format!("https://{}.{}{}", backend_domain, config.site, backend_path);
+        let client = http_client::get_client(config.clone());
```
duncanista, line 351: Should we avoid creating a client every time we send a payload?

bottlecap/src/traces/trace_agent.rs
```diff
@@ -298,6 +325,108 @@ impl TraceAgent {
...
+                    request_builder = request_builder.header(name.as_str(), header_value);
+                }
+            }
+        }
+        request_builder = request_builder.header("DD-API-KEY", &config.api_key);
+        request_builder = request_builder.header(
+            "X-Datadog-Additional-Tags",
+            format!(
+                "_dd.origin:lambda;functionname:{}",
+                env::var("AWS_LAMBDA_FUNCTION_NAME").unwrap_or_default()
```
duncanista, line 370: Can't you reuse the tag provider which should exist in the agent to get this?

bottlecap/src/traces/trace_agent.rs
```diff
@@ -298,6 +325,108 @@ impl TraceAgent {
...
+                if let Ok(header_value) =
+                    reqwest::header::HeaderValue::from_str(value.to_str().unwrap_or_default())
+                {
+                    request_builder = request_builder.header(name.as_str(), header_value);
+                }
+            }
+        }
+        request_builder = request_builder.header("DD-API-KEY", &config.api_key);
+        request_builder = request_builder.header(
+            "X-Datadog-Additional-Tags",
```
duncanista, line 367: Maybe let's make this a constant

bottlecap/src/traces/trace_agent.rs
```diff
@@ -298,6 +325,108 @@ impl TraceAgent {
...
+            format!(
+                "_dd.origin:lambda;functionname:{}",
```
duncanista, line 369: Should we add then all the other tags that the tags provider has?

duncanista (commented):
Everything looks good so far, left some comments – we might want to sync with the DSM/Profiling teams to fix the overhead caused by this pass-through proxy in the future

bottlecap/src/traces/trace_agent.rs
astuyve: Hmm. We could, but the docs sheet I was provided doesn't ask for them so I'm not inclined to add extra tags if they aren't used by those products.

We can always add them later.

bottlecap/src/traces/trace_agent.rs
astuyve: Yup

duncanista (approved):

PR #623, opened 2025-03-26
[dotnet] don't override env vars if set by user

lucaspimentel:
Change the way we set .NET environment variables in `datadog_wrapper`, so we only set them if they are not already set. This prevents the script from overriding env vars set by the user.

For example, setting `CORECLR_ENABLE_PROFILING=0` does not disable the tracer, because `datadog_wrapper` then sets `CORECLR_ENABLE_PROFILING=1`. See https://github.com/DataDog/dd-trace-dotnet-aws-lambda-layer/issues/14.


hghotra (approved):
✅

joeyzhao2018 (approved):
:shipit:

duncanista (approved):

PR #625, opened 2025-03-27
fix: handle non-nested 'message' values for structured logging without using Lambda's JSON logging feature

### LINKED ISSUE #620: Losing Logs record fields
Hi there! 👋 

We've recently managed to make our PHP based Lambda function work with:

- ddtrace-php sidecar trace sender feature on
- next-gen v75 lambda extension

Everything seems to be working fine, we get both traces and logs in Datadog, however they are not correlated. After investigation, I believe it is something to do with "losing logs data" somewhere in the process.

Our application is a PHP Symfony app, using Monolog as logger and have logs injection turned on, and we do log as JSON, so here is a typical log record in Cloudwatch:

```
{
    "message": "Read existing security token from the session.",
    "context": {
        "key": "_security_admin",
        "token_class": "Symfony\\Component\\Security\\Http\\Authenticator\\Token\\PostAuthenticationToken",
        "dd.trace_id": "13204366237012237818",
        "dd.span_id": "15763279168591393237",
        "dd.service": "notifications-api",
        "dd.version": "f8f423a934f5893992413072f2527cd6ff63bf11",
        "dd.env": "dev"
    },
    "level": 100,
    "level_name": "DEBUG",
    "channel": "security",
    "datetime": "2025-03-25T01:48:57.102952+00:00",
    "extra": {
        "X-CORRELATION-ID": "1f0091b5-0ac1-6a6e-9e44-c9dbdb3c4b5e",
        "X-REQUEST-ID": "1f0091b5-0ac1-6b36-8acf-c9dbdb3c4b5e"
    }
}
```

However when looking at this same record in Datadog console, information is missing:
Image

I would expect to see the `context` at least in Datadog console, 
Because we are using PHP which isn't natively supported on Lambda, do we need an additional setup?
Or are we simply missing something? 😄

astuyve:
The original code accounted for Lambda's own structured logging feature, but some users are using powertools or custom loggers and emitting one simple JSON log without the nested `message` param added by Lambda. The logic looking for Lambda's JSON logging version would inadvertently capture people using their own loggers but only parse/include the value of the `message` string, ignoring other attributes.

This handles both and includes a test.
You can see a custom logger + ddtags works in text mode:
image

As well as in JSON logging mode:
image

This also fixes an issue where the `level` wasn't being honored. It works now. Here it is for text mode:
image

And JSON:
image

Fixes #620

### DIFF
--- a/bottlecap/src/logs/lambda/processor.rs
+++ b/bottlecap/src/logs/lambda/processor.rs
@@ -220,65 +220,89 @@ impl LambdaProcessor {
     }
 
     fn get_intake_log(&mut self, mut lambda_message: Message) -> Result<IntakeLog, Box<dyn Error>> {
-        let request_id = match lambda_message.lambda.request_id {
-            // Log already has a `request_id`
+        // Assign request_id from message or context if available
+        lambda_message.lambda.request_id = match lambda_message.lambda.request_id {
             Some(request_id) => Some(request_id.clone()),
-            // Default to the `request_id` we've seen so far, if any.
-            None => (!&self.invocation_context.request_id.is_empty())
-                .then(|| self.invocation_context.request_id.clone()),
-        };
-
-        lambda_message.lambda.request_id = request_id;
-        // Test if message is a JSON object, it may have tags we need to pass along.
-        // Don't care about host, service, and status, since we'll override them.
-        let parsed_message =
-            serde_json::from_str::<serde_json::Value>(lambda_message.message.as_str());
-        let log = match parsed_message {
-            Ok(serde_json::Value::Object(mut obj)) => {
-                let mut tags = self.tags.clone();
-                let final_message = if let Some(inner_message) = obj.get_mut("message") {
-                    if let Some(serde_json::Value::String(message_tags)) =
-                        inner_message.get("ddtags")
-                    {
-                        tags.push(',');
-                        tags.push_str(message_tags);
-                        if let Some(inner_message_object) = inner_message.as_object_mut() {
-                            inner_message_object.remove("ddtags");
-                        }
-                    }
-                    inner_message.to_string()
+            None => {
+                if self.invocation_context.request_id.is_empty() {
+                    None
                 } else {
-                    lambda_message.message.clone()
-                };
-                IntakeLog {
-                    hostname: self.function_arn.clone(),
-                    source: LAMBDA_RUNTIME_SLUG.to_string(),
-                    service: self.service.clone(),
-                    tags,
-                    message: Message {
-                        message: final_message,
-                        lambda: lambda_message.lambda,
-                        timestamp: lambda_message.timestamp,
-                        status: lambda_message.status,
-                    },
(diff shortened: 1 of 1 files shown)


bottlecap/src/logs/lambda/processor.rs
```diff
@@ -220,67 +220,91 @@ impl LambdaProcessor {
...
+        json_obj: &mut serde_json::Map<String, serde_json::Value>,
+        tags: &mut String,
+        original_message: String,
+    ) -> String {
+        // Check for top-level ddtags
+        if let Some(serde_json::Value::String(message_tags)) = json_obj.get("ddtags") {
+            tags.push(',');
+            tags.push_str(message_tags);
+            json_obj.remove("ddtags");
+            return serde_json::to_string(json_obj).unwrap_or(original_message);
```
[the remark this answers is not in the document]
astuyve: I don't think we need to do that here because we deserialized the object in the first place, so reserializing it should never fail

bottlecap/src/logs/lambda/processor.rs
```diff
@@ -220,67 +220,91 @@ impl LambdaProcessor {
...
-                    tags,
-                    message: Message {
-                        message: final_message,
-                        lambda: lambda_message.lambda,
-                        timestamp: lambda_message.timestamp,
-                        status: lambda_message.status,
-                    },
-                }
+
+        let log = if let Ok(serde_json::Value::Object(mut json_obj)) = parsed_json {
```
duncanista, line 239: Maybe a match is better here?

bottlecap/src/logs/lambda/processor.rs
```diff
@@ -220,67 +220,91 @@ impl LambdaProcessor {
...
-            // Default to the `request_id` we've seen so far, if any.
-            None => (!&self.invocation_context.request_id.is_empty())
-                .then(|| self.invocation_context.request_id.clone()),
+            None => {
+                if self.invocation_context.request_id.is_empty() {
+                    None
+                } else {
+                    Some(self.invocation_context.request_id.clone())
+                }
+            }
```
duncanista, line 232: Is this change necessary?

astuyve: Clippy:
```
warning: unnecessary boolean `not` operation
   --> src/logs/lambda/processor.rs:227:17
    |
227 | /                 if !self.invocation_context.request_id.is_empty() {
228 | |                     Some(self.invocation_context.request_id.clone())
229 | |                 } else {
230 | |                     None
231 | |                 }
    | |_________________^
    |
    = help: remove the `!` and swap the blocks of the `if`/`else`

```

duncanista: I mean, what about the previous code?

astuyve: Oh, no it's not strictly necessary, I just thought it was easier to understand.

duncanista (approved):

PR #627, opened 2025-03-27
renaming and small refactor

alexgallotta:
some renaming without functional changes

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -55,7 +55,7 @@ pub struct Processor {
     // Helper to infer span information
     inferrer: SpanInferrer,
     // Current invocation span
-    pub span: Span,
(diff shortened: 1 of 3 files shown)


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -55,7 +55,7 @@ pub struct Processor {
     // Helper to infer span information
     inferrer: SpanInferrer,
     // Current invocation span
-    pub span: Span,
+    pub aws_lambda_span: Span,
```
duncanista, line 58: What about `invocation_span`? Since this is the `invocation/processor`?

alexgallotta: aws.lambda.span is very specific, and I think it fits this case the best. There is no other span that would end in that place, right?

duncanista:
Closing due to stale

PR #628, opened 2025-03-27
feat: add llmobs proxy paths to trace agent

sabrenner:
Adds LLM Observability Eval Metrics and Spans endpoints to the trace agent. They are EVP proxy paths to begin with, and then we forward to our actual backend (proxy path without `/evp_proxy/v2`). 

I verified this by using it as a layer in a small lambda I have for LLM Observability testing - with a debug log I added, I noted
```
DD_EXTENSION | DEBUG | Got response from https://llmobs-intake.datadoghq.com/api/v2/llmobs backend: 202 Accepted
```
and was able to see the span in LLM Observability. This was verified for both Python and Node.js.


duncanista (commented):
What are the endpoint limits? What's the max size package? Does the tracer already batch the data accordingly?

PR #634, opened 2025-04-01
aj/hyper 1

astuyve:
Hyper1 migration


bottlecap/src/traces/trace_agent.rs
```diff
@@ -148,50 +150,78 @@ impl TraceAgent {
...
         let addr = SocketAddr::from(([127, 0, 0, 1], port));
-        let server_builder = Server::try_bind(&addr)?;
-
-        let server = server_builder.serve(make_svc);
+        let listener = tokio::net::TcpListener::bind(&addr).await?;
 
         debug!("Trace Agent started: listening on port {TRACE_AGENT_PORT}");
         debug!(
-            "Time taken start the Trace Agent: {} ms",
+            "Time taken start the Mini Agent: {} ms",
```
duncanista:
```suggestion
            "Time taken start the Trace Agent: {} ms",
```

bottlecap/src/traces/trace_agent.rs
```diff
@@ -148,50 +150,78 @@ impl TraceAgent {
...
+                        Some(finished) => finished,
+                        None => std::future::pending().await,
+                    }
+                } => match finished {
+                    Err(e) if e.is_panic() => {
+                        std::panic::resume_unwind(e.into_panic());
+                    },
+                    Ok(()) | Err(_) => continue,
+                },
+            };
```
duncanista, line 208: Looks like we have the same thing in other places, does it make sense to make it a method/helper?

Also took me a while to understand this, too many matches, should we add some comments?

bottlecap/src/lifecycle/listener.rs
```diff
@@ -32,33 +32,67 @@ impl Listener {
...
-        let make_svc = make_service_fn(move |_| {
+        let service = service_fn(move |req| {
             let invocation_processor = invocation_processor.clone();
 
-            let service = service_fn(move |req| Self::handler(req, invocation_processor.clone()));
-
-            async move { Ok::<_, Infallible>(service) }
+            Self::handler(
+                req.map(hyper_migration::Body::incoming),
+                invocation_processor.clone(),
```
apiarian-datadog, line 40: rust question: why do we clone invocation processor twice?

bottlecap/src/traces/trace_agent.rs
astuyve: It's just a server loop. I can refactor it in another PR

bottlecap/src/lifecycle/listener.rs
astuyve: Good catch - previously we moved everything (with `async move`) so the clone was necessary. Almost all of these are ARCs so they're cheap/free to clone, but I'll remove them anyway!

duncanista (approved):

PR #638, opened 2025-04-04
Alexgallotta/svls 5285 lwa proxy hyper1

alexgallotta:
Add a proxy that will forward requests, capturing and de serializing the payload to trigger lifecycle events
This would allow tracing support for Lambda Web Adapter

### DIFF
--- a/bottlecap/Cargo.toml
+++ b/bottlecap/Cargo.toml
@@ -26,6 +26,7 @@ hyper-util = { version = "0.1.10", features = [
 http-body = "0.1"
 http-body-util = "0.1" 
 lazy_static = { version = "1.5", default-features = false }
+hyper-proxy = { version = "0.9.1", default-features = false }
 log = { version = "0.4", default-features = false }
 nix = { version = "0.26", default-features = false, features = ["feature", "fs"] }
 protobuf = { version = "3.5", default-features = false }
@@ -33,7 +34,7 @@ regex = { version = "1.10", default-features = false }
 reqwest = { version = "0.12.11", features = ["json", "http2", "rustls-tls"], default-features = false }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
-thiserror = { version = "1.0", default-features = false}
+thiserror = { version = "1.0", default-features = false }
 tokio = { version = "1.37", default-features = false, features = ["macros", "rt-multi-thread"] }
 tokio-util = { version = "0.7", default-features = false }
 tracing = { version = "0.1", default-features = false }
@@ -48,7 +49,7 @@ rustls = { version = "0.23.18", default-features = false, features = ["aws-lc-rs
 rand = { version = "0.8", default-features = false }
 prost = { version = "0.11.6", default-features = false }
 zstd = { version = "0.13.3", default-features = false }
-futures = { version = "0.3.31", default-features = false}
+futures = { version = "0.3.31", default-features = false }
 
 [dev-dependencies]
 figment = { version = "0.10", default-features = false, features = ["yaml", "env", "test"] }
--- a/bottlecap/Cargo.toml
+++ b/bottlecap/Cargo.toml
@@ -26,7 +26,7 @@ hyper-util = { version = "0.1.10", features = [
 http-body = "0.1"
 http-body-util = "0.1" 
 lazy_static = { version = "1.5", default-features = false }
-hyper-proxy = { version = "0.9.1", default-features = false }
+hyper-proxy2 = { version = "0.1.0", default-features = false }
 log = { version = "0.4", default-features = false }
 nix = { version = "0.26", default-features = false, features = ["feature", "fs"] }
 protobuf = { version = "3.5", default-features = false }
--- a/bottlecap/Cargo.toml
+++ b/bottlecap/Cargo.toml
@@ -6,6 +6,7 @@ publish = false
 
 [dependencies]
 async-trait = { version = "0.1", default-features = false }
+bytes = "1.2"
 chrono = { version = "0.4", features = ["serde", "std", "now"], default-features = false }
 datadog-protos = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki/" }
 ddsketch-agent = { version = "0.1.0", default-features = false, git = "https://github.com/DataDog/saluki/" }
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -499,8 +499,8 @@ async fn extension_loop_active(
 async fn flush_all(
     logs_flusher: &LogsFlusher,
     metrics_flusher: &mut MetricsFlusher,
-    trace_flusher: &dyn TraceFlusher,
-    stats_flusher: &dyn StatsFlusher,
+    trace_flusher: &impl TraceFlusher,
+    stats_flusher: &impl StatsFlusher,
     race_flush_interval: &mut tokio::time::Interval,
 ) {
     tokio::join!(
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -344,8 +344,7 @@ async fn extension_loop_active(
         Arc::clone(&invocation_processor),
     );
 
-    // let _ = start_lwa_proxy(Arc::clone(&invocation_processor), Arc::clone(&trace_processor));
-    let _ = start_lwa_proxy(Arc::clone(config), Arc::clone(&invocation_processor));
+    let _ = start_lwa_proxy(Arc::clone(&invocation_processor));
 
     let lifecycle_listener = LifecycleListener {
         invocation_processor: Arc::clone(&invocation_processor),
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -344,7 +344,7 @@ async fn extension_loop_active(
         Arc::clone(&invocation_processor),
     );
 
-    let _ = start_lwa_proxy(Arc::clone(&invocation_processor));
+    let lwa_proxy_handle = start_lwa_proxy(Arc::clone(&invocation_processor));
 
     let lifecycle_listener = LifecycleListener {
         invocation_processor: Arc::clone(&invocation_processor),
@@ -479,6 +479,10 @@ async fn extension_loop_active(
                         }
                     }
                 }
+            }
+            if let Some(lwa_proxy_task) = lwa_proxy_handle {
+                // use with graceful shutdown after rebase with hyper 1
+                lwa_proxy_task.abort();
             }
             dogstatsd_cancel_token.cancel();
             telemetry_listener_cancel_token.cancel();
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -345,7 +345,7 @@ async fn extension_loop_active(
     );
 
     // let _ = start_lwa_proxy(Arc::clone(&invocation_processor), Arc::clone(&trace_processor));
-    let _ = start_lwa_proxy(Arc::clone(&config), Arc::clone(&invocation_processor));
+    let _ = start_lwa_proxy(Arc::clone(config), Arc::clone(&invocation_processor));
 
     let lifecycle_listener = LifecycleListener {
         invocation_processor: Arc::clone(&invocation_processor),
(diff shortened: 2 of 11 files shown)


bottlecap/src/lwa/proxy.rs
duncanista, line 1: Since this proxy is going to be used by ASM, does it make sense to just create a `proxy/mod.rs` folder which spins up the proxy, and then keep it separate to LWA specific logic which could live in a separate folder? 

Just to keep a more organized set of features.

Feel free to dismiss this, but I wonder if there is no LWA application, but we still need to use the proxy, which parts would be used from here?

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -26,6 +26,7 @@ pub struct Context {
     /// In the tracer, this is created in order to have all children spans parented
     /// to a single span. This is useful when we reparent the tracer span children to
     /// the invocation span.
+    pub runtime_done_received: bool,
```
duncanista, line 29: I think the placement of this messes up with the documentation of the field below.

alexgallotta: right! Fixing it

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -324,10 +324,33 @@ impl Processor {
...
+                        &trace_processor,
+                        &trace_agent_tx,
+                        ctx,
+                    )
+                    .await;
+                }
+            }
+        }
+    }
+
```
duncanista, line 343: Too many ifs, can we simplify it?

```suggestion
        if self.tracer_detected {
            if let Some(context) = self.enrich_spans_at_platform_done(request_id, status) {
                if context.invocation_span.trace_id != 0 && context.invocation_span.span_id != 0 {
                    self.send_extension_spans(
                        &tags_provider,
                        &trace_processor,
                        &trace_agent_tx,
                        context,
                    )
                    .await;
                }
            }
        }
    }
```

Or something like 
```
let Some(context) = ... else { ... } ;
if self.tracer_detected && (trace_id and span_id) {
   ...
}
```

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -324,10 +324,33 @@ impl Processor {
...
+    fn enrich_spans_at_paltform_done(
```
duncanista, line 344: Maybe rename this to `process_platform_runtime_done` ? You're not only enriching spans but also modifying the context and then returning a clone of it? Not sure why we would return a clone

bottlecap/src/traces/trace_agent.rs
```diff
@@ -626,3 +641,30 @@ impl TraceAgent {
         self.tx.clone()
     }
 }
+
+fn handle_reparenting(
```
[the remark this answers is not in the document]
alexgallotta: That part should be pretty self explanatory: it handles the reparenting info, getting the trace id if needed and switch the parent if the parent_id matches the id to reparent.

I will add some documentation on the way where the reparenting info lays to explain better what it does

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -379,52 +402,59 @@ impl Processor {
             cold_start_span.trace_id = context.invocation_span.trace_id;
             cold_start_span.parent_id = context.invocation_span.parent_id;
         }
+        Some(context.clone())
+    }
 
-        if self.tracer_detected {
-            let mut body_size = std::mem::size_of_val(&context.invocation_span);
-            let mut traces = vec![context.invocation_span.clone()];
+    pub async fn send_extension_spans(
```
duncanista, line 408:
```suggestion
    pub async fn send_context_spans(
```

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -542,6 +572,86 @@ impl Processor {
...
+
+        self.context_buffer
+            .sorted_reparenting_info
+            .push_back(ReparentingInfo {
+                request_id: request_id.clone(),
+                invocation_span_id: span_id,
+                parent_id_to_reparent: parent_id,
+                guessed_trace_id: 0,
+                needs_trace_id: true,
+            });
```
duncanista, line 596: Can't this be upstreamed to the context class? Not sure why this would live in the processor

bottlecap/src/lifecycle/listener.rs
```diff
@@ -118,27 +129,29 @@ impl Listener {
...
     }
 
-    async fn start_invocation_handler(
-        req: hyper_migration::HttpRequest,
+    pub async fn universal_instrumentation_start(
+        headers: &HeaderMap,
+        body: Body,
         invocation_processor: Arc<Mutex<InvocationProcessor>>,
-    ) -> http::Result<hyper_migration::HttpResponse> {
+    ) -> (u64, http::Result<hyper_migration::HttpResponse>) {
```
duncanista, line 136: Why are we returning a tuple?, looks like the `u64` is not being used?

bottlecap/src/lifecycle/invocation/processor.rs
alexgallotta: Not sure if there is an easier way to simplify this. The if let cannot be inline with other ifs, so either a match, an indentation or an early return are needed.
Since I already minimized this method to be very small and simply, I don't see the need to simplify it

bottlecap/src/lifecycle/invocation/processor.rs
alexgallotta: - half of methods are called "process", not sure it is that meaningful anymore. But I agree, I will call it enrich context
- the clone is needed since multiple mutable reference were being used. Rust is very strict on this things and the way the code is done makes it hard to do otherwise

bottlecap/src/lifecycle/invocation/processor.rs
alexgallotta: I think it's cleaner to have the Context as a a low-logic data container and the Processor the one with business logic

bottlecap/src/lifecycle/listener.rs
alexgallotta: it is used

bottlecap/src/lifecycle/listener.rs
```diff
@@ -118,27 +129,29 @@ impl Listener {
...
-                let span_context = processor.on_universal_instrumentation_start(headers, body);
+                let headers = Self::headers_to_map(headers);
 
+                let extracted_span_context = {
+                    let mut processor = invocation_processor.lock().await;
+                    processor.on_universal_instrumentation_start(headers, body)
+                };
                 let mut response = Response::builder().status(200);
 
+                let found_parent_span_id;
```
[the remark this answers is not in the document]
alexgallotta: Why? I think it's very handy keep it immutable and easier to read/understand

bottlecap/src/lifecycle/invocation/processor.rs
alexgallotta: That is a similar point: the buffer should be a buffer. Not having too complex logic is a better approach imo. There are also other issue not easily visible:

- the lock is at processor level, not context/buffer level. But this operations need to be guarded. Until the lock is moved, this might be a better place
- there can be issue with mutable references being re borrowed as mutable

bottlecap/Cargo.toml
```diff
@@ -6,6 +6,7 @@ publish = false
 
 [dependencies]
 async-trait = { version = "0.1", default-features = false }
+bytes = "1.2"
```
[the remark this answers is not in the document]
astuyve: nvm github hid the big one

astuyve (commented):
a few notes but looking really good

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -343,6 +344,8 @@ async fn extension_loop_active(
         Arc::clone(&invocation_processor),
     );
 
+    let lwa_proxy_stopper = start_lwa_proxy(Arc::clone(&invocation_processor));
```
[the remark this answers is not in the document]
alexgallotta: exactly!

bottlecap/src/lwa/proxy.rs
```diff
@@ -0,0 +1,506 @@
...
+
+    *forward_response.headers_mut() = resp_part.headers.clone();
+
+    wait_for_uni_instr(uni_instr_start, "start").await;
+    wait_for_uni_instr(uni_instr_end, "end").await;
+
+    Ok(forward_response)
+}
+
+async fn wait_for_uni_instr(uni_instr_task: Option<JoinHandle<()>>, instr_type: &str) {
```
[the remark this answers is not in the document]
alexgallotta: Yes it is intended. The join handle should be always there for LWA intercepted requests that trigger universal instrumentation only

bottlecap/src/traces/trace_agent.rs
```diff
@@ -384,14 +386,29 @@ impl TraceAgent {
...
+            let invocation_processor = invocation_processor.lock().await;
+            invocation_processor.get_reparenting_info()
+        };
 
-        // Search for trace invocation span and send it to the invocation processor
-        for chunk in &traces {
-            for span in chunk {
+        for chunk in &mut traces {
+            for span in chunk.iter_mut() {
+                debug!("Processing span: {span:?}");
```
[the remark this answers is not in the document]
alexgallotta: That was actually only for dev, I think it's too spammy. I will remove it

bottlecap/src/traces/trace_agent.rs
```diff
@@ -384,14 +386,29 @@ impl TraceAgent {
...
+            }
+        }
+
+        {
+            let mut invocation_processor = invocation_processor.lock().await;
+            for ctx_to_send in invocation_processor.update_reparenting(reparenting_info) {
+                debug!("Invocation span is now ready. Sending: {ctx_to_send:?}");
+                invocation_processor
+                    .send_ctx_spans(&tags_provider, &trace_processor, &trace_tx, ctx_to_send)
+                    .await;
```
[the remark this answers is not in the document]
alexgallotta: This sending is only once per invocation, and it should happen only during cold start. So it should never be more than once. In the worst possible case, when there are consecutive invocations after cold start that didn't receive telemetry API or any span, the capacity of the request buffer is 5 anyway

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -39,6 +40,24 @@ pub struct Context {
...
+/// The struct contains initially the span ID of an invocation span, the lambda request ID
+/// causing the invocation, and the parent found (0 if no inferred spans or existing parent were
+/// found).
+///
+/// When receiving spans, the trace id for this request will be guessed based on the order of
+/// incoming spans. So it holds true when at least one span related to invocation N is received
+/// by the extension before the spans of request N+1
+#[derive(Clone, Debug)]
+pub struct ReparentingInfo {
+    pub request_id: String,
```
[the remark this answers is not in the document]
alexgallotta: The string slice means that the ownership belongs somewhere else, right? And the compiler will complain with lifetimes then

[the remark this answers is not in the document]
astuyve: Nvm you always heap allocate it because you stick it into a vecdeque so I think it's always gonna be on the heap

bottlecap/src/lwa/proxy.rs
```diff
@@ -0,0 +1,506 @@
...
+    let https = HttpConnector::new();
+    let client = Client::builder(hyper_util::rt::TokioExecutor::new())
+        .build::<_, http_body_util::Full<prost::bytes::Bytes>>(https);
+
+    // response after forwarding to AWS runtime API
+    let response_to_intercepted = match client.request(forward_intercepted).await {
+        Ok(response) => response,
+        Err(e) => {
+            error!("LWA: Error forwarding request to AWS runtime: {}", e);
+            return Ok(Response::new(BoxBody::new(Full::new(Bytes::new()))));
```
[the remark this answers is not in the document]
alexgallotta: nope that was redundant, removed it

bottlecap/src/lwa/proxy.rs
```diff
@@ -0,0 +1,506 @@
...
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
+mod tests {
+    use crate::config::{AwsConfig, Config};
+    use crate::lifecycle::invocation::processor::Processor;
+    use crate::lwa::proxy::start_lwa_proxy;
+    use crate::tags::provider::Provider;
+    use http_body_util::BodyExt;
+
```
astuyve, line 386:
```suggestion
use tracing::error;
```

PR #639, opened 2025-04-07
fix: Don't assume we're always getting a Full request payload, collec…

astuyve:
…t it anyway

Fixes an issue where hyper's `incoming` type fails here because I didn't fully understand the implementation details of the hyper_migration shim. Easy fix

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -902,11 +902,12 @@ mod tests {
 
         let context = p.context_buffer.get(&request_id).unwrap();
 
-        assert!(
-            !context
+        assert_eq!(
+            context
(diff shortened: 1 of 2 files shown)


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -902,12 +902,11 @@ mod tests {
 
         let context = p.context_buffer.get(&request_id).unwrap();
 
-        assert_eq!(
-            context
+        assert!(
```
astuyve, line 905: clippy got mad about this?

duncanista: maybe `assert_ne` is better than `!false` on `assert`?

duncanista (approved):

PR #644, opened 2025-04-10
feat: [SVLS-6242] bottlecap fips builds

apiarian-datadog:
Building bottlecap with fips mode.

This is entirely focused on removing `ring` (and other non-FIPS-compliant dependencies from our `fips`-featured builds.)


apiarian-datadog:
This does not yet work for alpine, unfortunately. We're seeing errors similar to https://github.com/aws/aws-lc-rs/issues/762 so maybe this is not just us.

bottlecap/build.rs
```diff
@@ -0,0 +1,74 @@
...
+                "--features=fips",
+                "--no-default-features",
+            ])
+            .output()
+            .expect("Failed to execute detailed cargo tree command");
+
+        let output_str = String::from_utf8_lossy(&output.stdout);
+
+        // Check if ring is in the dependency tree
+        if output_str.contains("ring v") {
```
apiarian-datadog, line 39: todo: we'll want to check a few other dependencies that we don't have now but don't want to sneak in in the future.

images/Dockerfile.bottlecap.compile
```diff
@@ -1,11 +1,12 @@
...
 ARG PLATFORM
 
-# FIPS is not used for now.
 ARG FIPS
 
 # Install dependencies
-RUN yum install -y curl gcc gcc-c++ make unzip
+RUN yum install -y curl clang make unzip cmake3 perl go
+# We need to use clang to make sure the FIPS build goes well.
+ENV CC=clang
```
apiarian-datadog, line 9: let's actually see if we can do this with the `AWS_LC_FIPS_CC` setting instead.

apiarian-datadog: done. looks like we only need to do the aws lc fips build with clang

bottlecap/Cargo.toml
```diff
@@ -65,3 +68,14 @@ opt-level = "z"  # Optimize for size.
 lto = true
 codegen-units = 1
 strip = true
+
+[features]
+default = ["reqwest/rustls-tls", "dogstatsd/default"]
+fips = [
```
apiarian-datadog, line 74: unfortunately, as written, the default feature is incompatible with the fips feature. so in order to build the fips feature correctly we have to disable the default feature. this is fine for normal builds where we don't care about fips, but it means that when building or checking the fips feature we have to remember to say `--no-default-features` in addition to `--features=fips`. our build scripts take care of this, but maybe there's a better way to do this?

bottlecap/build.rs
apiarian-datadog: done.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -74,6 +74,42 @@ use tokio_util::sync::CancellationToken;
...
+fn log_fips_status() {
+    debug!("FIPS mode is disabled");
+}
+
+/// Sets up the client provider for TLS operations.
+/// In FIPS mode, this installs the AWS-LC crypto provider.
+/// In non-FIPS mode, this is a no-op.
+#[cfg(feature = "fips")]
+fn prepare_client_provider() -> Result<()> {
+    rustls::crypto::default_fips_provider()
```
apiarian-datadog, line 95: unfortunately the `rustls::client::ClientBuilder` is buried very deep inside the `reqwest::Client` building process so we don't have an easy way to call `.fips()` on it at runtime to verify that the client we're building is FIPS compliant to have validation in depth.

.github/workflows/rs_ci.yml
```diff
@@ -76,8 +76,11 @@ jobs:
           components: clippy
           cache: false
       - uses: mozilla-actions/sccache-action@v0.0.9
+      # We need to do these separately because the fips feature is incompatible with the default feature.
       - working-directory: bottlecap
-        run: cargo clippy --workspace --all-features
+        run: cargo clippy --workspace --features default
+      - working-directory: bottlecap
+        run: cargo clippy --workspace --no-default-features --features fips
```
duncanista, line 83: Should we have a specific fips step?

.gitlab/templates/pipeline.yaml.tpl
```diff
@@ -22,23 +22,27 @@ cargo fmt:
...
 cargo clippy:
   stage: test
   tags: ["arch:amd64"]
   image: ${CI_DOCKER_TARGET_IMAGE}:${CI_DOCKER_TARGET_VERSION}
   needs: []
   script:
-    - cd bottlecap && cargo clippy --all-features
+    - apt-get update && apt-get install -y --fix-missing --no-install-recommends golang-go
+    - cd bottlecap
+    # We need to do these separately because the fips feature is incompatible with the default feature.
```
duncanista, line 43: Again, should it have its own step?

apiarian-datadog: what would it do? like a `cargo clippy` and then another `cargo clippy fips`? is the overhead worth it considering that they're both actually *clippy* steps and will share almost all of their logic?

bottlecap/src/bin/bottlecap/main.rs
apiarian-datadog: we'll see if this can be added: https://github.com/seanmonstar/reqwest/issues/2653

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -78,6 +78,42 @@ use tokio_util::sync::CancellationToken;
...
+compile_error!("When building in fips mode, the default feature must be disabled");
+
+#[cfg(feature = "fips")]
+fn log_fips_status() {
+    debug!("FIPS mode is enabled");
+}
+
+#[cfg(not(feature = "fips"))]
+fn log_fips_status() {
+    debug!("FIPS mode is disabled");
```
duncanista, line 91: Maybe we don't debug it when it's not enabled

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -78,6 +78,42 @@ use tokio_util::sync::CancellationToken;
...
+        })
+}
+
+#[cfg(not(feature = "fips"))]
+// this is not unnecessary since the fips version can return an error
+#[allow(clippy::unnecessary_wraps)]
+fn prepare_client_provider() -> Result<()> {
+    // No-op in non-FIPS mode
+    Ok(())
+}
```
duncanista, line 115: Could we move all of these things to a `fips` module?

apiarian-datadog: the fips ones or all the methods that have a fips and non-fips version?

apiarian-datadog:
using [`a183aea` (#664)](https://github.com/DataDog/datadog-lambda-extension/pull/664/commits/a183aea2316bba2c079200c6986509c1a4b0dc1f) i've confirmed that the clients we're creating are in fact fips-ish at runtime. if the associated reqwest pr gets merged, we'll update the the fips code here to check things always, but in the meantime we're ready to go. @duncanista could you take another look?

bottlecap/src/bin/bottlecap/main.rs
apiarian-datadog: we ended up building a reqwest::Client builder adapter in https://github.com/DataDog/serverless-components/pull/13/files .

bottlecap/src/bin/bottlecap/main.rs
apiarian-datadog: done!

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -189,14 +191,25 @@ async fn main() -> Result<()> {
     let (mut aws_config, config) = load_configs(start_time);
 
     enable_logging_subsystem(&config);
+    log_fips_status();
     let version_without_next = EXTENSION_VERSION.split('-').next().unwrap_or("NA");
     debug!("Starting Datadog Extension {version_without_next}");
-    let client = Client::builder().no_proxy().build().map_err(|e| {
```
apiarian-datadog, line 194: i wonder if there's a way to add a check so that in the future anyone trying to run `reqwest::Client::builder()` directly will get a warning to use the `create_reqwest_client_builder` instead.

apiarian-datadog: there is, with clippy.

bottlecap/src/fips/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+        .map_err(|e| {
+            Error::new(
+                std::io::ErrorKind::InvalidData,
+                format!("Failed to set up fips provider: {e:?}"),
+            )
+        })
+}
+
+#[cfg(not(feature = "fips"))]
+// this is not unnecessary since the fips version can return an error
```
duncanista, line 39:
```suggestion
// this is necessary since the fips version can return an error
```

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -189,14 +191,25 @@ async fn main() -> Result<()> {
     let (mut aws_config, config) = load_configs(start_time);
 
     enable_logging_subsystem(&config);
+    log_fips_status();
```
duncanista, line 194: I'm still against the idea of logging when something is not enabled in production, but I'm still going to approve

duncanista (approved):
LGTM – left a couple nits, but overall amazing job with this

bottlecap/src/bin/bottlecap/main.rs
apiarian-datadog: (we discussed life and decided that since layer is going to be shipped to commercial too and since it only logs at startup, we are okay keeping the log line in)

bottlecap/src/fips/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+compile_error!("When building in fips mode, the default feature must be disabled");
+
+#[must_use]
+pub fn runtime_layer_would_enable_fips_mode(region: &str) -> bool {
+    let is_gov_region = region.starts_with("us-gov-");
+
+    env::var("DD_LAMBDA_FIPS_MODE")
+        .map(|val| val.to_lowercase() == "true")
+        .unwrap_or(is_gov_region)
+}
```
duncanista, line 21: Not sure I understand what does this do

bottlecap/src/fips/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+#[cfg(not(feature = "fips"))]
+pub fn check_fips_mode_mismatch(region: &str) {
+    let runtime_would_enable = runtime_layer_would_enable_fips_mode(region);
+    if runtime_would_enable {
+        debug!("FIPS mode is disabled in this Extension layer but would be enabled in the runtime layer based on region and environment settings. Deploy the FIPS version of the Extension layer or set DD_LAMBDA_FIPS_MODE=false to ensure consistent FIPS behavior.");
+    }
+}
```
duncanista, line 37: Looks very very similar besides the `!`, maybe we can improve this?

Someone could confuse and remove this because they look veeery similar

apiarian-datadog: hmm, how about something like

``` rust
pub fn check_fips_mode_mismatch(region: &str) {
	let runtime_would_enable = ...
	if runtime_would_enable {
		#cfg[not(feature = "fips")]
		debug!(...)
	} else {
		#cfg[feature = "fips"]
		debug!(...)
	}
}
```

bottlecap/src/fips/mod.rs
apiarian-datadog: this is meant to replicate the logic from https://github.com/DataDog/datadog-lambda-python/pull/588:

look for `DD_LAMBDA_FIPS_MODE` and see if it is "true", defaulting to true in a gov region and false in commercial regions.

duncanista: Ah, I understand, name was a little confusing

I'd have done

```
let is_gov_region = ...
let lambda_layer_is_fips_enabled = ...

lambda_layer_is_fips_enabled || is_gov_region
```

Or whichever naming convention makes more sense to you

apiarian-datadog: ah, i think there's a subtle difference between those two approaches. if we implement it like this

```rust
pub fn runtime_layer_would_enable_fips_mode(region: &str) -> bool {
    let is_gov_region = region.starts_with("us-gov-");
    let lambda_layer_is_fips_enabled = env::var("DD_LAMBDA_FIPS_MODE")
        .map(|val| val.to_lowercase() == "true")
        .unwrap_or(false);

    lambda_layer_is_fips_enabled || is_gov_region
}
```

| region | DD_LAMBDA_FIPS_MODE_ENABLED | current result | simple boolean result |
| --- | --- | --- | --- |
| commercial | unset | false | false |
| commercial | "true" | true | true |
| commercial | "false" | false | false |
| gov | unset | true | true |
| gov | "true" | true | true |
| gov | "false "| false | true |

note that this basically means we cannot disable the feature in govcloud unless we actually read the lambda_layer_is_fips_enabled as a optional boolean. but that makes the logic more complex since we end up having to do a match of some kind instead of letting the env `unwrap_or` take care of the default value by reading it from the `is_gov_region`.

is there a better way to read the `DD_LAMBDA_FIPS_MODE` env that doesn't result in a more complicated match statement down the line?

apiarian-datadog: ok, i've kept the original logic but added a comment to explain what's going on.

PR #645, opened 2025-04-11
feat: Register first in main

astuyve:
We don't need any configuration options in order to register the extension, so make that request first.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -186,22 +186,19 @@ fn build_function_arn(account_id: &str, region: &str, function_name: &str) -> St
 #[tokio::main]
 async fn main() -> Result<()> {
     let start_time = Instant::now();
-    let (mut aws_config, config) = load_configs(start_time);
-
-    enable_logging_subsystem(&config);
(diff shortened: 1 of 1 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -186,22 +186,19 @@ fn build_function_arn(account_id: &str, region: &str, function_name: &str) -> St
...
-    let version_without_next = EXTENSION_VERSION.split('-').next().unwrap_or("NA");
-    debug!("Starting Datadog Extension {version_without_next}");
     let client = Client::builder().no_proxy().build().map_err(|e| {
         Error::new(
             std::io::ErrorKind::InvalidData,
             format!("Failed to create client: {e:?}"),
         )
     })?;
-
     let r = register(&client)
```
duncanista, line 195: Should we modify the logs in `register` to be `println!` in case they're from the `tracing` library which might not be enabled by now?

astuyve: I don't think we debug at all? We're only calling `panic!()` if we receive a non-200 status code from the register API call, which will crash the process and print to stdout

duncanista (approved):

PR #655, opened 2025-04-18
fix: Extract trace context from inferred span when AWS Xray is enabled

nhulston:
### What
- In `span_inferrer.rs`, we were looking for `generated_span_context` first before looking for the inferred span's trace context.
- AWS Xray's trace context is extracted from the AWSTraceHeader in `extract_trace_context_from_aws_trace_header`. 
- Therefore, if the user has AWS Xray enabled, we would use that trace context rather than the inferred span's trace context, breaking trace propagation.
- The fix is simply swapping the order of the if statements: first look for inferred span's trace context, then look for trace context in the AWS trace header.

### Testing
Before this change, trace propagation was broken when Xray was enabled:
Screenshot 2025-04-18 at 4 28 28 PM

After this change, trace propagation works as expected with Xray enabled:
Screenshot 2025-04-18 at 4 29 30 PM

The Datadog Java tracer injects trace context in the AWSTraceHeader for SQS, so I also manually verifed that this change does not break Java->SQS->Java trace propagation:
Screenshot 2025-04-18 at 4 31 53 PM

Lastly, the `generated_span_context` is also used by Step Functions, but there is no inferred span for step functions so this change should not affect Step Functions.
https://github.com/DataDog/datadog-lambda-extension/blob/56e6cc11b74eace24a8621d058d9b6d13d708a2b/bottlecap/src/lifecycle/invocation/span_inferrer.rs#L231-L236

### DIFF
--- a/bottlecap/src/lifecycle/invocation/span_inferrer.rs
+++ b/bottlecap/src/lifecycle/invocation/span_inferrer.rs
@@ -312,6 +312,9 @@ impl SpanInferrer {
     /// otherwise it will return `None`.
     ///
     pub fn get_span_context(&self, propagator: &impl Propagator) -> Option<SpanContext> {
+        // Order matters here: check inferred span for trace context first, then fallback to generated span context.
+        // If the order is flipped, trace propagation will be broken when AWS Xray is enabled.
+        // https://github.com/DataDog/datadog-lambda-extension/pull/655
         if let Some(sc) = self.carrier.as_ref().and_then(|c| propagator.extract(c)) {
             debug!("Extracted trace context from inferred span");
             return Some(sc);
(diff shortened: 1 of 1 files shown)


bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -312,16 +312,17 @@ impl SpanInferrer {
     /// otherwise it will return `None`.
     ///
     pub fn get_span_context(&self, propagator: &impl Propagator) -> Option<SpanContext> {
-        // Step Functions `SpanContext` is deterministically generated
-        if self.generated_span_context.is_some() {
-            return self.generated_span_context.clone();
-        }
-
         if let Some(sc) = self.carrier.as_ref().and_then(|c| propagator.extract(c)) {
```
astuyve, line 315: Can we add a comment here indicating that the order matters? Although the test will fail if the order is now reversed, someone may need the context for this change going forward. I'd suggest a 1-sentence summary and then a link to this PR.

astuyve (approved):

bottlecap/src/lifecycle/invocation/span_inferrer.rs
nhulston: Yes, good suggestion!

PR #658, opened 2025-04-22
Cold start spans for Node+Python

nhulston:
work in progress


bottlecap/src/traces/trace_agent.rs
```diff
@@ -393,6 +393,21 @@ impl TraceAgent {
...
+                            &tags_provider,
+                            &trace_processor,
+                            &trace_tx,
+                            span.trace_id,
+                        )
+                        .await;
+                    if let Some(cold_start_span_id) = cold_start_span_id {
+                        span.parent_id = cold_start_span_id;
+                    }
+                }
```
duncanista, line 409: Tags provider, trace processor, and trace tx should exist already in the invocation processor, why would you send them?

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -425,6 +424,69 @@ impl Processor {
...
+            traces,
+            body_size,
+            tags_provider,
+            trace_processor,
+            trace_agent_tx,
+        )
+        .await;
+    }
+
+    pub async fn update_and_send_cold_start_span(
```
duncanista, line 437: I'd separate this into two methods, also can you leverage the `request_id` from the span `aws.lambda.load`? AFAIK they should include the `request_id`, maybe?

bottlecap/src/traces/trace_agent.rs
nhulston: I don't think so, looks like they're created in `main.rs` and passed in to `trace_agent.rs`

duncanista: You're right, I was mislead because they appear in `on_platform_runtime_done`

bottlecap/src/lifecycle/invocation/processor.rs
nhulston: For visibility: we investigated and there's no `request_id` in the `aws.lambda.load` span. But as discussed, looping through the contexts in the buffer should be sufficient, and there's only up to 5 contexts in the buffer, so the performance should be good too.

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -397,6 +397,19 @@ impl ContextBuffer {
...
+        if let Some(context) = self
+            .buffer
+            .iter_mut()
+            .find(|context| context.cold_start_span.is_some())
+        {
+            return context.cold_start_span.as_mut();
+        }
+
+        None
+    }
```
duncanista, line 411: Not a fan of returning a field from a context, would it make sense to return instead the context and rename this to something like `get_context_with_cold_start` or something similar?

In reality, no other method here returns a field, but instead a context – it only modifies fields given a specific context key

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -324,9 +324,8 @@ impl Processor {
         trace_processor: Arc<dyn trace_processor::TraceProcessor + Send + Sync>,
         trace_agent_tx: Sender<SendData>,
     ) {
-        let context = self.enrich_ctx_at_platform_done(request_id, status);
-
         if self.tracer_detected {
+            let context = self.enrich_ctx_at_platform_done(request_id, status);
```
duncanista, line 328: Why is this change needed?

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -425,6 +424,60 @@ impl Processor {
...
+            traces,
+            body_size,
+            tags_provider,
+            trace_processor,
+            trace_agent_tx,
+        )
+        .await;
+    }
+
+    // For Node/Python: Updates the cold start span with the given trace ID and sends to the trace agent.
```
duncanista, line 437: Improve docs here too

bottlecap/src/traces/trace_agent.rs
```diff
@@ -393,6 +393,24 @@ impl TraceAgent {
...
+                        .send_cold_start_span(
+                            &tags_provider,
+                            &trace_processor,
+                            &trace_tx,
+                            span.trace_id,
+                        )
+                        .await;
+                    if let Some(cold_start_span_id) = cold_start_span_id {
+                        span.parent_id = cold_start_span_id;
+                    }
```
duncanista, line 411: Is there a way to not trigger a send here and instead just get the `span_id` while modifying the cold start on the other end?

I think we could separate this, as opposed to do all things at once, wdyt?

duncanista (commented):
It's looking good!

Left some comments

bottlecap/src/lifecycle/invocation/processor.rs
nhulston: Good catch. Not needed -- i thought this was just a small improvement.
But taking a second look, this `enrich_ctx` call has side effects. So this change actually looks like it will break stuff. Reverting now.

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -394,8 +397,10 @@ impl Processor {
 
         // Handle cold start span if present
         if let Some(cold_start_span) = &mut context.cold_start_span {
-            cold_start_span.trace_id = context.invocation_span.trace_id;
-            cold_start_span.parent_id = context.invocation_span.parent_id;
+            if cold_start_span.trace_id == 0 {
+                cold_start_span.trace_id = context.invocation_span.trace_id;
+                cold_start_span.parent_id = context.invocation_span.parent_id;
+            }
```
nhulston, line 403: prevent the trace ID from being overwritten to 0, since the invocation_span's trace_id + parent_id are 0 in python/node

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -397,6 +397,19 @@ impl ContextBuffer {
...
+    pub fn get_context_with_cold_start(&mut self) -> Option<&mut Context> {
+        if let Some(context) = self
+            .buffer
+            .iter_mut()
+            .find(|context| context.cold_start_span.is_some())
+        {
+            return Some(context);
+        }
+
+        None
```
duncanista, line 410: I think this:

```
self.buffer
    .iter_mut()
    .find(|context| context.cold_start_span.is_some())
```

does the same

bottlecap/src/traces/trace_agent.rs
```diff
@@ -393,6 +393,18 @@ impl TraceAgent {
 
         for chunk in &mut traces {
             for span in chunk.iter_mut() {
+                // If the aws.lambda.load span is found, we're in Python or Node.
+                // We need to update the trace ID of the cold start span, reparent the `aws.lambda.load`
+                // span to the cold start span, and eventually send the cold start span.
+                if span.name == "aws.lambda.load" {
```
duncanista, line 399: constant here?

bottlecap/src/traces/trace_agent.rs
```diff
@@ -393,6 +393,18 @@ impl TraceAgent {
...
+                    let mut invocation_processor = invocation_processor.lock().await;
+                    let cold_start_span_id =
+                        invocation_processor.set_cold_start_span_trace_id(span.trace_id);
+                    if let Some(cold_start_span_id) = cold_start_span_id {
+                        span.parent_id = cold_start_span_id;
+                    }
+                }
```
duncanista, line 406:
```suggestion
                    if let Some(span_id) = invocation_processor.set_cold_start_span_trace_id(span.trace_id) {
                        span.parent_id = span_id;
                    }
                }
```

duncanista (approved):
Left some nits, LGTM!

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -425,6 +430,72 @@ impl Processor {
...
+                    debug!("Not sending cold start span because trace ID is unset.");
+                    return;
+                }
+
+                let traces = vec![cold_start_span.clone()];
+                let body_size = size_of_val(cold_start_span);
+
+                self.send_spans(
+                    traces,
+                    body_size,
```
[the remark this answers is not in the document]
nhulston: Here, the .clone() needs to be done in a separate variable; otherwise we get two mutable borrows in the same period

PR #661, opened 2025-04-24
aj/fix route tag

### LINKED ISSUE #601: Rest APIGW {proxy+}
Hi,

We have recently upgraded our serverless plugin from `v5.49.0` -> `v5.78.0`, and with that the extension was upgraded from `49` -> `69` which uses the new gen extension by default.

We use the pattern of defining the path with a greedy path variable eg. `/api/{proxy+}` and handling routing internally within the lambda. With the previous extension, it captured the resource name similar to the actual path group eg. `/api/foo/bar/{id}`. Now it's capturing the actual path group `/api/{proxy+}`.

I'm not sure if the old behaviour can be supported or toggled, but it's quite important for us as we have alerts and tracing based off the actual resource names.

astuyve:
This is a breaking change, and a feature improvement/enhancement.

It and adds a parameterized path parser for APIGW rest/http events. Specifically this is most helpful for proxy routes.

Initial versions of the Go extension would just set the `resource` to the raw route, inclusive of IDs. This caused cardinality issues and `resource` limit exhaustion for certain customers as serverless would eat all of the cardinality available.

With the introduction of the next-generation lambda extension, we changed this to match the implementation of node/python, which tags all resources for `proxy` routes with `{proxy+}`. That's less helpful for APM customers doing internal routing with frameworks (but solved the cardinality issues).

For universal-instrumentation languages, we're unable to rely on the internal framework route matching so this change looks for:
- integer IDs between REST routes
- UUIDs
- GUIDs
- ULIDs

then replaces the content of the identifier with the previous string value in the REST URL, so:
```
/dev/proxy_route/users/1234/friends/abdx-3230234-asdfalksdj-2342/pizzas/7
->
/dev/proxy_route/users/{user_id}/friends/{friend_id}/pizzas/{pizza_id}
```

This does diverge from node and python (which we will hopefully extend this functionality to), but it does strike a useful balance.

This impacts users enabling APIGW Parameterized Routes as well. Previously we'd use the `resource` argument there, which may use customer placeholders. Now we override that.

I'm ammenable to changing that (we'd have to check if the `resource` contains `proxy` and then trigger this behavior instead. But, after spot checking a few internal users I think this change would result in basically the same thing. It's still breaking for users with monitors configured, but that can't be avoided

Fixes #601


bottlecap/src/lifecycle/invocation/triggers/mod.rs
```diff
@@ -31,6 +50,45 @@ pub trait ServiceNameResolver {
     fn get_generic_identifier(&self) -> &'static str;
 }
 
+#[must_use]
+pub fn parameterize_api_resource(resource: String) -> String {
```
alexgallotta, line 54: Can this logic be a regexp instead?

astuyve: It could, IMHO that was more confusing to grok than this

astuyve: I'll get chatgpt/claude code to generate the full regex and see if there's a performance difference. The regex will probably be faster

bottlecap/src/lifecycle/invocation/triggers/api_gateway_http_event.rs
```diff
@@ -1,3 +1,4 @@
+use super::parameterize_api_resource;
```
duncanista, line 1: nit `super`

bottlecap/src/lifecycle/invocation/triggers/mod.rs
astuyve: image
claude code used criterion and noted the regex was actually "much slower" so let's see what it comes up with

astuyve: update: Claude said it ran critereon and tested this code but it doesn't actually run or compile, so....

astuyve: yeah I think this is still faster/easier. Gonna resolve for now

bottlecap/src/lifecycle/invocation/triggers/mod.rs
```diff
@@ -31,6 +50,45 @@ pub trait ServiceNameResolver {
...
+            continue;
+        }
+
+        // Check if this part looks like an identifier
+        // Number, ULID, GUID, or UUID
+        if part.chars().all(|c| c.is_ascii_digit()) || ULID_UUID_GUID.is_match(part) {
+            // Determine the parameter name based on the previous segment
+            let param_name = if i > 1 && !parts[i - 1].is_empty() {
+                let singular = parts[i - 1].trim_end_matches('s');
+                format!("{singular}_id")
```
duncanista, line 78: Do we want to name id `_id`? 

Are we sure we are going to catch everything? If it's not a UUID, GUID, ULID, it won't catch any other stuff, this won't effectively reduce cardinality, right? I guess it will reduce it substantially tho

duncanista (approved):
LGTM!

bottlecap/src/lifecycle/invocation/triggers/mod.rs
astuyve: Let me write a test/catch `/id/12434`, but yeah – this won't fix everything. At some point we'll want to expose a configurable env var for this

Issue #652, opened 2025-04-17
Possible bug in InvocationTimes?


mheffner:
Hey folks, while taking a look at the code for the adaptive flushing I noticed what may be a possible bug. In `InvocationTimes` it appears that the `times` array is used as a circular buffer, with `head` indicating the next sample bucket:

https://github.com/DataDog/datadog-lambda-extension/blob/1662d8b93f613fc5a088819a905fee526c515a89/bottlecap/src/lifecycle/invocation_times.rs#L18-L31

What looks off is that `should_adapt_to_periodic` doesn't appear to iterate the `times` array in the same order. Instead of starting at head and walking backwards it iterates in array order.

nhulston:
Thanks for opening this! Yes, this definitely looks like a bug

nhulston:
Thanks, this has been fixed in https://github.com/DataDog/datadog-lambda-extension/pull/662 and will be out in the next release!

PR #662, opened 2025-04-24
fix: Correctly get oldest invocation in `should_adapt_to_periodic` 

nhulston:
### What
We switch to periodic flushing if there are at least 20 invocations and the average time between each invocation is less than 120 seconds. (see [datadog-agent implementation](https://github.com/DataDog/datadog-agent/blob/main/pkg/serverless/daemon/interval.go#L47-L76))

In `should_adapt_to_periodic`, we calculate the average time between invocations (for the last 20 invocations) by taking the time between the oldest and newest invocations in the buffer, then dividing by 20.

Prior to this, we were assuming the oldest invocation in the buffer was the last item in the array. This is incorrect because the buffer is a circular buffer; we can instead use `buffer[head]` to get the oldest item in the array.

I also added a small improvement where we exit early if the buffer is not full (meaning <20 invocations have occurred so far).

### Motivation
https://github.com/DataDog/datadog-lambda-extension/issues/652
https://datadoghq.atlassian.net/browse/SVLS-6684

### DIFF
--- a/bottlecap/src/lifecycle/invocation_times.rs
+++ b/bottlecap/src/lifecycle/invocation_times.rs
@@ -86,4 +86,38 @@ mod tests {
         assert_eq!(invocation_times.head, 1);
         assert!(!invocation_times.should_adapt_to_periodic(10000));
     }
+
+    #[test]
+    fn should_adapt_to_periodic_when_fast_invokes() {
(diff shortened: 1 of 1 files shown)


astuyve (approved):
Thanks for the fix!

PR #663, opened 2025-04-24
feat: Optionally disable proc enhanced metrics

### LINKED ISSUE #648: v76 seems affects lambda runtime
We just upgraded from v68  to v76. The only change to lambda was the diff in the version like below.
- COPY --from=public.ecr.aws/datadog/lambda-extension:68 /opt/extensions/ /opt/extensions
+ COPY --from=public.ecr.aws/datadog/lambda-extension:76 /opt/extensions/ /opt/extensions

Right after upgrade we see increase in `aws.lambda.enhanced.runtime_duration`, the scale below is in ms. The increase is not critical, but just wondering is that expected or can be configured.

Image

astuyve:
Fixes #648

For customers using very very fast/small lambda functions (usually just rust), there can be a small 1-2ms increase in runtime duration when collecing metrics like open file descriptors or tmp file usage.

We still enable these by default, but customers can now optionally disable them

### DIFF
--- a/bottlecap/src/config/env.rs
+++ b/bottlecap/src/config/env.rs
@@ -38,6 +38,7 @@ pub struct Config {
     pub logs_config_logs_dd_url: String,
     pub serverless_flush_strategy: FlushStrategy,
     pub enhanced_metrics: bool,
+    pub proc_enhanced_metrics: bool,
     /// Flush timeout in seconds
     pub flush_timeout: u64, //TODO go agent adds jitter too
     pub https_proxy: Option<String>,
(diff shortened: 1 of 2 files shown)


bottlecap/src/config/env.rs
```diff
@@ -38,6 +38,7 @@ pub struct Config {
     pub logs_config_logs_dd_url: String,
     pub serverless_flush_strategy: FlushStrategy,
     pub enhanced_metrics: bool,
+    pub proc_enhanced_metrics: bool,
```
duncanista, line 41: should we include anything `lambda|serverless` related?

astuyve: Hmmm, why? Today `enhanced metrics` as a concept only apply to lambda/serverless

duncanista: idk, just trying to be cautious about adding new env vars

astuyve: okay I can add lambda

duncanista: just a nit, i trust your gut 🫡

duncanista (approved):

PR #664, opened 2025-04-25
hack: [SVLS-6242] trying the proposed reqwest fips check

apiarian-datadog:


apiarian-datadog:
closing this in favor of using the existing `use_preconfigured_tls` builder method.

PR #666, opened 2025-04-28
feat(bottlecap): add MSK inferred spans

nhulston:
Add inferred spans for MSK events. Since there are multiple records, possibly from multiple different topics, we just get the first item from the first topic and use that information to create the inferred span.

Screenshot 2025-04-28 at 10 09 23 AM

https://datadoghq.atlassian.net/browse/SVLS-6711

### DIFF
--- a/bottlecap/src/lifecycle/invocation/triggers/msk_event.rs
+++ b/bottlecap/src/lifecycle/invocation/triggers/msk_event.rs
@@ -122,3 +122,140 @@ impl ServiceNameResolver for MSKEvent {
         "lambda_msk"
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::lifecycle::invocation::triggers::test_utils::read_json_file;
+
+    #[test]
+    fn test_new() {
+        let json = read_json_file("msk_event.json");
+        let payload = serde_json::from_str(&json).expect("Failed to deserialize into Value");
+        let result = MSKEvent::new(payload).expect("Failed to deserialize into MSKEvent");
+
+        let record = MSKRecord {
+            topic: String::from("topic1"),
+            partition: 0,
+            timestamp: 1745846213022.0,
+        };
+        let mut expected_records = HashMap::new();
+        expected_records.insert(String::from("topic1"), vec![record]);
+
+        let expected = MSKEvent {
+            event_source: String::from("aws:kafka"),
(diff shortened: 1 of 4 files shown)


bottlecap/src/lifecycle/invocation/triggers/msk_event.rs
```diff
@@ -0,0 +1,261 @@
...
+                    key_to_keep = Some(key_copy);
+                }
+            }
+
+            if let Some(ref key) = key_to_keep {
+                records_map.retain(|k, _| k == key);
+            } else {
+                records_map.clear();
+            }
+        }
```
duncanista, line 45: This can be improved, too many if let Some

bottlecap/src/lifecycle/invocation/triggers/msk_event.rs
```diff
@@ -0,0 +1,261 @@
...
+                    self.event_source_arn.clone(),
+                ),
+            ]);
+        }
+    }
+
+    fn get_tags(&self) -> HashMap<String, String> {
+        HashMap::from([(
+            FUNCTION_TRIGGER_EVENT_SOURCE_TAG.to_string(),
+            "msk".to_string(),
```
duncanista, line 95: I'd check with the Experiences team which value they'd like to use here, as this helps us showcase it in the UI

bottlecap/src/lifecycle/invocation/triggers/msk_event.rs
```diff
@@ -0,0 +1,261 @@
...
+    fn get_specific_identifier(&self) -> String {
+        self.event_source_arn
+            .split('/')
+            .nth(1)
+            .unwrap_or_default()
+            .to_string()
+    }
+
+    fn get_generic_identifier(&self) -> &'static str {
+        "lambda_msk"
```
duncanista, line 122: You'd have to add it here too: https://docs.datadoghq.com/serverless/aws_lambda/configuration?tab=datadogcli#rename-all-services-of-a-type

nhulston: Thanks, opened a PR here: https://github.com/DataDog/documentation/pull/29008

bottlecap/src/lifecycle/invocation/triggers/msk_event.rs
nhulston: I talked with them. `"msk"` is good here, but we'll need to add it to the enum here and get icons created for MSK
https://github.com/DataDog/web-ui/blob/preprod/packages/apps/serverless/lib/definitions/constants/aws/lambda/function-trigger.constants.ts

PR #668, opened 2025-05-01
Support trace extraction from multiValueHeaders in ALB

nhulston:
### What

Support trace extraction when we receive `multiValueHeaders` instead of `headers`. This can happen if multi value headers are turned on in ALB.

We do this by creating an ALB event in `span_inferrer` -- even though there's no inferred spans, this is still a common event type.

The event payload structure for multiValueHeaders is defined here: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html

This also has the benefit of creating these trigger tags:
Screenshot 2025-05-01 at 4 55 34 PM
Screenshot 2025-05-01 at 4 56 38 PM

### Motivation
Porting https://github.com/DataDog/datadog-lambda-js/pull/647 and https://github.com/DataDog/datadog-agent/pull/31542
https://datadoghq.atlassian.net/browse/SVLS-6757


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -734,6 +735,15 @@ impl Processor {
...
 
+        if let Some(multi_value_headers) = payload_value.get("multiValueHeaders") {
+            if let Some(flattened_headers) = flatten_multi_value_headers(multi_value_headers) {
+                if let Some(sc) = self.propagator.extract(&flattened_headers) {
+                    debug!("Extracted trace context from event multiValueHeaders");
+                    return Some(sc);
+                }
+            }
+        }
+
```
duncanista, line 746: Not happy about the idea that the span inferrer will know its API Gateway, yet it depends on this to first fail then go through this function to get the span context

hghotra (approved):
✅

bottlecap/src/lifecycle/invocation/processor.rs
nhulston: If `multiValueHeaders` exist, then `headers` won't exist, so all we have to wait for this if statement to fail:
```
if let Some(payload_headers) = payload_value.get("headers") {
```

We definitely could extract trace context in span inferrer, but then we'd have to implement the same logic in 3 places for API GW. (Also there's no inferred spans for ALB either)

So i think this is the best place to do it, wdyt?

duncanista: TLDR on our discussion earlier;

Okay to have this, let's have it also in API GW, even though logic is repeated, we're separating the concern / not making this function take that decision when there is an inferred span.

bottlecap/src/lifecycle/invocation/triggers/alb_event.rs
```diff
@@ -0,0 +1,93 @@
...
+                flat_map.insert(key.to_lowercase(), first_val.clone());
+            }
+        }
+        flat_map
+    }
+}
+
+impl ServiceNameResolver for ALBEvent {
+    fn get_specific_identifier(&self) -> String {
+        self.request_context.elb.target_group_arn.clone()
```
nhulston, line 87: is this fine for the specific identifier?

bottlecap/src/lifecycle/invocation/triggers/alb_event.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    fn enrich_span(&self, _span: &mut Span, _service_mapping: &HashMap<String, String>) {}
+
+    fn get_tags(&self) -> HashMap<String, String> {
+        HashMap::from([
+            ("http.method".to_string(), self.http_method.clone()),
+            (
+                FUNCTION_TRIGGER_EVENT_SOURCE_TAG.to_string(),
+                "application-load-balancer".to_string(),
```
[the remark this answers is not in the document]
nhulston: They said this string is ok!

bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -82,8 +83,14 @@ impl SpanInferrer {
...
         let mut is_step_function = false;
+        let mut is_alb_event = false;
 
-        if APIGatewayHttpEvent::is_match(payload_value) {
+        if ALBEvent::is_match(payload_value) {
+            if let Some(t) = ALBEvent::new(payload_value.clone()) {
+                trigger = Some(Box::new(t));
+                is_alb_event = true;
+            }
+        } else if APIGatewayHttpEvent::is_match(payload_value) {
```
duncanista, line 93: Can this go after the API Gateway Events? Not sure how common this one would be

bottlecap/src/lifecycle/invocation/triggers/alb_event.rs
duncanista: This is for Service Mapping, maybe we want to see if we can get the name of the ALB instead of the ARN? To keep consistency with https://docs.datadoghq.com/serverless/aws_lambda/configuration/?tab=datadogcli#rename-specific-services

duncanista (approved):
LGTM!

On the service mapping renaming, I wonder if we even need it because we cannot create the inferred span?

Maybe in the future we would be able? YAGNI is my thought, but better be safe than sorry, it doesn't take us much to get the name of the ALB from the ARN

bottlecap/src/lifecycle/invocation/triggers/alb_event.rs
nhulston: Thanks, changed it to get the actual name of the ALB

PR #676, opened 2025-05-15
fix(config): remove `apm_ignore_resources` check in OTEL

duncanista:
# What?

Removes usage of `DD_APM_IGNORE_RESOURCES` in the OTEL span transform.

# Why?

1. The implementation was incorrect and shouldn't check for resources to ignore in the transformation step.
2. It was not properly used in the `apm_config` for YAML files.

# Notes:

- Follow up PR to implement `APM_IGNORE_RESOURCES` properly in the Trace Agent.

# More

Learn about ignoring resources: https://docs.datadoghq.com/tracing/guide/ignoring_apm_resources/?tab=datadogyaml#ignoring-based-on-resources

`DD_APM_IGNORE_RESOURCES` is specified as:

```
A list of regular expressions can be provided to exclude certain traces based on their resource name.
All entries must be surrounded by double quotes and separated by commas.
```

A correct usage would be:

```env
DD_APM_IGNORE_RESOURCES="(GET|POST) /healthcheck,API::NotesController#index"
```

or in yaml
```yaml
apm_config:
  ignore_resources: ["(GET|POST) /healthcheck","API::NotesController#index"]
```

### DIFF
--- a/bottlecap/src/config/env.rs
+++ b/bottlecap/src/config/env.rs
@@ -72,7 +72,6 @@ pub struct Config {
     #[serde(deserialize_with = "deserialize_bool_from_anything")]
     pub apm_config_obfuscation_http_remove_paths_with_digits: bool,
     pub apm_features: Vec<String>,
-    pub apm_ignore_resources: Vec<String>,
     // Metrics overrides
     pub dd_url: String,
     pub url: String,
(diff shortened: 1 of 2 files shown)


astuyve (approved):
Do we need to parse it at all? Does bottlecap need to drop those spans, or will the tracer do it?

duncanista:
So @astuyve, it seems that this has to be done in the Trace Agent, not really on the OTLP processing sides, as this just sends it back to the trace agent.

Yes, Bottlecap would need to implement the logic for span dropping when this is set in the agent.

Defined in here: https://docs.datadoghq.com/tracing/guide/ignoring_apm_resources/?tab=datadogyaml#trace-agent-configuration-options

There's more than just the resources ones

PR #684, opened 2025-05-27
feat: continuous flushing strategy for high throughput functions

astuyve:
This is a heavy refactor and new feature.
- Introduces FlushDecision and separates it from FlushStrategy
- Cleans up FlushControl logic and methods

It also adds the ability to flush telemetry across multiple serial lambda invocations. This is done using the `continuous` strategy.

This is a huge win for busy functions as seen in our test fleet, where the p99/max drops precipitously, which also causes the average to plummet. This also helps reduce the number of cold starts encountered during scaleup events, which further reduces latency along with costs:
image

Technical implementation:
We spawn the task and collect the flush handles, then in the two periodic strategies we check if there were any errors or unresolved futures in the next flush cycle. If so, we switch to the `periodic` strategy to ensure flushing completes successfully.

We don't adapt to the periodic strategy unless the last 20 invocations occurred within the `config.flush_timeout` value, which has been increased by default. This is a naive implementation. A better one would be to calculate the first derivative of the invocation periodicity. If the rate is increasing, we can adapt to the continuous strategy. If the rate slows, we should fall back to the periodic strategy.
image

The existing implementation is cautious in that we could definitely adapt sooner but don't.

Todo: add a feature flag for continuous flushing?


bottlecap/src/config/env.rs
```diff
@@ -135,7 +135,7 @@ impl Default for Config {
             api_key_secret_arn: String::default(),
             kms_api_key: String::default(),
             serverless_flush_strategy: FlushStrategy::Default,
-            flush_timeout: 5,
+            flush_timeout: 90,
```
duncanista, line 138: Intentional?

astuyve: I'll move it to 10, which is the default for the go agent

bottlecap/src/logs/flusher.rs
```diff
@@ -55,73 +76,116 @@ impl Flusher {
...
+            match resp {
+                Ok(resp) => {
+                    let status = resp.status();
+                    _ = resp.text().await;
+                    if status == 202 {
+                        return Ok(());
+                    }
+                }
+                Err(e) => {
+                    if attempts > 3 {
```
duncanista, line 172: Maybe we want to make this a constant and share it across flushers?

bottlecap/src/lifecycle/invocation_times.rs
```diff
@@ -20,11 +23,11 @@ impl InvocationTimes {
         self.head = (self.head + 1) % LOOKBACK_COUNT;
     }
 
-    pub(crate) fn should_adapt_to_periodic(&self, now: u64) -> bool {
+    pub(crate) fn should_adapt(&self, now: u64, flush_timeout: u64) -> FlushStrategy {
```
duncanista, line 26: Maybe we rename this to something more significant now that we are returning the strategy?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -75,11 +78,106 @@ use std::{
...
+            trace_flush_handles: FuturesOrdered::new(),
+            log_flush_handles: FuturesOrdered::new(),
+            metric_flush_handles: FuturesOrdered::new(),
+        }
+    }
+
+    async fn await_flush_handles(
+        &mut self,
+        logs_flusher: &LogsFlusher,
+        trace_flusher: &ServerlessTraceFlusher,
```
duncanista, line 105: Shouldn't this be an Arc?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -75,11 +78,106 @@ use std::{
...
+                                .await;
+                        });
+                    }
+                }
+                Err(e) => {
+                    error!("redrive metrics error {e:?}");
+                }
+            }
+        }
+
```
duncanista, line 171: Shouldn't we do all of this in parallel?

bottlecap/src/bin/bottlecap/main.rs
astuyve: Nah at this point we've already cloned it

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -166,6 +166,27 @@ impl Lambda {
         }
     }
 
+    pub fn set_shutdown_metric(&self, timestamp: i64) {
+        if !self.config.enhanced_metrics {
+            return;
+        }
+        let metric = metric::Metric::new(
+            "aws.lambda.enhanced.shutdowns".into(),
```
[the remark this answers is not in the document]
astuyve: I used it for debugging, not sure if we want to actually keep it. But it's kinda neat to see when functions spin down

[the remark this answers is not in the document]
astuyve: update, we're gonna keep it

.github/workflows/rs_ci.yml
```diff
@@ -139,7 +139,7 @@ jobs:
           tool: nextest@0.9
       - uses: mozilla-actions/sccache-action@v0.0.9
       - working-directory: bottlecap
-        run: cargo nextest run --workspace
+        run: cargo nextest run --workspace --features integration_test
```
duncanista, line 142: Why was this changed?

bottlecap/Cargo.toml
```diff
@@ -97,3 +97,4 @@ fips = [
     "rustls/fips",
     "rustls-native-certs",
 ]
+integration_test = ["reqwest/rustls-tls", "dogstatsd/default", "datadog-fips/default" ]
```
duncanista, line 100: This was the reason

astuyve: Required because http-mock does not support http2 servers, so we have to wrap that in build tags (see `src/http_client.rs`)

.github/workflows/rs_ci.yml
astuyve: https://github.com/DataDog/datadog-lambda-extension/pull/684#discussion_r2121704198

bottlecap/src/logs/flusher.rs
```diff
@@ -55,73 +64,116 @@ impl Flusher {
...
+            let resp = cloned_req.send().await;
+            let elapsed = time.elapsed();
 
-                return Err(Box::new(e));
+            match resp {
+                Ok(resp) => {
+                    let status = resp.status();
+                    _ = resp.text().await;
+                    if status == 202 {
+                        return Ok(());
```
duncanista, line 156: Wondering if we had the `status != 202` as that because there would be no error but the request was not successful, would it make sense to debug here?

Not sure TBH

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -429,16 +543,46 @@ async fn extension_loop_active(
...
+                    .push_back(tokio::spawn(async move {
+                        cloned_metrics_flusher
+                            .lock()
+                            .await
+                            .flush()
+                            .await
+                            .unwrap_or_default()
+                    }));
+                race_flush_interval.reset();
+            } else if current_flush_decision == FlushDecision::Periodic {
```
duncanista: Would it make sense to have a function in the shutdown handles which internally spawns the task and clones the value?

duncanista (approved):
Looks great, left a couple nits!

bottlecap/src/bin/bottlecap/main.rs
astuyve: It could but the messy part is really just dealing with the lifetimes of the flushers, which would have to happen in that function anyway. Unless you think there's additional value to be gained, I'm inclined to leave it as is

bottlecap/src/logs/flusher.rs
astuyve: The common case here would be someone temporarily disabling an API key and then re-enabling it (eg: brown out test), so we'd retry in a 403 event. Any _successful_ request (eg: any status code at all), will hit this block. So I'm not sure we want to debug every case.

An argument could be made for not retrying in the case of a 403, but in the aforementioned brown-out case it may be nice.

bottlecap/src/bin/bottlecap/main.rs
astuyve: After discussion, it's clear that my use of variable names is very confusing. I'm going to rename them

PR #692, opened 2025-06-03
Combine handle_next_invocation and next_event. Don't block for PlatformReport on timeout shutdowns

astuyve:
1. Refactors handle_next_invocation to call next_event. The methods are separate because we call next_event in an idle loop, but otherwise we always call them together from main, so this simplifies things
2. In the shutdown loop, only block for the report line if the shutdown isn't a timeout. On timeouts, we won't get a report log.

That's not a problem for most use cases where the sandbox will re-initailize, the telemetry API will re-pass events, and then we can forward events on the next invocation. But if a function continuosly times out, we may not forward the custom `task timed out` log until a few invocations down the line.

I'll verify if this is true with OOMs

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -496,12 +496,11 @@ async fn extension_loop_active(
         start_time.elapsed().as_millis().to_string()
     );
     // first invoke we must call next
-    let next_lambda_response = next_event(client, &r.extension_id).await;
     let mut pending_flush_handles = PendingFlushHandles::new();
     let mut last_continuous_flush_error = false;
-    handle_next_invocation(next_lambda_response, invocation_processor.clone()).await;
+    handle_next_invocation(client, &r.extension_id, invocation_processor.clone()).await;
     loop {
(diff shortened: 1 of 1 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -759,39 +764,44 @@ async fn handle_event_bus_event(
...
+    client: &Client,
+    ext_id: &str,
     invocation_processor: Arc<TokioMutex<InvocationProcessor>>,
-) -> bool {
+) -> NextEventResponse {
+    let next_response = next_event(client, ext_id).await;
     match next_response {
         Ok(NextEventResponse::Invoke {
-            request_id,
+            ref request_id,
```
duncanista, line 776: Maybe the `ref` is not required for `request_id`?

duncanista (approved):

bottlecap/src/bin/bottlecap/main.rs
astuyve: image
Nah that's a partial move

PR #695, opened 2025-06-03
poc: App & API Protection

duncanista:


bottlecap/src/appsec/payload/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+
+trait ExtractRequest {
+    const TYPE: RequestType;
+    async fn extract(self) -> HttpData;
+}
+trait ExtractResponse {
+    async fn extract(self) -> HttpData;
+}
+
+pub(super) async fn extract_request_address_data(body: &Bytes) -> Option<(HttpData, RequestType)> {
```
duncanista, line 191: I remember you mentioned maybe we want to eventually consolidate this with the triggers – overall, great

Deserializing into type might be slow since payloads can technically be up to 6MB

duncanista: Ah, next file is doing a similar check from triggers code

duncanista: I see that specific logic, as you mentioned privately, was ported from Go, maybe for the ones that already exist here in the triggers, could we use that same logic and for the events that we don't cover for inferred spans, I.E. Kong, we maintain the Go port?

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -38,6 +40,56 @@ pub struct Context {
...
+
+        // Up-sert any attributes from the AppSec context into the invocation span. This includes
+        // synthetic attributes produced by the WAF (fingerprints, schemas, ...) as well as tracking
+        // spans expected by AAP. We do upsert to avoid overwriting anything previously set.
+        for (key, value) in appsec_context.tags() {
+            self.invocation_span
+                .meta
+                .entry(key.clone())
+                .or_insert(value.clone());
+        }
```
duncanista: Can't we use `extend` here?

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -38,6 +40,56 @@ pub struct Context {
...
+                }
+                Err(e) => {
+                    warn!("appsec: unable to encode WAF events: {e}");
+                }
+            }
+        }
+
+        // Note: We intentionally don't set `actor.ip` because we don't have a definitive signal here.
+
+        //TODO(romain.marcadier): Figure out whether we can set "_dd.runtime_family" here; and to what value.
```
duncanista, line 85: if there is a set of valid values, we can talk about how and when to set it, we get the specific runtime at a later point of init, but we guarantee it will arrive

bottlecap/src/lifecycle/invocation/context.rs
duncanista, line 1: My main concern is if there's a better way to tie contexts, I won't know until I inspect the code thoroughly, but so far this looks good

bottlecap/src/proxy/interceptor.rs
```diff
@@ -172,6 +187,25 @@ async fn invocation_next_proxy(
         });
     }
 
+    // K9 / ASM
+    if let Some(appsec_processor) = appsec_processor {
+        if let Some(request_id) = intercepted_parts
+            .headers
+            .get("Lambda-Runtime-Aws-Request-Id")
```
duncanista, line 192: I remember either in next or response we can get this by checking at the Path, we can leverage Axum for it, will check again which one was it

bottlecap/src/appsec/payload/mod.rs
RomainMuller: I think the current approach is more complex than necessary -- rust deserialization is more strict than Go's (i.e, unlike in Go, not everything is deemed optional)... Probably we can simply do the "try deserializing into the candidate types until one works".

The Kong payload we probably just need to carry a specific model struct for; but I am struggling to find documentation for what shape it's supposed to have from Kong... More research warranted.

bottlecap/src/lifecycle/invocation/context.rs
RomainMuller: Extend will replace existing keys with values found in the iterator, and I precisely want this NOT to happen.

bottlecap/src/lifecycle/invocation/context.rs
RomainMuller: There is indeed... Valid values are: 
- `nodejs`
- `ruby`
- `jvm`
- `dotnet`
- `go`
- `php`
- `python`

bottlecap/src/lifecycle/invocation/context.rs
RomainMuller: I am definitely open to suggestions on that front... The cold start handling is the main reason it is done this way (because its derection relies on the buffer being empty).

bottlecap/src/proxy/interceptor.rs
RomainMuller: The `response` route has the request ID as a path parameter, yes.

For the `next` path, we have to extract the header from the response we intercepted from the "real" runtime API.

bottlecap/LICENSE-3rdparty.yml
duncanista, line 1: This file can be removed

bottlecap/src/lifecycle/invocation/context.rs
duncanista: I see, we can do this at a later stage, but it should be possible as we get the set of runtimes with specific keys, we can just map them to the family later

bottlecap/src/config/yaml.rs
```diff
@@ -880,6 +881,9 @@ extension_version: "compatibility"
                 capture_lambda_payload: true,
                 capture_lambda_payload_max_depth: 5,
                 serverless_appsec_enabled: true,
+                appsec_rules: None,
+                appsec_waf_timeout: Duration::from_millis(5),
+                api_security_sample_delay: Duration::from_secs(30),
```
duncanista: Values need to be merged in trait implementation

bottlecap/src/appsec/processor.rs
```diff
@@ -0,0 +1,1361 @@
...
+    /// - If the [`Config::appsec_rules`] points to a file that is not a valid JSON-encoded ruleset;
+    /// - If the in-app WAF fails to initialize, integrate the ruleset, or build the WAF instance.
+    pub fn new(config: &Config) -> Result<Self, Error> {
+        if !is_enabled(config) {
+            return Err(Error::FeatureDisabled);
+        }
+        debug!("Starting ASM processor");
+
+        if is_standalone() {
+            info!(
```
duncanista, line 37: Should this be a warn?

bottlecap/src/config/yaml.rs
RomainMuller: I believe I have done this? The thing is I don't know that these values are even allowed to be present in the YAML but I don't know if that is allowed by the model (and if so, how to achieve it given the merge trait).

bottlecap/src/appsec/processor.rs
RomainMuller: In principle this is the result of an intentional action by the user... so I doubt it's relevant to _warn_ here? But I don't feel all too strongly about that.

bottlecap/src/config/yaml.rs
duncanista: Yeah, you've done it in `env.rs`, but since every source is different, we have to merge in each one:

https://github.com/DataDog/datadog-lambda-extension/blob/cec942805fe2fb1d62bfd09b51aa94453688fd3d/bottlecap/src/config/yaml.rs#L351-L356

duncanista: Resolved in latest commit

bottlecap/src/appsec/payload/body.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    let body = body.as_ref();
+    let mime_type = match content_type
+        .unwrap_or("application/json")
+        .parse::<mime::Mime>()
+    {
+        Ok(mime) => mime,
+        Err(e) => return Err(e.into()),
+    };
```
duncanista, line 23: Would this actually error at any moment since we're always setting the default as `application/json` which in theory should always parse successfully?

RomainMuller: You can't assume the body is even valid JSON -- it comes from the client-side of the chain... Folks make mistakes!

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -390,7 +408,8 @@ impl Processor {
             _ = offsets.process_chan_tx.send(());
         }
 
-        // todo(duncanista): Add missing metric tags for ASM
+        context.absorb_appsec_tags();
```
duncanista, line 411: I think we could, since this is important part, make this method receive a mutable span which it will add the tags to.

Maybe in the processor, or trace agent, whichever seems fit, we detect if a span is the Lambda span and we absorb the tags there. This would make it work for all runtimes, WDYT?

RomainMuller: Yes I think this is the way forward, really. And this logic can can probably be moved to the `appsec::Processor` as well.

RomainMuller:
Superceded by https://github.com/DataDog/datadog-lambda-extension/pull/755

PR #700, opened 2025-06-06
feat: set init_type on init duration metric

astuyve:
Lambda has been well-known to boost vCPU for on-demand invocations, decreasing the init_duration time because the CPU is not throttled like it is during normal invocations. This has lead to issues like the [Doom Loop](https://aaronstuyvenberg.com/posts/lambda-timeout-doom-loop).

One thing to know is that initializations occurring under `provisioned concurrency` do *not* receive this CPU boost, which leads to longer than expected reported `initialization_duration` latencies. This PR adds the `init_type` tag so users can see the difference. `SnapStart` is never going to be sent, but it's required to satisfy the enum.

Here are the tags being sent for a 128mb function. Note the 11s init duration! (it had to retry during the invoke phase!):
image

Now, note they are equal at 1769mb RAM:
image

As per the Lambda [docs](https://docs.aws.amazon.com/lambda/latest/dg/configuration-memory.html), at 1769mb RAM your function receives the equivalent of 1vCPU.

This PR will help prevent customers from being confused as they may experience longer or shorter initialization durations when using provisioned concurrency, indicating that a faster initialization was likely `on-demand`, vs the much slower `provisioned-concurrency` initialization.

https://datadoghq.atlassian.net/browse/SVLS-6991

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -717,9 +717,17 @@ async fn handle_event_bus_event(
                     p.on_platform_init_start(event.time);
                     drop(p);
                 }
-                TelemetryRecord::PlatformInitReport { metrics, .. } => {
+                TelemetryRecord::PlatformInitReport {
+                    metrics,
+                    initialization_type,
+                    ..
+                } => {
                     let mut p = invocation_processor.lock().await;
-                    p.on_platform_init_report(metrics.duration_ms, event.time.timestamp());
+                    p.on_platform_init_report(
+                        initialization_type,
+                        metrics.duration_ms,
+                        event.time.timestamp(),
+                    );
                     drop(p);
                 }
                 TelemetryRecord::PlatformStart { request_id, .. } => {
--- a/bottlecap/src/telemetry/events.rs
+++ b/bottlecap/src/telemetry/events.rs
@@ -2,6 +2,7 @@ use chrono::{DateTime, Utc};
 
 use serde::Deserialize;
 use serde_json::Value;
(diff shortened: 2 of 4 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -717,9 +717,17 @@ async fn handle_event_bus_event(
                     p.on_platform_init_start(event.time);
                     drop(p);
                 }
-                TelemetryRecord::PlatformInitReport { metrics, .. } => {
+                TelemetryRecord::PlatformInitReport {
+                    metrics,
+                    initialization_type,
+                    ..
```
duncanista, line 723: Would it make sense to use the `InitPhase` (init – invoke) too? 🤔 Although I feel like `InitializationType` is good enough right now

bottlecap/src/telemetry/events.rs
```diff
@@ -163,6 +164,17 @@ pub enum InitType {
     SnapStart,
 }
 
+impl Display for InitType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let style = match self {
+            InitType::OnDemand => "on-demand",
+            InitType::ProvisionedConcurrency => "provisioned-concurrency",
+            InitType::SnapStart => "SnapStart",
```
duncanista, line 172: are we sure we want `SnapStart` instead of `snapstart|snap-start`?

duncanista: Ok, seems like this is a trademark so it should stay as `SnapStart` ™️

PR #706, opened 2025-06-10
feat: Use http2 unless overridden or using a proxy

### LINKED ISSUE #705: HTTP/2 only when sending logs on v81
Hey there,

Upon upgrading to v81, I saw this error popped up

```
DD_EXTENSION | ERROR | Failed to send logs to datadog after 5 ms and 3 attempts: reqwest::Error { kind: Request, url: "https://proxy/", source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Http2, Error { kind: GoAway(b"", FRAME_SIZE_ERROR, Library) })) }
```

It seems like dd forces HTTP/2 when sending logs but not metrics, is there a way to support HTTP/1.1?

astuyve:
We rolled out HTTP/2 support for logs in v81, which seems to have broken logs for some users relying on proxies which may not support http2.

This change introduces a new configuration option called `use_http1`.

1. If `DD_HTTP_PROTOCOL` is explicitly set to http1, we'll use it
2. If `DD_HTTP_PROTOCOL` is not set and the user is using a proxy, we'll use http1 unless overridden by the `DD_HTTP_PROTOCOL` flag being set to anything other than `http1`.

fixes #705

### DIFF
--- a/bottlecap/src/http_client.rs
+++ b/bottlecap/src/http_client.rs
@@ -27,7 +27,7 @@ fn build_client(config: Arc<config::Config>) -> Result<reqwest::Client, Box<dyn 
     // Determine if we should use HTTP/1 or HTTP/2
     let should_use_http1 = match &config.http_protocol {
         // Explicitly set to true - use HTTP/1
-        Some(val) => *val == "http1".to_string(),
+        Some(val) => val == "http1",
         // Not set - use HTTP/1 if proxy is configured, otherwise HTTP/2
         None => config.https_proxy.is_some(),
     };
--- a/bottlecap/tests/logs_integration_test.rs
+++ b/bottlecap/tests/logs_integration_test.rs
@@ -42,7 +42,7 @@ async fn test_logs() {
     });
 
     let arc_conf = Arc::new(Config {
-        use_http1: Some(true),
+        http_protocol: Some("http1".to_string()),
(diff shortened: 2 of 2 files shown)


bottlecap/src/config/env.rs
```diff
@@ -135,6 +136,7 @@ impl Default for Config {
             api_key_secret_arn: String::default(),
             kms_api_key: String::default(),
             serverless_flush_strategy: FlushStrategy::Default,
+            use_http1: None, // None means: use HTTP/2 by default, but fall back to HTTP/1 if proxy is configured
```
astuyve: not sure I love this name, maybe `force_http1` would be better here?

bottlecap/src/config/env.rs
```diff
@@ -39,6 +39,7 @@ pub struct Config {
     pub logs_config_compression_level: i32,
     pub logs_config_logs_dd_url: String,
     pub serverless_flush_strategy: FlushStrategy,
+    pub use_http1: Option<bool>,
```
duncanista: Looks like you might want to use:

https://github.com/DataDog/datadog-agent/blob/39ac529d00ed13a29aeee58c47a6379a08d29999/pkg/config/config_template.yaml#L999-L1002

Not sure we have a `DD_USE_HTTP1` config

PR #707, opened 2025-06-11
feat: parallelize batch flusher

astuyve:
Parallelizes the trace flusher for cases where we send lots of spans. Tested on a function sending 100k spans

### DIFF
--- a/bottlecap/src/traces/trace_flusher.rs
+++ b/bottlecap/src/traces/trace_flusher.rs
@@ -94,14 +94,13 @@ impl TraceFlusher for ServerlessTraceFlusher {
 
         for task in tasks {
             match task.await {
-                Ok(result) => match result {
-                    Ok(_) => {}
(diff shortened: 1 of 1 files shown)


bottlecap/src/traces/trace_flusher.rs
```diff
@@ -82,20 +82,34 @@ impl TraceFlusher for ServerlessTraceFlusher {
...
+        for task in tasks {
+            match task.await {
+                Ok(result) => match result {
+                    Ok(_) => {}
+                    Err(e) => {
+                        error!("Error sending trace: {e:?}");
+                        // Return the original traces for retry
+                        return Some(traces_clone);
+                    }
+                },
```
duncanista:
```suggestion
                Ok(result) => {
                    if let Err(e) = result {
                        ...
                    }
                },
```

PR #708, opened 2025-06-11
feat: use skip missed tick behavior

astuyve:
By default, tokio uses the `burst` behavior which will fire ticks instantly in the case that a task blocks the main thread and the tick is missed. In our case I believe if there is a long/erroneous flush, this may cause the race interval to fire even when we'd reset it.
[docs](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html)

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -489,6 +489,7 @@ async fn extension_loop_active(
         FlushControl::new(config.serverless_flush_strategy, config.flush_timeout);
 
     let mut race_flush_interval = flush_control.get_flush_interval();
+    race_flush_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
     race_flush_interval.tick().await; // discard first tick, which is instantaneous
 
     debug!(


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -489,6 +489,7 @@ async fn extension_loop_active(
         FlushControl::new(config.serverless_flush_strategy, config.flush_timeout);
 
     let mut race_flush_interval = flush_control.get_flush_interval();
+    race_flush_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
```
duncanista: Should we do it inside of the flush control method?

astuyve: yeah

bottlecap/src/lifecycle/flush_control.rs
```diff
@@ -44,12 +44,18 @@ impl FlushControl {
     pub fn get_flush_interval(&self) -> Interval {
         match &self.flush_strategy {
             FlushStrategy::Default => {
-                tokio::time::interval(tokio::time::Duration::from_millis(DEFAULT_FLUSH_INTERVAL))
+                let mut i = tokio::time::interval(tokio::time::Duration::from_millis(
+                    DEFAULT_FLUSH_INTERVAL,
+                ));
+                i.set_missed_tick_behavior(Skip);
```
duncanista, line 50: Just curious, it doesn't let you do something like

```rust
tokio::time::interval(tokio::time::Duration::from_millis(
                    DEFAULT_FLUSH_INTERVAL,
                )).set_missed_tick_behavior(Skip);
```

?

duncanista (approved):

bottlecap/src/lifecycle/flush_control.rs
astuyve: no :( unfortunately it's not a builder-pattern method
image
So it doesn't return anything:
image

PR #709, opened 2025-06-11
chore(config): separate config from sources

duncanista:
# What?

Separates the configuration from sources, allowing it to be used in more use cases.

# How?

Creates new default configuration and separates the environment variables and YAML sources from the default.

# Why?

Make it easier to track changes in every source, as the field names might be different to what they are used at the configuration level.

Also [SVLS-6688](https://datadoghq.atlassian.net/browse/SVLS-6688)

# Notes

I expect to abstract this even more by providing it as a crate which can have features, that way customers can only use the sources and product specific fields they need.

[SVLS-6688]: https://datadoghq.atlassian.net/browse/SVLS-6688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab/datasources/flavors.yaml
+++ b/.gitlab/datasources/flavors.yaml
@@ -6,7 +6,7 @@ flavors:
     needs_layer_publish: true
     suffix: amd64
     layer_name_base_suffix: ""
-    max_layer_compressed_size_mb: 24
+    max_layer_compressed_size_mb: 25
     max_layer_uncompressed_size_mb: 54
 
   - name: arm64


bottlecap/src/config/mod.rs
```diff
@@ -100,164 +368,97 @@ fn fallback(config: &EnvConfig, yaml_config: &YamlConfig) -> Result<(), ConfigEr
...
-            config.https_proxy = Some(https_proxy);
-        }
-    }
+pub fn get_config(config_directory: &Path) -> Result<Config, ConfigError> {
+    let path: std::path::PathBuf = config_directory.join("datadog.yaml");
+    let mut config_builder = ConfigBuilder::default()
+        .add_source(Box::new(YamlConfigSource { path }))
+        .add_source(Box::new(EnvConfigSource));
+
+    let config = config_builder.build();
```
duncanista, line 481: This is the API usage cc @duncanpharvey

duncanista:
Marking it ready to review – as no unit tests fail. 

But I will add more unit tests just to ensure this is extremely reliable and we can trust it more.

bottlecap/src/config/mod.rs
```diff
@@ -9,33 +9,306 @@ pub mod service_mapping;
...
+                    failed_sources += 1;
+                }
+            }
+        }
+
+        if failed_sources == self.sources.len() {
+            debug!("All sources failed to load config, using default config.");
+        }
+
+        if self.config.site.is_empty() {
```
apiarian-datadog, line 181: why is this something we set after all the loading is done rather than starting with it as a default initial value?

bottlecap/src/config/mod.rs
```diff
@@ -9,33 +9,306 @@ pub mod service_mapping;
...
+
+    pub flush_timeout: u64,
+
+    // Proxy
+    pub proxy_https: Option<String>,
+    pub proxy_no_proxy: Vec<String>,
+    pub http_protocol: Option<String>,
+
+    // Endpoints
+    pub dd_url: String,
```
apiarian-datadog, line 248: nit: these are metrics intake urls, not general purpose urls

bottlecap/src/config/mod.rs
```diff
@@ -9,33 +9,306 @@ pub mod service_mapping;
...
+    pub url: String,
+    pub additional_endpoints: HashMap<String, Vec<String>>,
```
apiarian-datadog, line 250: for this (and maybe other non-trivial structures) can we use some sort of aliases which tell us what the key strings are and what the vectors of strings are. especially since this is for something called "additional endpoints" which i would have expected to be... a vector of strings, maybe?

bottlecap/src/config/mod.rs
duncanista: I don't remember the specific reason, but there was something why we did it after merging all configs, will LYK if I find it

bottlecap/src/config/mod.rs
duncanista: Sounds good, will add some comments and docs on the fields

bottlecap/src/config/mod.rs
duncanista: I think this is because how `additional_endpoints` is parsed, but yeah, I can better document this, I am definitely missing that right now, I think it will be per source tho

bottlecap/src/config/env.rs
```diff
@@ -1,238 +1,525 @@
...
+
+    if let Some(otlp_config_metrics_histograms_send_count_sum_metrics) =
+        &env_config.otlp_config_metrics_histograms_send_count_sum_metrics
+    {
+        config
+            .otlp_config_metrics_histograms_send_count_sum_metrics
+            .clone_from(otlp_config_metrics_histograms_send_count_sum_metrics);
+    }
+
+    if let Some(otlp_config_metrics_histograms_send_aggregation_metrics) =
```
apiarian-datadog: could we use macros to generate all of these if statements? or do you think it wouldn't be worth it? how do we ensure that as we add, remove, and otherwise modify the fields in the various structs, we are maintaining parity between them all?

bottlecap/src/config/mod.rs
duncanista: Updated to show that these are from metrics

bottlecap/src/config/mod.rs
duncanista: Added docs, cannot create alias bc of how this is used internally.

bottlecap/src/config/env.rs
duncanista: Although maybe possible during build time, we might change this whole thing to use crate features, so possibly not worth it now

Although I have yet to find a test to make sure every field is used and merged

duncanista: Will update to use macros, still working on how to add unit tests on checking that every field from the config is used

bottlecap/src/config/env.rs
```diff
@@ -1,238 +1,451 @@
...
+    /// setting defined in "site". It does not affect APM, Logs, Remote Configuration,
+    /// or Live Process intake which have their own "*_`dd_url`" settings.
+    ///
+    /// If `DD_DD_URL` and `DD_URL` are both set, `DD_DD_URL` is used in priority.
+    pub dd_url: Option<String>,
+    /// @env `DD_URL`
+    pub url: Option<String>,
+    /// @env `DD_ADDITIONAL_ENDPOINTS`
+    ///
+    /// Additional endpoints to send telemetry to.
```
shreyamalpani: This config is for metrics additional endpoints only, there will be a different config for logs and APM

duncanista: Updated docs!

bottlecap/src/config/env.rs
```diff
@@ -1,238 +1,451 @@
...
+
+    // Metrics
+    /// @env `DD_DD_URL`
+    ///
+    /// The host of the Datadog intake server to send **metrics** to, only set this option
+    /// if you need the Agent to send **metrics** to a custom URL, it overrides the site
+    /// setting defined in "site". It does not affect APM, Logs, Remote Configuration,
+    /// or Live Process intake which have their own "*_`dd_url`" settings.
+    ///
+    /// If `DD_DD_URL` and `DD_URL` are both set, `DD_DD_URL` is used in priority.
```
apiarian-datadog, line 75: nit: could we add some examples for the various `*_dd_url` configs? they're slightly different (including port or not, iirc some of them take a scheme prefix while others don't) so it would be good to include that if we're adding these docs.

bottlecap/src/config/mod.rs
```diff
@@ -9,20 +9,100 @@ pub mod service_mapping;
...
+    flush_strategy::FlushStrategy,
+    log_level::LogLevel,
     processing_rule::{deserialize_processing_rules, ProcessingRule},
-    yaml::Config as YamlConfig,
+    trace_propagation_style::TracePropagationStyle,
+    yaml::YamlConfigSource,
 };
 
-pub type Config = EnvConfig;
+/// Helper macro to merge Option<String> fields to String fields
```
apiarian-datadog, line 32: what if both are `None`?

bottlecap/src/config/mod.rs
apiarian-datadog: this is documented for the env variable. would it be too noisy to document it here on the actual config struct?

apiarian-datadog:
do we have tests which 1) set all the env vars 2) set all the configs through yaml 3) do both and override as expected? 4) exercise at least the url overrides for `site` through both routes?

bottlecap/src/config/mod.rs
duncanista: Wdym, this only merges an `Option<String>` to a `String`, therefore, if the incoming is None, it does nothing 🤔 

This macro would fail during compile time if you pass it to a field which is illegal

apiarian-datadog: ah, ok, the comment says merge two option strings into a string, but it's really merge two option strings into an option string, right?

also, why is merge_string separate from merge_option? is there some subtle rustish difference between them?

bottlecap/src/config/mod.rs
```diff
@@ -31,11 +111,279 @@ pub enum ConfigError {
...
+
+        // If extraction is not set, set it to the same as the propagation style
+        if self.config.trace_propagation_style_extract.is_empty() {
+            self.config
+                .trace_propagation_style_extract
+                .clone_from(&self.config.trace_propagation_style);
+        }
+
+        // If Logs URL is not set, set it to the default
+        if self.config.logs_config_logs_dd_url.is_empty() {
```
apiarian-datadog, line 216: does this mean that `logs_config_logs_dd_url` is technically empty-able but is invalid if so? i.e. that we have a de-facto invalid null state? i guess as long as a user of this package always goes through the builder, they'll get a properly valid value for this setting. but if they happen to make a default onfig, they won't since they'll get an empty string in this place. in this case, is the default method private? that's probably fine then. i'd be a bit concerned if it was publicly accessible, tho.

bottlecap/src/config/mod.rs
duncanista: `merge_string` extracts the value if it is `Some` and sets it as the value in the config field, whereas `merge_option` just checks `if the value is_some()` while not extracting it, and if it `is_some()` then clones the `Option<Type>` into the config field.

It's mainly for merging purposes, some fields are expected to be non-option, whereas some can be in the final config!

bottlecap/src/config/mod.rs
duncanista: The initial value is empty, because we allow the customer to set something, if it keeps being empty, that means that the customer hasn't decided to set any other URL for logs endpoint.

This is logic that previously existed in the config, this is not new.

duncanista: I think to get a config we explicitly expect to use the config builder which should always set these values

Are you suggesting we make the config fields private so they don't get set?

duncanista: Currently config is not accessible and you need to get it by doing a `build()` on the `ConfigBuilder`

bottlecap/src/config/env.rs
```diff
@@ -1,238 +1,778 @@
...
+    /// The flush strategy to use for AWS Lambda.
+    pub serverless_flush_strategy: Option<FlushStrategy>,
+    /// @env `DD_ENHANCED_METRICS`
+    ///
+    /// Enable enhanced metrics for AWS Lambda. Default is `true`.
+    #[serde(deserialize_with = "deserialize_optional_bool_from_anything")]
+    pub enhanced_metrics: Option<bool>,
+    /// @env `DD_LAMBDA_PROC_ENHANCED_METRICS`
+    ///
+    /// Enable enhanced metrics for AWS Lambda. Default is `true`.
```
apiarian-datadog: nit: these docs suggest that `enhanced_metrics` and `lambda_proc_enhanced_metrics` are the same thing? at least they are documented the same.

bottlecap/src/config/env.rs
```diff
@@ -1,238 +1,778 @@
...
+    use super::*;
+    use crate::config::{
+        flush_strategy::{FlushStrategy, PeriodicStrategy},
+        log_level::LogLevel,
+        trace_propagation_style::TracePropagationStyle,
+        Config,
+    };
+
+    #[test]
+    fn test_merge_config_overrides_with_environment_variables() {
```
apiarian-datadog, line 471: is there a way to ensure that we keep this test up-to-date with the base config object changes and with all of the many lines of code in `merge_config`? is that going to rely on us in code review?

bottlecap/src/config/yaml.rs
```diff
@@ -194,94 +209,708 @@ pub struct OtlpReceiverHttpConfig {
...
+#[derive(Debug, PartialEq, Deserialize, Clone, Default)]
 #[serde(default)]
 #[allow(clippy::module_name_repetitions)]
 pub struct OtlpTracesConfig {
-    #[serde(deserialize_with = "deserialize_bool_from_anything")]
-    pub enabled: bool,
-    #[serde(deserialize_with = "deserialize_bool_from_anything")]
-    pub span_name_as_resource_name: bool,
+    #[serde(deserialize_with = "deserialize_optional_bool_from_anything")]
+    pub enabled: Option<bool>,
```
apiarian-datadog, line 222: nit: is an optional bool an omittable bool? or is there some special distinction between null and false?

bottlecap/src/config/yaml.rs
```diff
@@ -194,94 +209,708 @@ pub struct OtlpReceiverHttpConfig {
...
+#[derive(Debug, PartialEq, Clone, Deserialize, Default, Copy)]
+#[serde(default)]
+pub struct OtlpLogsConfig {
+    #[serde(deserialize_with = "deserialize_optional_bool_from_anything")]
+    pub enabled: Option<bool>,
+}
+
+impl OtlpConfig {
+    #[must_use]
+    pub fn receiver_protocols_http_endpoint(&self) -> Option<String> {
```
apiarian-datadog, line 286: this seems to break the pattern of struct fields and replaces it with functions instead. why do we do this?

bottlecap/src/config/yaml.rs
```diff
@@ -194,94 +209,708 @@ pub struct OtlpReceiverHttpConfig {
...
+  no_proxy: ["localhost", "127.0.0.1"]
+dd_url: "https://metrics.datadoghq.com"
+http_protocol: "http1"
+
+# Endpoints
 additional_endpoints:
   "https://app.datadoghq.com":
-    - apikey2
-    - apikey3
+    - "apikey2"
```
apiarian-datadog: is this a sign of a breaking change? do we have to add quotes to strings?

bottlecap/src/config/yaml.rs
```diff
@@ -194,94 +209,708 @@ pub struct OtlpReceiverHttpConfig {
...
+            assert_eq!(config.otlp_config_metrics_delta_ttl, Some(3600));
+            assert_eq!(
+                config.otlp_config_metrics_histograms_mode,
+                Some("counters".to_string())
+            );
+            assert_eq!(
+                config.otlp_config_metrics_histograms_send_count_sum_metrics,
+                true
+            );
+            assert_eq!(
```
apiarian-datadog: would this test be easier to read if we were comparing two full config objects rather than comparing the individual config fields? would that ensure that the test is kept complete since we might have less of an opportunity to change the config object structure without updating the tests?

bottlecap/src/config/env.rs
duncanista: Will update – AI generated this lol

bottlecap/src/config/mod.rs
apiarian-datadog: ah, i see!

bottlecap/src/config/mod.rs
apiarian-datadog: ah, great. if the builder is the only way to get a valid config, i have no objections to this logic living in the builder method.

bottlecap/src/config/env.rs
duncanista: So the tests explicitly checks that the fields in here are not equal to the default, so whenever someone adds anything to `env` or `yaml`, this would error if they haven't added a check

bottlecap/src/config/yaml.rs
duncanista: Sometimes we do define values by default for this – sometimes we dont. This one in particular has a default value of `True` – I think I changed all fields here to `Option<bool>` because we want to merge it at the end of the day, and we want to ensure it exists, as opposed to doing a boolean operation on the field.

This is more of a check of "merge if exists"

bottlecap/src/config/yaml.rs
duncanista: Mainly because this are nested values and getting them is kinda hard, this is already hard to get, makes it harder for us to use the macros here :(

bottlecap/src/config/yaml.rs
duncanista: not really, it doesnt matter, can remove it if you want

bottlecap/src/config/yaml.rs
duncanista: That actually makes sense, will update

bottlecap/src/config/yaml.rs
duncanista: removed quotes

bottlecap/src/config/yaml.rs
duncanista: updated

bottlecap/src/config/env.rs
duncanista: ok with the current unit tests, this should have been fixed, if a new config field is added, then they'd know bc it wouldn't compile in testing

bottlecap/src/config/yaml.rs
apiarian-datadog: you know, it's things like this that make me think we're missing a fourth config type. there's the raw base config, the env and the yaml configs, and then the finalized and ready to use config.

bottlecap/src/config/yaml.rs
apiarian-datadog: ah, ok

bottlecap/src/config/yaml.rs
duncanista: not in this PR please 💀

apiarian-datadog: haha, ok

apiarian-datadog (approved):
looking good, thanks!

PR #711, opened 2025-06-12
chore(otlp): move agent to axum

duncanista:
# What?

Moves OTLP Agent to Axum – also adds a cancellation token to be used on shutdown

# Motivation

Leverage Axum handle connection errors

### DIFF
--- a/bottlecap/src/otlp/agent.rs
+++ b/bottlecap/src/otlp/agent.rs
@@ -1,15 +1,18 @@
-use datadog_trace_agent::http_utils::{
-    log_and_create_http_response, log_and_create_traces_success_http_response,
+use axum::{
+    extract::{FromRequest, Request, State},
+    http::StatusCode,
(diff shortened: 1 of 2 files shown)


bottlecap/src/otlp/agent.rs
```diff
@@ -66,145 +85,125 @@ impl Agent {
...
+
+async fn extract_request_body(
+    request: Request,
+) -> Result<(hyper::http::request::Parts, Bytes), Box<dyn std::error::Error>> {
+    let (parts, body) = request.into_parts();
+    let bytes = Bytes::from_request(Request::from_parts(parts.clone(), body), &()).await?;
+
+    Ok((parts, bytes))
+}
+
```
duncanista: To move to an http utils file?

duncanista: Resolved

PR #716, opened 2025-06-24
chore: Separate AwsCredentials from AwsConfig

lym953:


bottlecap/src/secrets/decrypt.rs
```diff
@@ -228,14 +230,14 @@ async fn request(
 }
 
 fn build_get_secret_signed_headers(
-    aws_config: &AwsConfig,
+    aws_credentials: &AwsCredentials,
     region: String,
```
lym953, line 259: I don't see a region to pass region separately in the old code. It should be the same as the region in aws_config.
In the new code, removing `aws_config` since region is the only field needed from `aws_config`.

astuyve: secrets can be cross region, see: https://github.com/DataDog/datadog-lambda-extension/issues/594

lym953: Got it. Let me add back `aws_config` param.

PR #717, opened 2025-06-24
feat: Lazily resolve api key

lym953:


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -329,33 +330,47 @@ async fn main() -> Result<()> {
...
+            Box::pin(async move {
+                resolve_secrets(config, aws_config, aws_credentials)
+                    .await
+                    .unwrap_or_else(|| {
+                        error!("Failed to resolve API key");
+                        String::new()
+                    })
+            })
+        })))
+    };
```
lym953: During INIT phase, instead of resolving API key, just initialize an API key factory.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -415,11 +430,11 @@ async fn extension_loop_idle(client: &Client, r: &RegisterResponse) -> Result<()
 
 #[allow(clippy::too_many_lines)]
 async fn extension_loop_active(
-    aws_config: &AwsConfig,
+    aws_config: Arc<AwsConfig>,
```
lym953, line 466: Changing it to `Arc` so it can be passed to `ApiKeyFactory` and shared across threads.

bottlecap/src/logs/flusher.rs
```diff
@@ -26,42 +28,24 @@ pub struct Flusher {
     client: reqwest::Client,
     aggregator: Arc<Mutex<Aggregator>>,
     config: Arc<config::Config>,
-    headers: HeaderMap,
+    api_key_factory: Arc<ApiKeyFactory>,
+    headers: OnceCell<HeaderMap>,
```
lym953, line 34: Lazily initialize `headers`, which includes api key.

bottlecap/src/secrets/decrypt.rs
```diff
@@ -14,13 +14,14 @@ use sha2::{Digest, Sha256};
...
+use tokio::sync::RwLock;
 use tracing::debug;
 use tracing::error;
 
 pub async fn resolve_secrets(
     config: Arc<Config>,
-    aws_config: &AwsConfig,
-    aws_credentials: &mut AwsCredentials,
+    aws_config: Arc<AwsConfig>,
+    aws_credentials: Arc<RwLock<AwsCredentials>>,
```
lym953, line 24: A core change: added `RwLock`

bottlecap/src/secrets/decrypt.rs
```diff
@@ -42,30 +43,36 @@ pub async fn resolve_secrets(
                 }
             };
 
-            if aws_credentials.aws_secret_access_key.is_empty()
-                && aws_credentials.aws_access_key_id.is_empty()
-                && !aws_credentials
+            let aws_credentials_read = aws_credentials.read().await;
```
lym953, line 46: The name is not pretty. Let me know if there's a better naming.

bottlecap/src/traces/stats_flusher.rs
```diff
@@ -34,36 +36,47 @@ pub struct ServerlessStatsFlusher {
     // pub buffer: Arc<Mutex<Vec<pb::ClientStatsPayload>>>,
     aggregator: Arc<Mutex<StatsAggregator>>,
     config: Arc<config::Config>,
-    endpoint: Endpoint,
+    api_key_factory: Arc<ApiKeyFactory>,
+    endpoint: OnceCell<Endpoint>,
```
lym953, line 40: Lazily resolve `endpoint`, which contains api key.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -329,33 +330,44 @@ async fn main() -> Result<()> {
...
+
+        Arc::new(ApiKeyFactory::new_from_resolver(Arc::new(move || {
+            let config = Arc::clone(&config);
+            let aws_config = Arc::clone(&aws_config);
+            let aws_credentials = Arc::clone(&aws_credentials);
+
+            Box::pin(async move {
+                resolve_secrets(config, aws_config, aws_credentials)
+                    .await
+                    .expect("Failed to resolve API key")
```
lym953: Any better way to handle this?

duncanista: Mmm ideally, if failing on resolve you'd enter the noop loop, what would happen if you added that here?

lym953: It will panic at flush time and stop the runtime, so I wonder if there's a way for the extension to stop gracefully at that time without stopping the runtime.

bottlecap/src/secrets/decrypt.rs
astuyve: ah and we need an RwLock here because the factory will lazily write/update this struct member?

bottlecap/src/secrets/decrypt.rs
astuyve: this name is fine I think

astuyve (approved):
Nice work Yiming! Let's get this onto self monitoring to test for a few days

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -329,33 +330,44 @@ async fn main() -> Result<()> {
...
+            })
+        })))
+    };
```
duncanista: Wondering if we could make this a method here on in the api key resolver, really don't like the pattern of having the declaration of the same variable 3 times just because we're nesting it, given how main code has increasingly grown, it would be good to hide this in some way and document it

duncanista (approved):
Great PR @lym953 !

bottlecap/src/bin/bottlecap/main.rs
lym953: Done! Extracted a function `create_api_key_factory()`

bottlecap/src/secrets/decrypt.rs
lym953: Yes. The factory fills in `aws_access_key_id`, `aws_secret_access_key` and `aws_session_token` if we are in snap start.

duncanista (approved):

PR #718, opened 2025-06-25
Dual Shipping Logs Support

shreyamalpani:
Adds support for dual shipping logs to endpoints configured using the `logs_config` YAML or `DD_LOGS_CONFIG_ADDITIONAL_ENDPOINTS` env var config.

Implemented a `LogsFlusher` as a wrapper to all the `Flusher` instances to manages flushing to all configured endpoints. 

Moved retry logic to `LogsFlusher`, as the retry request contains the endpoint details and does not have to be tied to a particular flusher.

[SVLS-6886](https://datadoghq.atlassian.net/browse/SVLS-6886)

[SVLS-6886]: https://datadoghq.atlassian.net/browse/SVLS-6886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/logs/flusher.rs
+++ b/bottlecap/src/logs/flusher.rs
@@ -69,15 +69,15 @@ impl Flusher {
         }
     }
 
-    pub async fn flush(&self, batches: Option<Vec<Vec<u8>>>) -> Vec<reqwest::RequestBuilder> {
+    pub async fn flush(&self, batches: Option<Arc<Vec<Vec<u8>>>>) -> Vec<reqwest::RequestBuilder> {
         let mut set = JoinSet::new();
 
         if let Some(logs_batches) = batches {
-            for batch in logs_batches {
+            for batch in logs_batches.iter() {
                 if batch.is_empty() {
                     continue;
                 }
-                let req = self.create_request(batch);
+                let req = self.create_request(batch.clone());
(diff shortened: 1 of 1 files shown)


bottlecap/src/logs/flusher.rs
```diff
@@ -196,3 +182,88 @@ impl Flusher {
...
+                let batches = logs_batches.clone();
+                let flusher = flusher.clone();
+                async move { flusher.flush(Some(batches)).await }
+            });
+
+            let results = join_all(futures).await;
+            for failed in results {
+                failed_requests.extend(failed);
+            }
+        }
```
duncanista: Would it be possible to create the logs_batches, pass as a reference to all flushers, and then drop when used by all flushers? That way we don't clone? 

Not sure if you have answered this question before for metrics, I imagine maybe we do need a copy to send it?

Just LMK! Trying to reduce the memory footprint here

shreyamalpani: Now using `Arc::clone()` instead of `.clone()` in `LogsFlusher`, compressing before we send to each flusher, and explicitly dropping the logs_batch after it has been passed to all flushers

PR #719, opened 2025-06-25
Aj/migrate trace agent to axum

astuyve:
Migrates the trace agent to Axum. Not sure I love using State to hold our mpscs or arc's, but that's the existing pattern and it does work.

this also solves some of the noisy log issues like this one:
```
DD_EXTENSION | DEBUG | Trace agent connection error: error shutting down connection
```

which Axum handles gracefully

### DIFF
--- a/bottlecap/src/traces/stats_processor.rs
+++ b/bottlecap/src/traces/stats_processor.rs
@@ -4,26 +4,30 @@
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use async_trait::async_trait;
-use ddcommon::hyper_migration;
-use hyper::{http, StatusCode};
+use axum::{
+    extract::Request,
+    http::StatusCode,
+    response::{IntoResponse, Response},
+};
 use tokio::sync::mpsc::Sender;
-use tracing::debug;
+use tracing::{debug, error};
 
 use datadog_trace_protobuf::pb;
 use datadog_trace_utils::stats_utils;
+use ddcommon::hyper_migration;
 
 use super::trace_agent::MAX_CONTENT_LENGTH;
 use datadog_trace_agent::http_utils::{self, log_and_create_http_response};
 
 #[async_trait]
 pub trait StatsProcessor {
-    /// Deserializes trace stats from a hyper request body and sends them through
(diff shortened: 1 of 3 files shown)


bottlecap/src/traces/stats_processor.rs
```diff
@@ -4,26 +4,30 @@
...
+    http::StatusCode,
+    response::{IntoResponse, Response},
+};
 use tokio::sync::mpsc::Sender;
-use tracing::debug;
+use tracing::{debug, error};
 
 use datadog_trace_protobuf::pb;
 use datadog_trace_utils::stats_utils;
+use ddcommon::hyper_migration;
```
duncanista, line 17: Do we still need this?

bottlecap/src/traces/stats_processor.rs
```diff
@@ -4,26 +4,30 @@
...
 
 use super::trace_agent::MAX_CONTENT_LENGTH;
```
duncanista, line 19: Can we fix super usage?

bottlecap/src/traces/stats_processor.rs
astuyve: yeah trace_utils still expects it

bottlecap/src/traces/trace_agent.rs
```diff
@@ -133,273 +147,298 @@ impl TraceAgent {
...
-            return response;
+    ) -> Response {
+        let (parts, body) = match extract_request_body(request).await {
+            Ok(r) => r,
+            Err(e) => return error_response(StatusCode::INTERNAL_SERVER_ERROR, e),
+        };
+
+        if let Some(content_length) = parts.headers.get("content-length") {
+            if let Ok(length_str) = content_length.to_str() {
+                if let Ok(length) = length_str.parse::<usize>() {
```
duncanista: i think this could be simplified

duncanista (approved):

bottlecap/src/traces/trace_agent.rs
duncanista: Simplified it so we can merge

PR #724, opened 2025-06-30
fix: move env mapping from ignore_missing_datadog_fields=true to false block

### LINKED ISSUE #723: DD_OTLP_CONFIG_IGNORE_MISSING_DATADOG_FIELDS Creates False Choice Between Environment Mapping and Service/Operation Names
## Summary

The `DD_OTLP_CONFIG_IGNORE_MISSING_DATADOG_FIELDS` configuration flag creates a false choice between proper OTEL environment mapping and proper service/operation name extraction. Users cannot get both behaviors simultaneously when using standard OTEL instrumentation.

## Environment

- **Extension Version**: v81+ (feature introduced in commit 02b5891)
- **Runtime**: .NET 8 with OTEL instrumentation
- **OTEL Libraries**: ADOT .NET libraries
- **Lambda Configuration**: 
  ```hcl
  OTEL_RESOURCE_ATTRIBUTES = "service.namespace=MyApp,deployment.environment=production"
  DD_OTLP_CONFIG_RECEIVER_PROTOCOLS_HTTP_ENDPOINT = "localhost:4318"
  ```

## Expected Behavior

When using standard OTEL semantic conventions, users should be able to:
1. Set service names via OTEL `service.name` resource attribute → Datadog `service` field
2. Set operation names via OTEL span names → Datadog `name` field  
3. Set environment via OTEL `deployment.environment` resource attribute → Datadog `env` field

All three mappings should work simultaneously.

### When `DD_OTLP_CONFIG_IGNORE_MISSING_DATADOG_FIELDS = false` (default):
- ✅ Service names: Correctly mapped from OTEL `service.name`
- ✅ Operation names: Correctly mapped from OTEL span names
- ❌ Environment: NOT mapped from OTEL `deployment.environment`, results in `"env":"none"`

### When `DD_OTLP_CONFIG_IGNORE_MISSING_DATADOG_FIELDS = true`:
- ❌ Service names: Become `"unnamed-service"` (empty string → downstream default)
- ❌ Operation names: Become `"unnamed_operation"` (empty string → downstream default)
- ✅ Environment: Correctly mapped from OTEL `deployment.environment`

## Root Cause Analysis

The flag controls two independent behaviors with opposite logic:

### Code Analysis (`bottlecap/src/otlp/transform.rs`):

**Service/Operation Name Processing (lines 826-837):**
```rust
if !config.otlp_config_ignore_missing_datadog_fields {
    if dd_span.service.is_empty() {
        dd_span.service = get_otel_service(otel_res, true);  // OTEL service.name
    }
    if dd_span.name.is_empty() {
        dd_span.name = get_otel_operation_name(otel_span);   // OTEL span name
    }
}
```

**Environment Processing (lines 1025-1034):**
```rust
if config.otlp_config_ignore_missing_datadog_fields {
    if !dd_span.meta.contains_key("env") {
        let env = get_otel_attribute_value_as_string(
            &otel_res.attributes,
            DEPLOYMENT_ENVIRONMENT,  // "deployment.environment"
            true,
        );
        if !env.is_empty() {
            dd_span.meta.insert("env".to_string(), env);
        }
    }
}
```

**The Problem:** 
- OTEL service/operation mapping happens when flag is `false`
- OTEL environment mapping happens when flag is `true`
- These are mutually exclusive, creating a false choice

### Option 1: Separate Configuration Flags
Replace the single flag with separate controls:
```rust
pub otlp_config_enable_otel_attribute_fallbacks: bool,  // For env mapping
pub otlp_config_disable_auto_generation: bool,          // For service/operation
```

### Option 2: Fix the Logic
Move environment mapping outside the conditional block so it always runs:
```rust
// Always check for OTEL environment mapping
if !dd_span.meta.contains_key("env") {
    let env = get_otel_attribute_value_as_string(&otel_res.attributes, DEPLOYMENT_ENVIRONMENT, true);
    if !env.is_empty() {
        dd_span.meta.insert("env".to_string(), env);
    }
}
```

### Option 3: Invert Environment Logic
Make environment mapping follow the same pattern as service/operation names:
```rust
if !config.otlp_config_ignore_missing_datadog_fields {
    // ... existing service/operation logic ...
    
    // Add environment mapping here
    if !dd_span.meta.contains_key("env") {
        let env = get_otel_attribute_value_as_string(&otel_res.attributes, DEPLOYMENT_ENVIRONMENT, true);
        if !env.is_empty() {
            dd_span.meta.insert("env".to_string(), env);
        }
    }
}
```

## Workaround

Currently, users must choose between:
1. Proper service/operation names with manual `DD_ENV` setting
2. Proper environment mapping with manual Datadog-namespaced attributes

Example workaround for option 1:
```hcl
DD_OTLP_CONFIG_IGNORE_MISSING_DATADOG_FIELDS = "false"
DD_ENV = var.env  # Manual environment setting
```

## Impact

This affects all users trying to use standard OTEL semantic conventions with the Datadog Lambda Extension. The current behavior is counterintuitive and forces users to use non-standard Datadog-specific attributes or environment variables instead of OTEL conventions.

## Additional Context

The flag name `DD_OTLP_CONFIG_IGNORE_MISSING_DATADOG_FIELDS` is also misleading - it actually controls OTEL semantic convention processing, not just "missing field" behavior.

rcdailey:
## Summary
- Fixes environment mapping logic that created a false choice between env field mapping and service/operation name mapping
- Moves OTEL deployment.environment → Datadog env field mapping to work alongside service/operation name mapping when DD_OTLP_CONFIG_IGNORE_MISSING_DATADOG_FIELDS=false (default)
- Resolves issue where spans show env="none" despite correct OTEL resource attributes

## Background
Previously, environment mapping from OTEL `deployment.environment` to Datadog `env` field only worked when `DD_OTLP_CONFIG_IGNORE_MISSING_DATADOG_FIELDS=true`, but service/operation name mapping only worked when this flag was false (default). This created a false choice where users had to choose between environment mapping OR service/operation name mapping.

## Changes
- Moved environment mapping logic from the `config.otlp_config_ignore_missing_datadog_fields = true` block to the `= false` block in `bottlecap/src/otlp/transform.rs`
- Now all OTEL semantic convention mappings work together when the flag is false (default behavior)

## Test Plan
- [x] All existing tests pass (341 tests via cargo nextest)
- [x] Clippy linting passes
- [x] Code formatting verified

Fixes #723

### DIFF
--- a/bottlecap/src/otlp/transform.rs
+++ b/bottlecap/src/otlp/transform.rs
@@ -13,11 +13,12 @@ use opentelemetry_proto::tonic::trace::v1::{
     Span as OtelSpan,
 };
 use opentelemetry_semantic_conventions::attribute::{
-    DB_QUERY_TEXT, DB_STATEMENT, DB_SYSTEM_NAME, DEPLOYMENT_ENVIRONMENT, FAAS_INVOKED_NAME,
-    FAAS_INVOKED_PROVIDER, FAAS_TRIGGER, GRAPHQL_OPERATION_NAME, GRAPHQL_OPERATION_TYPE,
-    HTTP_ROUTE, HTTP_STATUS_CODE, MESSAGING_DESTINATION_NAME, MESSAGING_OPERATION,
-    MESSAGING_SYSTEM, OTEL_LIBRARY_NAME, OTEL_LIBRARY_VERSION, OTEL_STATUS_CODE,
-    OTEL_STATUS_DESCRIPTION, RPC_METHOD, RPC_SERVICE, RPC_SYSTEM,
+    DB_QUERY_TEXT, DB_STATEMENT, DB_SYSTEM_NAME, DEPLOYMENT_ENVIRONMENT,
+    DEPLOYMENT_ENVIRONMENT_NAME, FAAS_INVOKED_NAME, FAAS_INVOKED_PROVIDER, FAAS_TRIGGER,
+    GRAPHQL_OPERATION_NAME, GRAPHQL_OPERATION_TYPE, HTTP_ROUTE, HTTP_STATUS_CODE,
+    MESSAGING_DESTINATION_NAME, MESSAGING_OPERATION, MESSAGING_SYSTEM, OTEL_LIBRARY_NAME,
+    OTEL_LIBRARY_VERSION, OTEL_STATUS_CODE, OTEL_STATUS_DESCRIPTION, RPC_METHOD, RPC_SERVICE,
+    RPC_SYSTEM,
 };
 use opentelemetry_semantic_conventions::resource::{SERVICE_NAME, SERVICE_VERSION}; // CONTAINER_ID, SERVICE_VERSION, TELEMETRY_SDK_LANGUAGE, TELEMETRY_SDK_VERSION,
 use opentelemetry_semantic_conventions::trace::{
@@ -873,13 +874,24 @@ pub fn otel_span_to_dd_span(
                 .insert("http.status_code".to_string(), f64::from(http_status_code));
         }
 
-        // Map OTEL deployment.environment to Datadog env field
+        // Map OTEL deployment.environment to Datadog env field with fallback
         if !dd_span.meta.contains_key("env") {
-            let env = get_otel_attribute_value_as_string(
+            // Try new standard first: deployment.environment.name
+            let mut env = get_otel_attribute_value_as_string(
                 &otel_res.attributes,
-                DEPLOYMENT_ENVIRONMENT,
+                DEPLOYMENT_ENVIRONMENT_NAME,
                 true,
             );
+
+            // Fallback to deprecated deployment.environment if new standard not found
+            if env.is_empty() {
+                env = get_otel_attribute_value_as_string(
+                    &otel_res.attributes,
+                    DEPLOYMENT_ENVIRONMENT,
+                    true,
+                );
+            }
+
             if !env.is_empty() {
                 dd_span.meta.insert("env".to_string(), env);
             }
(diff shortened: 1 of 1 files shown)


bottlecap/src/otlp/transform.rs
```diff
@@ -872,6 +872,18 @@ pub fn otel_span_to_dd_span(
                 .metrics
                 .insert("http.status_code".to_string(), f64::from(http_status_code));
         }
+
+        // Map OTEL deployment.environment to Datadog env field
+        if !dd_span.meta.contains_key("env") {
+            let env = get_otel_attribute_value_as_string(
+                &otel_res.attributes,
+                DEPLOYMENT_ENVIRONMENT,
```
duncanista: Seems like `DEPLOYMENT_ENVIRONMENT` (`deployment.environment`) has been deprecated in favor of `DEPLOYMENT_ENVIRONMENT_NAME` (`deployment.environment.name`)

Could we first check if the the name exist and if not fallback to the environment so it works for both new and older libraries?

rcdailey: Yes that was my understanding as well after reading the otel specification. However, I didn't want to make an assumption about addressing that (somewhat unrelated) concern in this PR. Since you've requested the change, I am more than happy to do this!

duncanista: Thanks @rcdailey,

Overall, we accept any changes that might make sense, we definitely might have missed some stuff here in OTLP conversion, sorry about that

rcdailey: No problem! I appreciate the opportunity to help. I've pushed the requested changes; please let me know if you have any additional feedback. I'll let you resolve this conversation to ensure you are happy with the solution.

rcdailey:
I had to force push once more to fix the incorrect email address on my commits -- I'm currently bouncing between work and personal github accounts and I'm not managing it great! Sorry for the churn!

PR #725, opened 2025-07-01
fix(appsec): remove conflicting appsec_enabled configuration

florentinl:
## Background

i am currently working on Appsec enablement inside the `datadog-lambda-python` tracer and I have noticed that the extension unnecessarily falls back to the go extension when Appsec is enabled at the tracer level.

Appsec inside the extension should only be managed by the `DD_SERVERLESS_APPSEC_ENABLED` env var. 
The `DD_APPSEC_ENABLED` one is reserved for tracer enablement.

## Changes
Remove the conflicting `appsec_enabled` configuration option.

CC @Julio-Guerra

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -427,7 +427,7 @@ fn fallback(config: &Config) -> Result<(), ConfigError> {
     }
 
     if config.serverless_appsec_enabled {
-        log_fallback_reason("serverless_appsec_enabled");
+        log_fallback_reason("appsec_enabled");
         return Err(ConfigError::UnsupportedField(
             "serverless_appsec_enabled".to_string(),
         ));
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -265,9 +265,6 @@ pub struct Config {
     //
     pub service_mapping: HashMap<String, String>,
     //
-    // AppSec
-    pub appsec_enabled: bool,
-    //
     pub apm_dd_url: String,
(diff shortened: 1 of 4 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -50,7 +50,7 @@ fn fallback(config: &EnvConfig, yaml_config: &YamlConfig) -> Result<(), ConfigEr
         ));
     }
 
-    if config.serverless_appsec_enabled || config.appsec_enabled {
+    if config.serverless_appsec_enabled {
         log_fallback_reason("appsec_enabled");
```
RomainMuller, line 430: Probably wanna update the error message here to reflect the serverless appsec enablement being the cause.

florentinl: Absolutely right. Done.

bottlecap/src/config/mod.rs
```diff
@@ -400,9 +396,11 @@ fn fallback(config: &Config) -> Result<(), ConfigError> {
         ));
     }
 
-    if config.serverless_appsec_enabled || config.appsec_enabled {
-        log_fallback_reason("appsec_enabled");
-        return Err(ConfigError::UnsupportedField("appsec_enabled".to_string()));
+    if config.serverless_appsec_enabled {
+        log_fallback_reason("serverless_appsec_enabled");
```
duncanista:
```suggestion
        log_fallback_reason("appsec_enabled");
```

you can leave this one as `appsec_enabled` since this is reported as a metric

PR #729, opened 2025-07-07
chore(telemetry-api): move listener to axum

duncanista:
# What?

Moves the Telemetry API listener from hyper to axum

# Motivation

Let axum handle connections and errors gracefully

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -493,7 +493,7 @@ async fn extension_loop_active(
     let telemetry_listener_cancel_token =
         setup_telemetry_client(&r.extension_id, logs_agent_channel).await?;
 
-    let otlp_shutdown_token = start_otlp_agent(
+    let otlp_cancel_token = start_otlp_agent(
         config,
         tags_provider.clone(),
         trace_processor.clone(),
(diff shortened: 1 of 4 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -1062,8 +1062,8 @@ async fn setup_telemetry_client(
         host: EXTENSION_HOST.to_string(),
         port: TELEMETRY_PORT,
     };
-    let telemetry_listener_cancel_token = tokio_util::sync::CancellationToken::new();
-    let ct_clone = telemetry_listener_cancel_token.clone();
+    let cancel_token = tokio_util::sync::CancellationToken::new();
```
duncanista: It's ok to name it here bc its under a method

bottlecap/src/telemetry/listener.rs
```diff
@@ -23,75 +28,45 @@ impl TelemetryListener {
     pub async fn spin(
         config: &TelemetryListenerConfig,
         event_bus: Sender<TelemetryEvent>,
-        _cancel_token: tokio_util::sync::CancellationToken, // todo cancel token
+        cancel_token: CancellationToken,
```
duncanista: Not sure I like the idea of having the cancel token to be passed, as opposed to be given by the method

bottlecap/src/bin/bottlecap/main.rs
duncanista: moved to inside

bottlecap/src/telemetry/listener.rs
duncanista: moved inside

PR #732, opened 2025-07-07
feat: Handle API key resolution failure

lym953:
# Context
The previous PR https://github.com/DataDog/datadog-lambda-extension/pull/717 defers API key resolution from extension init stage to flush time. However, that PR doesn't well handle the failure case.
- Before that PR, if resolution fails in init stage, the extension will run an idle loop.
- After that PR, the extension will crash at flush time, which will kill the runtime as well, which is not desired.

# What does this PR do?
1. For traces, defer key resolution from `TraceProcessor.process_traces()` to `TraceFlusher.flush()`.
    - (This should ideally be in the previous PR, but since that is already approved, let me add this change in this new PR.)
2. If resolution fails at flush time, then make flush a no-op, so the extension can keep running and consume events without crashing.

# Dependencies
1. https://github.com/DataDog/serverless-components/pull/25
2. https://github.com/DataDog/libdatadog/pull/1140

# Manual Test

## Steps
1. Create a layer in sandbox
2. Apply the layer to a Lambda function
3. Set the env var `DD_API_KEY_SECRET_ARN` to an invalid value
5. Run the Lambda
6. Then set `DD_API_KEY_SECRET_ARN` to a valid value
7. Run the Lambda

## Result
1. The function was successful
image

2. The extension printed some error logs
image

image

3. With valid secret ARN, the Lambda runs successfully and reports to Datadog
image

image

# Automated Test

I didn't add any automated test because from what I see in the codebase, existing tests are usually unit tests for short functions and not for long functions that this PR touches. Please let me know if you think I should add automated tests.

### DIFF
--- a/bottlecap/src/traces/trace_flusher.rs
+++ b/bottlecap/src/traces/trace_flusher.rs
@@ -84,7 +84,7 @@ impl TraceFlusher for ServerlessTraceFlusher {
             let traces: Vec<_> = trace_builders
                 .into_iter()
                 // Lazily set the API key
-                .map(|builder| builder.with_api_key(&api_key))
+                .map(|builder| builder.with_api_key(api_key))
                 .map(SendDataBuilder::build)
                 .collect();
             if let Some(failed) = self.send(traces).await {


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -527,8 +527,12 @@ impl Processor {
...
             .await;
 
-        if let Err(e) = trace_agent_tx.send(send_data).await {
-            debug!("Failed to send context spans to agent: {e}");
+        if let Some(data) = send_data {
+            if let Err(e) = trace_agent_tx.send(data).await {
+                debug!("Failed to send context spans to agent: {e}");
+            }
+        } else {
+            error!("Failed to process traces, skipping send");
```
lym953: `Processor` won't send spans to `TraceAggregator`

bottlecap/src/logs/flusher.rs
```diff
@@ -61,8 +61,12 @@ impl Flusher {
                 if batch.is_empty() {
                     continue;
                 }
-                let req = self.create_request(batch.clone()).await;
-                set.spawn(async move { Self::send(req).await });
+                if let Some(req) = self.create_request(batch.clone()).await {
+                    set.spawn(async move { Self::send(req).await });
+                } else {
+                    error!("Failed to create request");
```
lym953: `Flusher` won't create HTTP requests to send to data to Datadog at `/api/v2/logs`

bottlecap/src/otlp/agent.rs
```diff
@@ -164,33 +164,41 @@ impl Agent {
...
+                Err(err) => {
+                    error!("OTLP | Error sending traces to the trace flusher: {err}");
+                    (
+                            StatusCode::INTERNAL_SERVER_ERROR,
+                            json!({ "message": format!("Error sending traces to the trace flusher: {err}") }).to_string()
+                        ).into_response()
+                }
             }
+        } else {
+            error!("Failed to process traces, skipping send");
```
lym953: OLTP `Agent` won't send traces to `TraceFlusher`

bottlecap/src/traces/stats_flusher.rs
```diff
@@ -60,60 +60,71 @@ impl StatsFlusher for ServerlessStatsFlusher {
...
-            elapsed.as_millis()
-        );
-        match resp {
-            Ok(()) => debug!("Successfully flushed stats"),
-            Err(e) => {
-                error!("Error sending stats: {e:?}");
-            }
-        };
+        } else {
+            error!("Failed to create endpoint");
```
lym953: `ServerlessStatsFlusher` won't send stats to Datadog's endpoint.

bottlecap/src/traces/trace_agent.rs
```diff
@@ -490,13 +491,29 @@ impl TraceAgent {
...
+        if let Some(send_data) = send_data {
+            match trace_tx.send(send_data).await {
+                Ok(()) => success_response("Successfully buffered traces to be flushed."),
+                Err(err) => error_response(
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    format!("Error sending traces to the trace flusher: {err}"),
+                ),
+            }
+        } else {
+            error!("Failed to process traces, skipping send");
```
lym953: `TraceAgent` won't send traces to `TraceFlusher`

bottlecap/src/traces/trace_agent.rs
```diff
@@ -530,57 +547,63 @@ impl TraceAgent {
...
-            Err(err) => error_response(
+            match builder.body(axum::body::Body::from(response_body)) {
+                Ok(response) => response.into_response(),
+                Err(err) => error_response(
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    format!("Error building response: {err}"),
+                ),
+            }
+        } else {
+            error_response(
```
lym953: `TraceAgent` proxy won't send data to Datadog

bottlecap/src/traces/trace_processor.rs
```diff
@@ -162,26 +162,37 @@ impl TraceProcessor for ServerlessTraceProcessor {
...
+                .build();
+            send_data.set_retry_strategy(RetryStrategy::new(
+                1,
+                100,
+                RetryBackoffType::Exponential,
+                None,
+            ));
+            Some(send_data)
+        } else {
+            error!("Failed to resolve API key");
```
lym953: `TraceProcessor` won't process traces

bottlecap/src/traces/stats_flusher.rs
```diff
@@ -60,60 +60,71 @@ impl StatsFlusher for ServerlessStatsFlusher {
...
-                            .expect("can't make URI from stats url, exiting"),
-                        api_key: Some(api_key.clone().into()),
-                        timeout_ms: self.config.flush_timeout * 1_000,
-                        test_token: None,
+                    if let Some(api_key) = self.api_key_factory.get_api_key().await {
+                        let stats_url = trace_stats_url(&self.config.site);
+                        Some(Endpoint {
+                            url: hyper::Uri::from_str(&stats_url)
+                                .expect("can't make URI from stats url, exiting"),
+                            api_key: Some(api_key.to_string().clone().into()),
```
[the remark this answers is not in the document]
lym953: Won't do. I would get an error after doing this: 
image

bottlecap/src/logs/flusher.rs
```diff
@@ -61,8 +61,12 @@ impl Flusher {
...
+                    continue;
+                }
```
ava-silver: [let-else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) is the normal way to avoid extra indentation
```suggestion
                let Some(req) = self.create_request(batch.clone()).await else {
                    error!("Failed to create request");
                    continue;
                }
                set.spawn(async move { Self::send(req).await });
```

lym953: That's the perfect answer!!

bottlecap/src/lifecycle/invocation/processor.rs
duncanista: can't we just avoid the extra allocation by doing if let Some(send_data) = trace_processor... else {} ?

bottlecap/src/logs/flusher.rs
```diff
@@ -93,14 +96,18 @@ impl Flusher {
         failed_requests
     }
 
-    async fn create_request(&self, data: Vec<u8>) -> reqwest::RequestBuilder {
+    async fn create_request(&self, data: Vec<u8>) -> Option<reqwest::RequestBuilder> {
         let url = format!("{}/api/v2/logs", self.endpoint);
-        let headers = self.get_headers().await;
-        self.client
+        let Some(headers) = self.get_headers().await else {
+            return None;
```
duncanista: Mmm, although I understand the code, I find it a little confusing that `get_headers` is responsible for deciding wether or not we're creating a request.

Would it make more sense to rearchitect this so that whenever we definitely know we are about to flush, let's say in `flush()` method, we try to get the API Key?

duncanista (commented):
Left a comment which I would like to see if we can work around. The main idea is, could we rearchitect so that whenever we hit flush we try to resolve the API key and then start doing later work based on it? Instead, we're failing in headers when trying to get an API key, but this looks like they should be separated 🤔 

LMK what you think

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -78,6 +79,8 @@ pub struct Processor {
     ///
     /// These tags are used to capture runtime and initialization.
     dynamic_tags: HashMap<String, String>,
+    /// Function to resolve Datadog API key.
+    api_key_factory: Arc<ApiKeyFactory>,
```
lym953: Add it to an outer struct `Processor`

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -502,6 +507,11 @@ impl Processor {
         trace_processor: &Arc<dyn TraceProcessor + Send + Sync>,
         trace_agent_tx: &Sender<SendData>,
     ) {
+        let Some(api_key) = self.api_key_factory.get_api_key().await else {
```
lym953: ... so we can abort earlier here, without needing to touch many functions

bottlecap/src/logs/flusher.rs
```diff
@@ -55,13 +55,18 @@ impl Flusher {
 
     pub async fn flush(&self, batches: Option<Arc<Vec<Vec<u8>>>>) -> Vec<reqwest::RequestBuilder> {
         let mut set = JoinSet::new();
+        let api_key = self.api_key_factory.get_api_key().await;
+        let Some(api_key) = api_key else {
```
lym953: Abort early at the beginning of `Flusher.flush()`

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -516,19 +526,20 @@ impl Processor {
...
-        let send_data: SendData = trace_processor
+        let send_data = trace_processor
             .process_traces(
                 self.config.clone(),
                 tags_provider.clone(),
                 header_tags,
                 vec![traces],
                 body_size,
                 self.inferrer.span_pointers.clone(),
+                api_key,
```
lym953: Passing `api_key` to `process_traces()`, so `process_traces()` won't need to handle failure inside.

bottlecap/src/logs/flusher.rs
```diff
@@ -146,10 +151,9 @@ impl Flusher {
         }
     }
 
-    async fn get_headers(&self) -> &HeaderMap {
+    async fn get_headers(&self, api_key: &str) -> &HeaderMap {
```
lym953: Passing in `api_key`, so `get_headers()` won't need to handle the failure

bottlecap/src/otlp/agent.rs
```diff
@@ -30,6 +32,7 @@ type AgentState = (
     OtlpProcessor,
     Arc<dyn TraceProcessor + Send + Sync>,
     Sender<SendData>,
+    Arc<ApiKeyFactory>,
```
lym953: Adding it to `AgentState` and `Agent`

bottlecap/src/otlp/agent.rs
```diff
@@ -127,9 +134,20 @@ impl Agent {
     }
 
     async fn v1_traces(
-        State((config, tags_provider, processor, trace_processor, trace_tx)): State<AgentState>,
+        State((config, tags_provider, processor, trace_processor, trace_tx, api_key_factory)): State<AgentState>,
         request: Request,
     ) -> Response {
+        let Some(api_key) = api_key_factory.get_api_key().await else {
```
lym953: Abort at the beginning of `v1_traces` API handler

bottlecap/src/traces/stats_flusher.rs
```diff
@@ -60,16 +60,20 @@ impl StatsFlusher for ServerlessStatsFlusher {
             return;
         }
 
+        let Some(api_key) = self.api_key_factory.get_api_key().await else {
```
lym953, line 63: Abort at the beginning of `StatsFlusher.send()`

bottlecap/src/traces/trace_agent.rs
```diff
@@ -405,7 +411,15 @@ impl TraceAgent {
         invocation_processor: Arc<Mutex<InvocationProcessor>>,
         tags_provider: Arc<provider::Provider>,
         version: ApiVersion,
+        api_key_factory: Arc<ApiKeyFactory>,
     ) -> Response {
+        let Some(api_key) = api_key_factory.get_api_key().await else {
```
lym953: Abort at the beginning of v0.4 and v0.5 traces API handler

bottlecap/src/traces/trace_agent.rs
```diff
@@ -531,6 +546,14 @@ impl TraceAgent {
             Err(e) => return error_response(StatusCode::INTERNAL_SERVER_ERROR, e),
         };
 
+        let Some(api_key) = api_key_factory.get_api_key().await else {
```
lym953: Abort at the beginning of `handle_proxy()`

bottlecap/src/otlp/agent.rs
lym953: Is it possible that the customer's code calls `/v1/traces` api synchronously, and we slow down the customer's Lambda by doing the heavy operation of resolving api key here?
If so, it might be better to further defer key resolution by moving it out of the API handler.

bottlecap/src/traces/trace_agent.rs
lym953: dit: Is it possible that the customer's code calls traces api synchronously, and we slow down the customer's Lambda by doing the heavy operation of resolving api key here?
If so, it might be better to further defer key resolution by moving it out of the API handler.

lym953:
> Left a comment which I would like to see if we can work around. The main idea is, could we rearchitect so that whenever we hit flush we try to resolve the API key and then start doing later work based on it? Instead, we're failing in headers when trying to get an API key, but this looks like they should be separated

@duncanista Good point! Made a lot of changes.

One concern is this PR (and the last one) only defers key resolution from init time to trace API handler (if trace API handler) is called, not to flush time. Although it can shorten cold start, it can make `invoke` phase slower. Is that a problem? (Correct me if my understanding is wrong.)

bottlecap/src/traces/trace_agent.rs
astuyve: yeah we need to defer this to the flusher as this API is called synchronously

bottlecap/src/otlp/agent.rs
```diff
@@ -22,6 +22,8 @@ use crate::{
     traces::trace_processor::TraceProcessor,
 };
 
+use dogstatsd::api_key::ApiKeyFactory;
```
duncanista: Question, do we want to move this to its own file? Wondering if all other components should be relying in dogstatsd as a dependency just for an ApiKeyFactory

duncanista: Not needed to be done now, but would be good to not make them dependent on a metrics module

bottlecap/src/otlp/agent.rs
duncanista: This gets called by an exporter, probably at the end of the function, so yeah, it would be done in runtime time

bottlecap/src/config/mod.rs
```diff
@@ -197,7 +197,7 @@ impl ConfigBuilder {
         // or in the `proxy_no_proxy` config field.
         if self.config.proxy_https.is_some() {
             let site_in_no_proxy = std::env::var("NO_PROXY")
-                .map_or(false, |no_proxy| no_proxy.contains(&self.config.site))
+                .is_ok_and(|no_proxy| no_proxy.contains(&self.config.site))
```
lym953: fixes a new clippy error due to upgrade

bottlecap/src/config/mod.rs
```diff
@@ -543,7 +543,7 @@ where
 {
     struct KeyValueVisitor;
 
-    impl<'de> serde::de::Visitor<'de> for KeyValueVisitor {
+    impl serde::de::Visitor<'_> for KeyValueVisitor {
```
lym953: fixes a new clippy error due to upgrade

bottlecap/src/traces/stats_aggregator.rs
```diff
@@ -1,6 +1,7 @@
 use datadog_trace_protobuf::pb::ClientStatsPayload;
 use std::collections::VecDeque;
 
+#[allow(clippy::empty_line_after_doc_comments)]
```
lym953: New clippy error

bottlecap/src/traces/trace_agent.rs
```diff
@@ -85,7 +84,6 @@ pub struct StatsState {
 pub struct ProxyState {
     pub config: Arc<config::Config>,
     pub proxy_aggregator: Arc<Mutex<proxy_aggregator::Aggregator>>,
-    pub api_key_factory: Arc<ApiKeyFactory>,
```
lym953, line 89: Moved to `trace_flusher`.

bottlecap/src/traces/trace_flusher.rs
```diff
@@ -57,6 +76,10 @@ impl TraceFlusher for ServerlessTraceFlusher {
         // Process new traces from the aggregator
         let mut guard = self.aggregator.lock().await;
         let mut traces = guard.get_batch();
+        // Lazily set the API key
+        for trace in &mut traces {
+            trace.get_target_mut().api_key = Some(api_key.to_string().into());
```
lym953: I need to add `get_target_mut()` in https://github.com/DataDog/libdatadog/pull/1140

PR #733, opened 2025-07-08
chore:add gitlab task to deploy to sandbox us-west-2 for e2e test

joeyzhao2018:


.gitlab/templates/pipeline.yaml.tpl
```diff
@@ -227,6 +227,34 @@ publish layer sandbox [us-east-1] ({{ $flavor.name }}):
   script:
     - .gitlab/scripts/publish_layers.sh
 
+publish layer sandbox [us-west-2] ({{ $flavor.name }}):
```
duncanista: I think this is unnecessary, since we already have a self monitoring stage, we can just reuse the same job and make it a range

purple4reina (commented):
Nice!

PR #737, opened 2025-07-10
feat(apm): aggregate and flush data accordingly for DSM, Profiling, LLMObs, and Live Debugger

duncanista:
# What?

Creates an aggregator and flusher for APM proxy data.

# Motivation

With the introduction of #684, the Extension is way faster to continue invocations, I think this could have affected how data is aggregated for this various APM products in which we were just sending the request to intake as soon as it arrives.

Now, we aggregate it and respect the flushing pattern and algorithm.

# Testing

1. Profiling:
Screenshot 2025-07-11 at 11 13 55 AM

2. LLMObs:
Screenshot 2025-07-15 at 3 12 01 PM

3. DSM:
tbd

4. Live Debugger:
image

### DIFF
--- a/bottlecap/src/traces/proxy_aggregator.rs
+++ b/bottlecap/src/traces/proxy_aggregator.rs
@@ -14,7 +14,7 @@ pub struct Aggregator {
 impl Default for Aggregator {
     fn default() -> Self {
         Aggregator {
-            queue: Vec::with_capacity(128), // arbritrary capacity for request queue
+            queue: Vec::with_capacity(128), // arbitrary capacity for request queue
         }
     }
 }
--- a/bottlecap/src/traces/proxy_aggregator.rs
+++ b/bottlecap/src/traces/proxy_aggregator.rs
@@ -1,0 +1,30 @@
+use bytes::Bytes;
+use reqwest::header::HeaderMap;
+
+pub struct ProxyRequest {
+    pub headers: HeaderMap,
+    pub body: Bytes,
+    pub target_url: String,
+}
+
+pub struct Aggregator {
+    queue: Vec<ProxyRequest>,
+}
+
+impl Default for Aggregator {
+    fn default() -> Self {
+        Aggregator {
+            queue: Vec::with_capacity(128), // arbritrary capacity for request queue
+        }
+    }
+}
+
+impl Aggregator {
+    pub fn add(&mut self, request: ProxyRequest) {
+        self.queue.push(request);
+    }
+
+    pub fn get_batch(&mut self) -> Vec<ProxyRequest> {
+        std::mem::take(&mut self.queue)
+    }
+}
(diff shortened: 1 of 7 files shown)


bottlecap/src/traces/proxy_aggregator.rs
duncanista, line 1: Don't like how simple this aggregator is, its just a wrapper for a queue, but I think so far we should respect the way we are following this pattern, maybe we can create a trait and consolidate how we create aggregators

bottlecap/src/traces/proxy_flusher.rs
```diff
@@ -0,0 +1,186 @@
...
+                    if status == 202 || status == 200 {
+                        debug!(
+                            "Proxy Flusher | Successfully sent request in {} ms",
+                            elapsed.as_millis()
+                        );
+                    } else {
+                        error!("Proxy Flusher | Request failed with status {status}: {body:?}");
+                    }
+
+                    return Ok(());
```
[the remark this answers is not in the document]
duncanista: This is on purpose, if we encounter something different than 200, meaning a 400, 403, we might want to not send it to not timeout

bottlecap/src/traces/proxy_flusher.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        // Remove headers that are not needed for the proxy request
+        incoming_headers.remove("host");
+        incoming_headers.remove("content-length");
+
+        incoming_headers.extend(self.headers.clone());
+
+        self.client
+            .post(&request.target_url)
+            .headers(request.headers)
```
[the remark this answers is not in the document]
duncanista: Thanks, solved it!

bottlecap/src/traces/proxy_aggregator.rs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+pub struct Aggregator {
+    queue: Vec<ProxyRequest>,
+}
+
+impl Default for Aggregator {
+    fn default() -> Self {
+        Aggregator {
+            queue: Vec::with_capacity(128), // arbritrary capacity for request queue
```
lym953: typo: arbritrary

bottlecap/src/traces/proxy_flusher.rs
```diff
@@ -0,0 +1,186 @@
...
+            client,
+            aggregator,
+            config,
+            headers,
+        }
+    }
+
+    pub async fn flush(
+        &self,
+        retry_requests: Option<Vec<reqwest::RequestBuilder>>,
```
lym953, line 69: For my learning: is there a limit on the number of retries? I assumed the retry count in `send()` is a different things.

duncanista: The retry count in the `send` method is mainly for the amount of times we try to retry a payload during a flush.

I think there's no limit on the number of failed payloads we try to retry – this is mainly bc we could have a network error which we need to handle.

For 400/500 errors we don't retry, maybe we should at some point, but we want to ensure we don't timeout the lambda or fill the buffer with data that will never arrive

astuyve (approved):

PR #740, opened 2025-07-11
chore: Add doc and rename function for flushing strategy

lym953:
# Motivation

It took me quite some effort to understand flushing strategies.

# This PR
Tries to make flushing strategy code more readable:
1. Add/move comments
2. Create an enum `ConcreteFlushStrategy`, which doesn't contain `Default` because it is required to be resolved to a concrete strategy
3. Rename `should_adapt` to `evaluated_flush_strategy()`

# To reviewers
There are still a few things I don't understand, which are marked with `TODO`. Appreciate explanation!

### DIFF
--- a/bottlecap/src/config/flush_strategy.rs
+++ b/bottlecap/src/config/flush_strategy.rs
@@ -8,11 +8,40 @@ pub struct PeriodicStrategy {
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum FlushStrategy {
+    // Flush every 1s and at the end of the invocation
     Default,
+    // User specifies the interval in milliseconds, will not block on the runtimeDone event
+    Periodically(PeriodicStrategy),
+    // Always flush at the end of the invocation
+    End,
+    // Flush both (1) at the end of the invocation and (2) periodically with the specified interval
+    EndPeriodically(PeriodicStrategy),
+    // Flush in a non-blocking, asynchronous manner, so the next invocation can start without waiting
+    // for the flush to complete
+    Continuously(PeriodicStrategy),
+}
+
+// A restricted subset of `FlushStrategy`. The Default strategy is now allowed, which is required to be
+// translated into a concrete strategy.
+#[allow(clippy::module_name_repetitions)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum ConcreteFlushStrategy {
+    Periodically(PeriodicStrategy),
     End,
     EndPeriodically(PeriodicStrategy),
-    Periodically(PeriodicStrategy),
     Continuously(PeriodicStrategy),
+}
+
+impl From<FlushStrategy> for ConcreteFlushStrategy {
+    fn from(strategy: FlushStrategy) -> ConcreteFlushStrategy {
+        match strategy {
+            FlushStrategy::Periodically(p) => ConcreteFlushStrategy::Periodically(p),
+            FlushStrategy::End => ConcreteFlushStrategy::End,
+            FlushStrategy::Continuously(p) => ConcreteFlushStrategy::Continuously(p),
+            FlushStrategy::EndPeriodically(p) => ConcreteFlushStrategy::EndPeriodically(p),
+            FlushStrategy::Default => unreachable!("Default strategy is not allowed"),
+        }
+    }
 }
 
 // Deserialize for FlushStrategy
--- a/bottlecap/src/lifecycle/flush_control.rs
+++ b/bottlecap/src/lifecycle/flush_control.rs
@@ -1,4 +1,4 @@
-use crate::config::flush_strategy::FlushStrategy;
+use crate::config::flush_strategy::{ConcreteFlushStrategy, FlushStrategy};
 use std::time;
 use tokio::time::{Interval, MissedTickBehavior::Skip};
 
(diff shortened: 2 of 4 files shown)


bottlecap/src/config/flush_strategy.rs
```diff
@@ -8,13 +8,42 @@ pub struct PeriodicStrategy {
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum FlushStrategy {
+    // Flush every 1s and at the end of the invocation
```
lym953, line 11: Some doc is moved from `flush_control.rs`

bottlecap/src/lifecycle/flush_control.rs
```diff
@@ -57,12 +52,15 @@ impl FlushControl {
                 i.set_missed_tick_behavior(Skip);
                 i
             }
+            // TODO: Why is this 15 minutes?
```
lym953: Need help here

bottlecap/src/lifecycle/invocation_times.rs
```diff
@@ -40,22 +47,23 @@ impl InvocationTimes {
         let should_adapt = (elapsed as f64 / (LOOKBACK_COUNT - 1) as f64) < ONE_TWENTY_SECONDS;
         if should_adapt {
             // Both units here are in seconds
+            // TODO: What does this mean?
```
lym953: Need help here

bottlecap/src/lifecycle/flush_control.rs
astuyve: Looking at the history/blame can be helpful here: https://github.com/DataDog/datadog-lambda-extension/pull/599

This timeout is used for the race flush block. But when users want to flush at end, we should never hit the race flush. Initially it was set to a max unsigned int but that caused tokio to integer overflow when calculating the timeout

bottlecap/src/lifecycle/invocation_times.rs
astuyve: The comment addresses an earlier misconception when the author (me) thought that the flush_timeout was in milliseconds, so the elapsed time was then in seconds and it never hit.

The logic says that if we should adapt to a periodic strategy (which is defined above on 47, if the last 20 requests arrived within 2 minutes). The next bit decides if we should use the continuous strategy or if we should use a regular periodic strategy, which from your comment above on line 37-41 seems you've got handled pretty well.

I hope this helps!

astuyve (approved):

bottlecap/src/lifecycle/invocation_times.rs
lym953: Sorry if I was not clear. What I don't understand is why compare `elapsed` with `flush_timeout`.

For example, if `flush_timeout` is 30s, and there's one invocation every 1.6s, then `elapsed` will be 1.6s * 19 = 30.4s, and we will choose the synchronous "periodic" strategy with 20s interval. If flush takes 29s, then this will block lots of invocations.

Should we compare the average invocation interval (`elapsed / (LOOKBACK_COUNT - 1)`) with `flush_timeout` instead?

astuyve: The idea is that we want to make sure the periodicity is fast enough that nonblocking flushing is unlikely to be broken by lambda pausing/freezing the CPU between invocations. Because the client will timeout after the `flush_timeout`, we make this comparison. We're just choosing 20 invocations arbitrarily to pick "very fast functions".

It's unrelated to `what happens if a periodic flush takes 29s`. If we choose the periodic strategy, we'll block on the call to `/next` until the periodic flush is complete. If we choose the continuous flush, we don't.

bottlecap/src/lifecycle/invocation_times.rs
```diff
@@ -23,39 +23,63 @@ impl InvocationTimes {
...
+                // We get the average time between each invocation by taking the difference between newest (`now`) and the
+                // oldest invocation in the buffer, then dividing by `LOOKBACK_COUNT - 1`.
+                let oldest = self.times[self.head];
+
+                let elapsed = now - oldest;
+                let should_adapt =
+                    (elapsed as f64 / (LOOKBACK_COUNT - 1) as f64) < ONE_TWENTY_SECONDS;
+                if should_adapt {
+                    // Both units here are in seconds
+                    // TODO: What does this mean?
```
astuyve: Did that answer your question (in the other thread)?

astuyve (approved):

bottlecap/src/lifecycle/invocation_times.rs
lym953: I get the overall idea that the threshold of avg interval should be positively correlated to `flush_timeout`, but I don't get why we directly compare `elapsed` (which is 19 * avg interval) with `flush_timeout`. I'm thinking of an alternative threshold, which more aggressively chooses the continuous strategy: (ignore int/float conversions and division errors in the code below)
```
let avg_interval = elapsed / (LOOKBACK_COUNT - 1);
if avg_interval * 4 < flush_timeout {
  return Continuously;
} else {
  return Periodically;
}
```
This is similar to the current code, except that we use 4 here instead of 19. About this threshold:
1. Usually it's not likely that the actual interval between any two invocations is longer than `4 * avg_interval`, so the freezing duration is unlikely longer than `4 * avg_interval`, so the risk of flush timeout is small.
2. On the up side, it brings the benefits of the continuous strategy for a wider range of `avg_interval` values, i.e. we will transition from `Periodic` to `Continuous` when it's in the range of `(flush_timeout / 19, flush_timeout / 4)`, i.e. `(1.58s, 7.5s)` if `flush_timeout == 30s`.

What do you think?

We can talk in person tomorrow if that's easier.

PR #742, opened 2025-07-14
chore: Add workflow_dispatch trigger for rs_ci.yml

lym953:
# This PR
Allows the workflow `rs_ci.yml` to be triggered manually.

# Motivation
Right now `rs_ci.yml` can only use the workflow from the main branch, making it hard to test https://github.com/DataDog/datadog-lambda-extension/pull/741.

### DIFF
--- a/.github/workflows/rs_ci.yml
+++ b/.github/workflows/rs_ci.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 env:
   PB_VERSION: 25.3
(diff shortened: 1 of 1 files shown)


.github/workflows/rs_ci.yml
```diff
@@ -146,7 +147,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-22.04, macos-latest ]
+        os: [ubuntu-22.04, macos-latest]
```
lym953, line 150: Auto format on save

PR #745, opened 2025-07-14
chore: pass SendDataBuilderInfo instead of SendData until flush time

lym953:
# Background
Right now `SendData` is passed around across channels.

# This PR

Instead of passing `SendData`, pass `SendDataBuilderInfo`, which bundles `SendDataBuilder` and payload size. Just before flush, call `SendDataBuilder.build()` to build `SendData`.

# Motivation
https://github.com/DataDog/libdatadog/pull/1140#discussion_r2205269256 It is suggested that the function `set_api_key()` shouldn't be added on `SendData`, but should be added on `SendDataBuilder`. Because need to call `set_api_key()` just before flush, we need to make sure the object is `SendDataBuilder` instead of `SendData` until flush time.

And because we need payload size in Trace Aggregator, and `SendDataBuilder` doesn't expose this field, we need to pass it explicitly along with `SendDataBuilder`.

# Next steps
Update https://github.com/DataDog/datadog-lambda-extension/pull/717 https://github.com/DataDog/datadog-lambda-extension/pull/732 so that `get_api_key()` is called just before flush.

# Dependency
https://github.com/DataDog/libdatadog/pull/1140


bottlecap/src/otlp/agent.rs
```diff
@@ -157,7 +157,7 @@ impl Agent {
 
         let tracer_header_tags: DatadogTracerHeaderTags = (&parts.headers).into();
         let body_size = size_of_val(&traces);
-        let send_data = trace_processor.process_traces(
+        let send_data_builder = trace_processor.process_traces(
```
astuyve, line 168: Check if traces vec is empty here

bottlecap/src/traces/trace_aggregator.rs
```diff
@@ -7,11 +7,24 @@ use std::collections::VecDeque;
...
+    pub builder: SendDataBuilder,
+    pub size: usize,
+}
+
+impl SendDataBuilderInfo {
+    pub fn new(builder: SendDataBuilder, size: usize) -> Self {
+        Self { builder, size }
+    }
+}
+
```
lym953, line 22: Adding `SendDataBuilderInfo` in `trace_aggregator.rs` because this is where the `size` is consumed. Let me know if you have a better place.

bottlecap/src/traces/trace_aggregator.rs
```diff
@@ -35,26 +48,26 @@ impl TraceAggregator {
...
+    pub fn get_batch(&mut self) -> Vec<SendDataBuilder> {
         let mut batch_size = 0;
 
         // Fill the batch
         while batch_size < self.max_content_size_bytes {
-            if let Some(payload) = self.queue.pop_front() {
+            if let Some(payload_info) = self.queue.pop_front() {
                 // TODO(duncanista): revisit if this is bigger than limit
-                let payload_size = payload.len();
+                let payload_size = payload_info.size;
```
lym953, line 62: `size` is consumed here

bottlecap/src/traces/trace_flusher.rs
```diff
@@ -56,17 +59,21 @@ impl TraceFlusher for ServerlessTraceFlusher {
...
         // Process new traces from the aggregator
         let mut guard = self.aggregator.lock().await;
-        let mut traces = guard.get_batch();
+        let mut trace_builders = guard.get_batch();
 
-        while !traces.is_empty() {
+        while !trace_builders.is_empty() {
+            let traces: Vec<_> = trace_builders
+                .into_iter()
+                .map(SendDataBuilder::build)
```
lym953, line 67: `build()` is called here

bottlecap/src/traces/trace_processor.rs
```diff
@@ -166,17 +168,16 @@ impl TraceProcessor for ServerlessTraceProcessor {
             test_token: None,
         };
 
-        let send_data_builder = SendDataBuilder::new(body_size, payload, header_tags, &endpoint);
-        let mut send_data = send_data_builder
+        let builder = SendDataBuilder::new(body_size, payload, header_tags, &endpoint)
             .with_compression(Compression::Zstd(6))
-            .build();
```
lym953, line 172: `build()` is no longer called here

bottlecap/src/traces/trace_processor.rs
```diff
@@ -166,17 +168,16 @@ impl TraceProcessor for ServerlessTraceProcessor {
...
-        ));
-        send_data
+            .with_retry_strategy(RetryStrategy::new(
+                1,
+                100,
+                RetryBackoffType::Exponential,
+                None,
+            ));
+
+        SendDataBuilderInfo::new(builder, body_size)
```
lym953, line 180: `SendDataBuilderInfo` is constructed here

PR #746, opened 2025-07-15
Aj/fix shutdown deadlock

astuyve:
We added this check for the platform.Report item because we'd occasionally race past it when shutting down, leading to missing metrics (namely post-runtime duration for the last request). This predominantly affected hourly/infrequent functions.

But that change introduced a race condition where sometimes Lambda would respond to /next with the shutdown event at the same time we'd receive the last invocation's report event. In some cases, Tokio would run the report event first followed by the response to /next, where then we'd hit the shutdown event bus loop and wait for a report event that was never coming.

This change also adds a 300ms timeout in case of a deadlock.

Jira: https://datadoghq.atlassian.net/browse/SLES-2151

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -679,20 +679,11 @@ async fn extension_loop_active(
             'shutdown: loop {
                 tokio::select! {
                     Some(event) = event_bus.rx.recv() => {
-                        match event {
-                            Event::Telemetry(telemetry_event) => {
-                                if let TelemetryRecord::PlatformTombstone = telemetry_event.record {
-                                    debug!("Received tombstone event, proceeding with shutdown");
-                                    break 'shutdown;
-                                }
-                                // Process other telemetry events normally
-                                handle_event_bus_event(Event::Telemetry(telemetry_event), invocation_processor.clone(), tags_provider.clone(), trace_processor.clone(), trace_agent_channel.clone()).await;
-                            }
-                            other_event => {
-                                // Process non-telemetry events
-                                handle_event_bus_event(other_event, invocation_processor.clone(), tags_provider.clone(), trace_processor.clone(), trace_agent_channel.clone()).await;
-                            }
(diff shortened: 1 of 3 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -669,29 +669,36 @@ async fn extension_loop_active(
...
                                     break 'shutdown;
                                 }
+                                // Process other telemetry events normally
+                                handle_event_bus_event(Event::Telemetry(telemetry_event), invocation_processor.clone(), tags_provider.clone(), trace_processor.clone(), trace_agent_channel.clone()).await;
+                            }
+                            other_event => {
+                                // Process non-telemetry events
+                                handle_event_bus_event(other_event, invocation_processor.clone(), tags_provider.clone(), trace_processor.clone(), trace_agent_channel.clone()).await;
                             }
                         }
```
lym953: nit: Is there a reason we prefer `match` to `if let`? How about this:
```
if let Event::Telemetry(ref telemetry_event) = event {
    if let TelemetryRecord::PlatformTombstone = telemetry_event.record {
        debug!("Received tombstone event, proceeding with shutdown");
        break 'shutdown;
    }
}
handle_event_bus_event(event, invocation_processor.clone(), tags_provider.clone(), trace_processor.clone(), trace_agent_channel.clone()).await;
```
so no need to write `handle_event_bus_event()` twice?

lym953: Or make it more compact:
```
if let Event::Telemetry(TelemetryEvent { record: TelemetryRecord::PlatformTombstone, .. }) = event {
    debug!("Received tombstone event, proceeding with shutdown");
    break 'shutdown;
}
handle_event_bus_event(event, invocation_processor.clone(), tags_provider.clone(), trace_processor.clone(), trace_agent_channel.clone()).await;
```

astuyve: yeah this is better, thanks Yiming!

PR #751, opened 2025-07-18
Enhance Synthetic Span Service Representation

zarirhamza:
### What does this PR do?

Rollout of span naming changes to align serverless product with tracer to create streamlined Service Representation for Serverless

Key Changes:

- Change service name to match instance name for all managed services (aws.lambda -> lambda name, etc) (breaking)

- Add `span.kind:server` on synthetic spans made via span-inferrer

### Motivation

Improve Service Map for Serverless. This allows for synthetic spans to have their own service on the map which connects with the inferred spans from the tracer side.

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -89,11 +89,11 @@ impl Processor {
         aws_config: Arc<AwsConfig>,
         metrics_aggregator: Arc<Mutex<MetricsAggregator>>,
     ) -> Self {
-        let service = config.service.clone().unwrap_or(String::from("aws.lambda"));
         let resource = tags_provider
             .get_canonical_resource_name()
             .unwrap_or(String::from("aws.lambda"));
 
+        let service = config.service.clone().unwrap_or(resource.clone());
         let propagator = DatadogCompositePropagator::new(Arc::clone(&config));
 
         Processor {
--- a/bottlecap/src/lifecycle/invocation/triggers/mod.rs
+++ b/bottlecap/src/lifecycle/invocation/triggers/mod.rs
@@ -113,12 +113,19 @@ pub trait Trigger: ServiceNameResolver {
     fn resolve_service_name(
         &self,
         service_mapping: &HashMap<String, String>,
-        fallback: &str,
+        instance_name: &str,
+        fallback: Option<&str>,
     ) -> String {
         service_mapping
             .get(&self.get_specific_identifier())
             .or_else(|| service_mapping.get(self.get_generic_identifier()))
-            .unwrap_or(&fallback.to_string())
+            .unwrap_or_else(|| {
+                if !instance_name.is_empty() {
+                    instance_name.to_string()
+                } else {
+                    fallback.unwrap_or_default().to_string()
+                }
+            })
             .to_string()
     }
 }
--- a/bottlecap/src/lifecycle/invocation/triggers/alb_event.rs
+++ b/bottlecap/src/lifecycle/invocation/triggers/alb_event.rs
@@ -230,7 +230,8 @@ mod tests {
         assert_eq!(
             event.resolve_service_name(
                 &specific_service_mapping,
-                &event.request_context.elb.target_group_arn
+                &event.request_context.elb.target_group_arn,
+                None
             ),
             "specific-service"
         );
@@ -242,7 +243,8 @@ mod tests {
         assert_eq!(
             event.resolve_service_name(
                 &generic_service_mapping,
-                &event.request_context.elb.target_group_arn
+                &event.request_context.elb.target_group_arn,
+                None
             ),
             "generic-service"
         );
--- a/bottlecap/src/lifecycle/invocation/triggers/api_gateway_http_event.rs
+++ b/bottlecap/src/lifecycle/invocation/triggers/api_gateway_http_event.rs
@@ -79,7 +79,7 @@ impl Trigger for APIGatewayHttpEvent {
         let start_time = (self.request_context.time_epoch as f64 * MS_TO_NS) as i64;
 
         let service_name =
-            self.resolve_service_name(service_mapping, &self.request_context.domain_name);
+            self.resolve_service_name(service_mapping, &self.request_context.domain_name, None);
 
         span.name = "aws.httpapi".to_string();
         span.service = service_name;
@@ -421,7 +421,8 @@ mod tests {
         assert_eq!(
             event.resolve_service_name(
                 &specific_service_mapping,
-                &event.request_context.domain_name
+                &event.request_context.domain_name,
+                None
             ),
             "specific-service"
         );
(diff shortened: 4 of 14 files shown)


joeyzhao2018 (approved):
LGTM

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -89,11 +89,14 @@ impl Processor {
...
     ) -> Self {
-        let service = config.service.clone().unwrap_or(String::from("aws.lambda"));
         let resource = tags_provider
             .get_canonical_resource_name()
             .unwrap_or(String::from("aws.lambda"));
 
+        let service = match std::env::var("DD_TRACE_AWS_SERVICE_REPRESENTATION_ENABLED") {
+            Ok(val) if val.eq_ignore_ascii_case("false") => String::from("aws.lambda"),
+            _ => config.service.clone().unwrap_or(resource.clone()),
+        };
```
duncanista: not a safe way to read environment variables, use the config struct for this

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -195,6 +198,9 @@ impl Processor {
 
         self.dynamic_tags
             .insert(String::from("cold_start"), cold_start.to_string());
+        self.dynamic_tags
+            .insert(String::from("span.kind"), "server".to_string());
+
```
duncanista: How is this dynamic?

bottlecap/src/lifecycle/invocation/triggers/mod.rs
```diff
@@ -113,13 +113,29 @@ pub trait Trigger: ServiceNameResolver {
...
-            .unwrap_or(&fallback.to_string())
-            .to_string()
+            .cloned()
+            .unwrap_or_else(|| {
+                if instance_name.is_empty() {
+                    fallback.to_string()
+                } else {
+                    instance_name.to_string()
+                }
+            })
```
duncanista: Not readable enough

bottlecap/tests/service_naming_test.rs
duncanista, line 1: Why are you testing here instead of the tests of every payload?

bottlecap/src/lifecycle/invocation/processor.rs
zarirhamza: Fixed moved to cold_start and invocation instances separately

bottlecap/tests/service_naming_test.rs
zarirhamza: moved to testing every service with and without it enabled

bottlecap/tests/service_naming_test.rs
```diff
@@ -0,0 +1,151 @@
...
+    let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    path.push("tests/payloads");
+    path.push(file_name);
+    fs::read_to_string(path).expect("Failed to read test payload file")
+}
+
+#[test]
+#[serial]
+fn test_dynamodb_service_name_instance_fallback() {
+    unsafe { std::env::remove_var("DD_TRACE_AWS_SERVICE_REPRESENTATION_ENABLED") };
```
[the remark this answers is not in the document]
zarirhamza: removed

bottlecap/tests/service_naming_test.rs
```diff
@@ -0,0 +1,151 @@
...
+fn with_env_var<F: FnOnce()>(f: F) {
+    // Helper to set env var to "false" and restore original value after running closure
+    const VAR: &str = "DD_TRACE_AWS_SERVICE_REPRESENTATION_ENABLED";
+    let original = env::var(VAR).ok();
+    unsafe { env::set_var(VAR, "false") };
+    f();
+    if let Some(val) = original {
+        unsafe { env::set_var(VAR, val) };
+    } else {
+        unsafe { env::remove_var(VAR) };
```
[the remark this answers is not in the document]
zarirhamza: removed

bottlecap/src/lifecycle/invocation/triggers/mod.rs
```diff
@@ -96,14 +96,32 @@ pub fn parameterize_api_resource(resource: String) -> String {
...
+    fallback: &str,
+    aws_service_representation_enabled: bool,
+) -> String {
+    if !aws_service_representation_enabled {
+        return fallback.to_string();
+    }
+
+    instance_name.to_string()
+}
+
```
duncanista, line 111: Does this have to be a method? Can't it just be inside of the `resolve_service_name`? Don't see the method being used anywhere else

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -89,16 +91,23 @@ impl Processor {
         aws_config: Arc<AwsConfig>,
         metrics_aggregator: Arc<Mutex<MetricsAggregator>>,
     ) -> Self {
-        let service = config.service.clone().unwrap_or(String::from("aws.lambda"));
         let resource = tags_provider
             .get_canonical_resource_name()
             .unwrap_or(String::from("aws.lambda"));
 
+        let service = get_default_service_name(
```
duncanista, line 98: Ah, I see it being used here

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -238,6 +248,9 @@ impl Processor {
         );
         cold_start_span.span_id = generate_span_id();
         cold_start_span.start = start_time;
+        cold_start_span
+            .meta
+            .insert("span.kind".to_string(), "server".to_string());
```
duncanista: Can we add this into the span generation method? There should be a method which generates the cold start span and the invocation span, maybe you can just add it there, as we add some default over there too

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -398,6 +411,11 @@ impl Processor {
             .meta
             .extend(self.dynamic_tags.clone());
 
+        context
+            .invocation_span
+            .meta
+            .insert("span.kind".to_string(), "server".to_string());
```
duncanista: Which then we can remove this

bottlecap/src/lifecycle/invocation/processor.rs
duncanista: And maybe also add a comment when we fix this to WHY we are adding this

bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -282,6 +341,7 @@ impl SpanInferrer {
                 String::from("peer.service"),
                 invocation_span.service.clone(),
             );
+            s.meta.insert("span.kind".to_string(), "server".to_string());
```
duncanista, line 344: I think this would also be solved in the same comment as before?

duncanista (approved):
LGTM – just left a comment!

PR #761, opened 2025-07-28
feat: App & API Protection processor

RomainMuller:
The AAP processor observes incoming requests and responses, evaluates the configured ruleset against those, and collects findings in a Context object.
The last 5 requests' context objects are retained in a buffer, and they can be used to annotate traces by acting as a span processor.


bottlecap/src/appsec/processor/default-recommended-ruleset.json.gz
duncanista, line 1: Nice!

bottlecap/src/config/yaml.rs
```diff
@@ -91,6 +93,13 @@ pub struct YamlConfig {
     pub capture_lambda_payload_max_depth: Option<u32>,
     #[serde(deserialize_with = "deserialize_optional_bool_from_anything")]
     pub serverless_appsec_enabled: Option<bool>,
+    pub appsec_rules: Option<String>,
+    #[serde(default)]
```
duncanista: Do you really need the `#[serde(default)]`? The `YamlConfig` field already includes it, which should be passed down to this fields

bottlecap/src/config/env.rs
```diff
@@ -307,6 +309,22 @@ pub struct EnvConfig {
...
     #[serde(deserialize_with = "deserialize_optional_bool_from_anything")]
     pub serverless_appsec_enabled: Option<bool>,
+    /// @env `DD_APPSEC_RULES`
+    ///
+    /// The path to a user-configured App & API Protection ruleset (in JSON format).
+    pub appsec_rules: Option<String>,
+    /// @env `DD_APPSEC_WAF_TIMEOUT`
+    ///
+    /// The timeout for the WAF to process a request, in microseconds.
+    #[serde(default)]
```
duncanista: Same here

bottlecap/src/appsec/processor/mod.rs
```diff
@@ -0,0 +1,381 @@
...
+            Self::WafRulesetLoadingError(diag) => {
+                write!(f, "aap: failed to load ruleset: {diag:?}")
+            }
+            Self::WafInitializationFailed => write!(
+                f,
+                "aap: failed to initialize WAF (is there any active rule or processor in the ruleset?)"
+            ),
+        }
+    }
+}
```
duncanista: Can this be simplified using `thiserror` crate?

duncanista (approved):

bottlecap/src/config/yaml.rs
RomainMuller: Ah you're right I forgot this could also be set at container level!

bottlecap/src/appsec/processor/mod.rs
RomainMuller: Ohhhh I did not know about this crate, shineyyy!

PR #765, opened 2025-07-30
feat: write trigger data extraction logic

RomainMuller:


bottlecap/src/lifecycle/invocation/triggers/mod.rs
```diff
@@ -123,6 +126,61 @@ pub trait Trigger: ServiceNameResolver {
     }
 }
 
+macro_rules! identified_triggers {
```
duncanista, line 137: Maybe let's add some docs for folks who aren't aware of how macros work

bottlecap/src/proxy/interceptor.rs
```diff
@@ -153,6 +160,18 @@ async fn invocation_next_proxy(
             }
         };
 
+    if let Some(appsec) = appsec_processor {
+        if let Some(rid) = intercepted_parts
+            .headers
+            .get("Lambda-Runtime-Aws-Request-Id")
+            .and_then(|v| v.to_str().ok())
+        {
+            if let Ok(trigger) = IdentifiedTrigger::from_slice(&intercepted_bytes) {
```
duncanista, line 169: What happens if no trigger is found, is there a chance that you still receive elements that are worth checking for ASM?

Was this the same way as the Go counterpart?

bottlecap/src/proxy/interceptor.rs
```diff
@@ -153,6 +160,18 @@ async fn invocation_next_proxy(
             }
         };
 
+    if let Some(appsec) = appsec_processor {
+        if let Some(rid) = intercepted_parts
```
duncanista:
```suggestion
        if let Some(request_id) = intercepted_parts
```

Just to keep consistency in this file

bottlecap/src/lifecycle/invocation/triggers/body.rs
```diff
@@ -0,0 +1,32 @@
...
+use bytes::{Buf, Bytes};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Default, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Body {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub body: Option<String>,
+    #[serde(default)]
+    pub is_base64_encoded: bool,
```
duncanista, line 13: Will this exist on every payload where the Body was added or should this be something that should be ignored during serialization?

bottlecap/src/lifecycle/invocation/triggers/api_gateway_http_event.rs
```diff
@@ -13,37 +13,51 @@ use std::collections::HashMap;
...
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    pub query_string_parameters: HashMap<String, String>,
+    #[serde(flatten)]
+    pub body: Body,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
 pub struct RequestContext {
+    #[serde(default)]
```
duncanista, line 37: Confused on why so many defaults here, but only in specific fields, could you elaborate?

bottlecap/src/lifecycle/invocation/triggers/serde_utils.rs
```diff
@@ -0,0 +1,17 @@
...
+    'de,
+    K: Deserialize<'de> + Eq + Hash,
+    V: Deserialize<'de>,
+    D: Deserializer<'de>,
+>(
+    deserializer: D,
+) -> Result<HashMap<K, V>, D::Error> {
+    let maybe = Option::deserialize(deserializer)?;
+    Ok(maybe.unwrap_or_default())
+}
```
duncanista, line 17: Doesn't a `#[serde(default)]` do this?

Not sure tbh, but sounds like a solved problem that should work OOTB

bottlecap/src/lifecycle/invocation/triggers/body.rs
```diff
@@ -0,0 +1,32 @@
+use std::io::{Cursor, Read};
+
+use base64::Engine;
+use bytes::{Buf, Bytes};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Default, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct Body {
```
duncanista, line 9: Body can be up to 6MB in AWS Lambda, seems like span inferring doesn't actually require this body in order to provide its work, is there a way this could be optional?

bottlecap/src/proxy/interceptor.rs
RomainMuller: If we don't have an identified trigger, then there is nothing for us to look at... We are interested in the HTTP payload's content, which requires we can parse the lambda input.

This is the same as is done in Go today (no identified trigger => nothing done).

bottlecap/src/appsec/processor/mod.rs
```diff
@@ -247,11 +260,375 @@ pub trait InvocationPayload {
...
+            | Self::DynamoDbRecord(_)
+            | Self::S3Record(_)
+            | Self::EventBridgeEvent(_)
+            | Self::KinesisRecord(_)
+            | Self::StepFunctionEvent(_)
+            | Self::Unknown => None,
+        }
+    }
+}
+
```
duncanista, line 631: Wondering if some of these methods should be struct methods for the events/triggers that we care about, although I guess the decision for this was that some of them won't actually have the method?

bottlecap/src/lifecycle/invocation/triggers/body.rs
RomainMuller: I've copied the `#[serde]` attributes from the https://docs.rs/aws_lambda_events/latest/aws_lambda_events crate... so I expect these "make sense" for what we do.

I'll note that the one place I've found serializing those types appears to do it on fallacious and probably incorrect reasons...

bottlecap/src/lifecycle/invocation/triggers/api_gateway_http_event.rs
RomainMuller: I've copied the `#[serde]` attributes from the [docs.rs/aws_lambda_events/latest/aws_lambda_events](https://docs.rs/aws_lambda_events/latest/aws_lambda_events) crate.

The idea being to avoid these cause deserialization failures when they should not.

All the `Option` fields are "implicitly `#[default]`", but other fields aren't, and it makes it a deserialization error to not have the field in the input payload. Not having some of these caused some tests to fail.

bottlecap/src/lifecycle/invocation/triggers/serde_utils.rs
RomainMuller: It does not -- `#[serde(default)]` makes it valid to NOT have the field at all in the document, and in that case uses the `Default` impl to get a value...

Thing is, some Lambda maps may receive `null` (explicitly), and that is "valid"... I've copied the logic here from the `aws_lambda_events` crate, too.

duncanista (commented):
Left a couple comments, but overall looks good!

bottlecap/src/lifecycle/invocation/triggers/body.rs
RomainMuller: I wanna say I can look into making the `body` field be a `Cow` on the input data; but the reality is this is likely containing JSON which means it'll need `\`-escapes decoded and will be a copy in any case...

It seems to me the only way to make the body optional (and decide on this at run time) would be to effectively duplicate all those structures... :/

Also - the 6MiB limit applies to the ENTIRE payload here, not just the `body` field in there (not that it makes a significant difference here).

bottlecap/src/proxy/interceptor.rs
duncanista: I see, I wonder what happens in the use case where a customer invokes the lambda directly using the AWS SDK, we can go through this edge case later, but AFAIK, it doesnt get identified as any trigger, yet it does receive HTTP headers and other metadata

bottlecap/src/appsec/processor/mod.rs
RomainMuller: Yeah I _could_ have implemented this trait on the trigger structs themselves; and just have had the `IdentifiedTriggers` have a `fn into_trigger(self) -> Option<Box<dyn InvocationPayload>>` instead...

It kind of made sense for me to have these implementations under the `appsec` module as they're sensitive to how appsec thinks of these concepts... but I'm not feeling all too strongly about this aspect of it.

duncanista (approved):

bottlecap/src/lifecycle/invocation/triggers/body.rs
duncanista: So what if we skip deserialization on `Body`, and then, depending on the method we use, let's say (from_slice), the body does get set? Is that even possible? Just throwing an idea

bottlecap/src/proxy/interceptor.rs
RomainMuller: Yah if it's not an AppSec-relevant payload there isn't much we can do with it...

bottlecap/src/lifecycle/invocation/triggers/body.rs
RomainMuller: The body will be processed by `serde_json` no matter what I believe... Either we get it out and "store" it, or we discard it right away.

PR #755, opened 2025-07-25
feat: port of Serverless AAP from Go to Rust

RomainMuller:
# What?

Ports the Serverless App & API Protection feature (AAP, also known as Serverless AppSec) from the Go extension to Rust.

This is using https://github.com/DataDog/libddwaf-rust to provide bindings to the in-app WAF.

This provides enhanced support for API Protection (notably, response schema collection) compared to the Go version.

Tradeoff is that XML request and response security processing is not currently supported in this version (it was in Go, but likely seldom used).

This introduces a `bottlecap::appsec::processor::Processor` that is integrated in the `bottlecap::proxy::Interceptor` (for request & response acquisition) as well as in the `bottlecap::trace_processor::TraceProcessor` (to decorate the `aws.lambda` span with security data).

# Why?

We plan on decommissioning the Go version of the agent and a tracer-side version of the Serverless AAP feature will not be available across all supported language runtimes before several weeks/months.

Also [SVLS-5286](https://datadoghq.atlassian.net/browse/SVLS-5286)

# Notes

[SVLS-5286]: https://datadoghq.atlassian.net/browse/SVLS-5286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #767, opened 2025-07-31
feat: wire in AAP span processor in the trace_processor

RomainMuller:
Annotates the spans in `TraceProcessor::process_traces` if there is a corresponding security context, and the span's name is `aws.lambda`.

If the response for the request has not been seen yet, temporarily buffers the span to wait for completion before actually sending it out.


bottlecap/src/appsec/processor/context.rs
```diff
@@ -358,14 +509,53 @@ impl Context {
         }
     }
 }
+impl Drop for Context {
+    fn drop(&mut self) {
```
duncanista, line 518: Would this mean that there's a possibility the `aws.lambda` span will be completely dropped if appsec is not seen?

bottlecap/src/proxy/interceptor.rs
```diff
@@ -167,10 +167,11 @@ async fn invocation_next_proxy(
             .and_then(|v| v.to_str().ok())
         {
             if let Ok(trigger) = IdentifiedTrigger::from_slice(&intercepted_bytes) {
+                let mut tasks = tasks.lock().await;
                 appsec
                     .lock()
                     .await
-                    .process_invocation_next(request_id, &trigger);
+                    .process_invocation_next(request_id, &trigger, &mut tasks);
```
duncanista: This might not work with how we support streaming

bottlecap/src/appsec/processor/context.rs
RomainMuller: Theoretically, yes... But all code paths that lead to dropping a `Context` from the buffer properly finalize it. This `Drop` here is additional assurance that this does not diverge in the future.

bottlecap/src/proxy/interceptor.rs
RomainMuller: Request streaming isn't really a thing in Lambda, though, is it? (Response streaming is)

bottlecap/src/appsec/processor/context.rs
```diff
@@ -24,25 +30,134 @@ pub struct Context {
...
+    /// The number of times the WAF timed out during this request.
+    waf_timed_out_occurrences: usize,
     // Whether this context has events.
     has_events: bool,
     /// Trace tags to be applied for this invocation.
     trace_tags: HashMap<TagName, TagValue>,
+    /// Whether we have had a chance to see the response body or not.
+    response_seen: bool,
+    /// Holds a trace for sending once the response is seen.
+    on_response_seen: Option<(Vec<Span>, SendingTraceProcessor, HoldArguments)>,
```
duncanista: Should this instead be the something explaining which values it holds, and then we can use the existing method of `is_pending_response` as the boolean check?

That way, when we are doing `take()` its easier to read?

bottlecap/src/appsec/processor/context.rs
```diff
@@ -24,25 +30,134 @@ pub struct Context {
...
     }
 
+    /// Returns [`true`] if the response for this request was not seen yet and
+    /// it might still be occurring.
+    pub(crate) const fn is_pending_response(&self) -> bool {
+        !self.response_seen
+    }
+
+    /// Marks the response of this request as having been seen.
+    pub(super) async fn set_response_seen(&mut self) {
```
duncanista, line 95: the name `set_response_seen` is confusing because `hold_trace` is the one who actually sets the response as seen 🤔  whereas in here we are just processing + sending data to the trace aggregator

bottlecap/src/appsec/processor/context.rs
```diff
@@ -24,25 +30,134 @@ pub struct Context {
...
             expected_response_format: ExpectedResponseFormat::default(),
             waf: waf_handle.new_context(),
             waf_timeout,
+            waf_duration: Duration::ZERO,
+            waf_timed_out_occurrences: 0,
             has_events: false,
-            trace_tags: HashMap::from([(TagName::AppsecEnabled, TagValue::Metric(1.0))]),
+            trace_tags,
+            response_seen: false,
+            on_response_seen: None,
```
duncanista: Oh yeah having `response_seen` and `on_response_seen` confused me a lot, I just saw these are two different things

bottlecap/src/appsec/processor/context.rs
```diff
@@ -24,25 +30,134 @@ pub struct Context {
...
+    /// # Panics
+    /// If called after [`Self::is_pending_response`] returns `false`.
+    pub(crate) fn hold_trace(
+        &mut self,
+        trace: Vec<Span>,
+        sender: SendingTraceProcessor,
+        args: HoldArguments,
+    ) {
+        if !self.is_pending_response() {
+            unreachable!("Context::hold_trace called after response was seen!");
```
duncanista: Maybe we don't want to panic here and instead log and return? How would this look if it panics?

Just ensuring we don't crash the Extension, which in turn would crash the runtime

bottlecap/src/appsec/processor/context.rs
duncanista: Ok, sounds good, just wanted to check because if this happens, the invocation won't appear in the customers Serverless view, as we depend on the `aws.lambda` span

bottlecap/src/appsec/processor/context.rs
```diff
@@ -358,14 +507,53 @@ impl Context {
...
+impl Drop for Context {
+    fn drop(&mut self) {
+        // First off, gentle log nudge -- we should have marked the context's response as seen before this can drop...
+        if !self.response_seen {
+            debug!(
+                "aap: Context being dropped without the response being marked as seen, this may cause traces to be dropped"
+            );
+        }
+        // In debug assertions mode, hard-crash if it means we're effectively dropping a trace.
+        debug_assert!(
```
duncanista, line 526: Can we ensure a panic won't crash the Extension, or make it idle for the next invocations?

bottlecap/src/appsec/processor/ruleset.rs
```diff
@@ -1,12 +1,13 @@
 use std::io;
 
 use bytes::{Buf, Bytes};
-use flate2::read::GzDecoder;
+use zstd::Decoder;
 
-/// The ruleset is embedded as a gzipped JSON file (it takes ~10 times less space this way).
-const DEFAULT_RECOMMENDED_RULES: &[u8] = include_bytes!("default-recommended-ruleset.json.gz");
+/// The ruleset is embedded as a ZSTD-compressed JSON file (it takes ~10 times less space this way).
```
duncanista, line 5: neat!

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -502,6 +503,17 @@ async fn extension_loop_active(
         Arc::clone(&aws_config),
         Arc::clone(&metrics_aggr),
     )));
+    // AppSec processor (if enabled)
+    let appsec_processor = match AppSecProcessor::new(config) {
+        Ok(p) => Some(Arc::new(TokioMutex::new(p))),
+        Err(AppSecFeatureDisabled) => None,
+        Err(e) => {
+            error!(
+                "PROXY | aap | error creating App & API Protection processor, the feature will be disabled: {e}"
```
duncanista:
```suggestion
                "AAP | error creating App & API Protection processor, the feature will be disabled: {e}"
```

bottlecap/src/proxy/interceptor.rs
duncanista: I meant the #771 when we are proxying and streaming requests to the Runtime Interface Client so our proxy is fast

bottlecap/src/appsec/processor/context.rs
RomainMuller: I see the confusion here... We mark the `Context` as having seen the response, which possibly allows a trace to be flushed if one was held; or will cause the trace to not be held if we haven't seen it yet.

I was thinking about possibly renaming this to something like `finalize` so it's `context.finalize()`... I'd then add some `debug_assert!` in other setters to ensure we don't try to re-use a finalized context.

bottlecap/src/appsec/processor/context.rs
RomainMuller: It's a "classic" `finalized`/`onFinalize` pair. But we can model this a little more to make it more obvious.

bottlecap/src/appsec/processor/context.rs
RomainMuller: Fair. I guess the `unreachable!` here means it's not supposed to be happening, but that can't be statically guaranteed (or else there would not be this `unreachable!`). I can make it a log unless the extension is built with `debug-assertions`.

bottlecap/src/appsec/processor/context.rs
RomainMuller: This statement is entirely omitted from builds without `debug-assertions`, which is the case for `release` profile builds by default.

I think it is helpful to have "not supposed to happen" conditions hard-crash in specially-made/testing builds so the issue is more obvious & can't easily be missed.

bottlecap/src/appsec/processor/context.rs
RomainMuller: Yes yes, I am pretty well aware of this... If everything wasn't `async`-tainted, I'd have released the trace in `impl Drop` but I can't, because I can't guarantee I have an async runtime available when `Drop` runs.

bottlecap/src/proxy/interceptor.rs
RomainMuller: How would this have to be changed then?

bottlecap/src/appsec/processor/default-recommended-ruleset.json.zst
duncanista, line 1: Size differs from the one in here https://github.com/DataDog/datadog-lambda-extension/pull/755/commits/1df2130d77ff9365e8fc9026f7df0db5deaf9b62

Are they different?

RomainMuller: Ah yes it's quite possible git resolved the conflict the wrong way... lemme make sure this is latest.

PR #768, opened 2025-07-31
fix(proxy): ensure correct headers are set when forwarding request

duncanista:
# What?

Fixes an issue where proxy would fail because a response could come as a streamed buffer

# Why?

We were not taking into account that we are collecting the body before even forwarding the request to the Lambda Runtime, so headers will sent `as-is`. Causing issues when a streamed buffer arrived to ourselves as `transfer-encoding: chunked` which will confuse the runtime since we are proxying it already.

### DIFF
--- a/bottlecap/src/proxy/interceptor.rs
+++ b/bottlecap/src/proxy/interceptor.rs
@@ -304,6 +304,12 @@ fn build_forward_response(
 
     if let Some(h) = forward_response.headers_mut() {
         *h = parts.headers;
+
+        // Since the body has been already collected, we can set the content-length header instead.
+        if h.contains_key("transfer-encoding") {
+            h.remove("transfer-encoding");
+            h.insert("content-length", body_bytes.len().to_string().parse()?);
+        }
     }
 
     let forward_response = forward_response.body(Body::from(body_bytes))?;
@@ -333,6 +339,12 @@ fn build_proxy_request(
 
     if let Some(h) = request.headers_mut() {
         *h = parts.headers.clone();
+
+        // Since the body has been already collected, we can set the content-length header instead.
+        if h.contains_key("transfer-encoding") {
+            h.remove("transfer-encoding");
+            h.insert("content-length", body_bytes.len().to_string().parse()?);
+        }
     }
 
     let request = request.body(Body::from(body_bytes))?;
--- a/bottlecap/src/proxy/mod.rs
+++ b/bottlecap/src/proxy/mod.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{env, sync::Arc};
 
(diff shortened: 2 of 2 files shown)


duncanista:
Weirdly the License check isn't passing, but even running the commands manually to update don't fix it by updating the license files.

bottlecap/src/proxy/interceptor.rs
```diff
@@ -304,6 +304,12 @@ fn build_forward_response(
 
     if let Some(h) = forward_response.headers_mut() {
         *h = parts.headers;
+
+        // Since the body has been already collected, we can set the content-length header instead.
+        if h.contains_key("transfer-encoding") {
+            h.remove("transfer-encoding");
+            h.insert("content-length", body_bytes.len().to_string().parse()?);
```
RomainMuller, line 311: Confused by the `.parse()` here - the header value is string-ish, so `.into()` might be enough (if even necessary)?

bottlecap/src/proxy/mod.rs
```diff
@@ -18,14 +18,19 @@ pub fn should_start_proxy(config: &Arc<Config>, aws_config: Arc<AwsConfig>) -> b
         .as_ref()
         .is_some_and(|s| s.eq("/opt/datadog_wrapper"));
 
-    lwa_proxy_set || (datadog_wrapper_set && config.serverless_appsec_enabled)
+    // We are not setting this as a config option because we will only allow it as an experimental feature.
+    // It is mainly expected to be used in a development environment.
+    let experimental_proxy_enabled =
+        env::var("DD_EXPERIMENTAL_ENABLE_PROXY").is_ok_and(|v| v.to_lowercase().eq("true"));
```
RomainMuller, line 24: To be clear, this enables the proxy even if there is neither AppSec nor LWA? Like it's operating in no-op mode? (That's a very good idea ... TBH I might have named the variable differently but that is ultimately your turf so if that name makes sense, then I'm cool with it, too)

bottlecap/src/proxy/mod.rs
```diff
@@ -18,14 +18,19 @@ pub fn should_start_proxy(config: &Arc<Config>, aws_config: Arc<AwsConfig>) -> b
...
+
+    lwa_proxy_set
+        || (datadog_wrapper_set && (config.serverless_appsec_enabled || experimental_proxy_enabled))
 }
 
 #[cfg(test)]
 mod tests {
-    use std::time::Instant;
-
     use super::*;
```
RomainMuller, line 28: This change is a little surprising to me as it violates the grouping that is found elsewhere?

RomainMuller (approved):

astuyve (approved):
approved but looking forward to the non-blocking variant!

PR #772, opened 2025-08-05
chore: build development layer files in CI

RomainMuller:
The layers are uploaded as job artifacts by GitHub actions, and built with the `release-with-debug-assertions` profile (optimized for size, but with `debug-assertions`), so they can easily be used for somke-testing without bothering to locally build them.


.github/workflows/rs_ci.yml
```diff
@@ -167,3 +161,34 @@ jobs:
       - run: cargo install dd-rust-license-tool --locked
       - run: dd-rust-license-tool check
         working-directory: bottlecap
+
```
duncanista, line 164: Why would we need this when we have it available in GitLab?

RomainMuller: 1. I don't like GitLab (weak, I know, but that is a reason to me)
2. I have no idea whatsoever how to get layer artifacts from GitLab (perhaps related to reason number 1)
3. I'd really like to do my smoke testing with `debug-assertions` enabled

That being said -- happy to drop this if I get informed :)

duncanista: Can definitely help you leverage GitLab to do the smoke testing there! Let's chat on it, sorry I just saw this, we can definitely add a step on the building which adds the debug-assertions!

PR #776, opened 2025-08-05
feat: For OOM metric, also use error type in PlatformRuntimeDone

lym953:
# Background
To compute the `aws.lambda.enhanced.out_of_memory` metric, right now we only rely on runtime-specific errors such as `Runtime exited with error: signal: killed` for Node.

# Problem
A customer reported that their Lambda OOMed but the metric was not logged.
After testing, I found that the existing approach only works for:
- .NET
- Node
- Java

but doesn't work for:
- Go
- Ruby
- Python

(The result is only based on my experiments. There could likely be other cases I didn't cover.)

# This PR
In addition to the current approach, also increment the metric if the `PlatformRuntimeDone` event contains `{status: 'Status::Error', error_type: 'Runtime.OutOfMemory'`

# Test
Summary: Ruby 3.3 and Python 3.13 start working.

**.NET 6**: still works
image

**Node 22**: still works
image

**Java 11**: still mostly works
image

**Ruby 3.3**: works (new)
image

**Python 3.13**: mostly works (new)
image

**Go 1**: still doesn't work. Will address it in future PRs.
image

# Notes
1. More details:
https://datadoghq.atlassian.net/wiki/spaces/SLS/pages/5371986568/out_of_memory+metric

2. Like `PlatformRuntimeDone` event, `PlatformReport` event can also contain `{status: 'Status::Error', error_type: 'Runtime.OutOfMemory'`. Why not use that event?
If the extension shuts down due to OOM, the `PlatformReport` event will be processed at next invocation when the extension restarts. However, if there is no next invocation, then `PlatformReport` event will never be processed, so we will miss this OOM. Therefore, `PlatformRuntimeDone` is better than `PlatformReport`. (Though it can sometimes also miss an OOM if the extension hard crashes.)

Jira: https://datadoghq.atlassian.net/browse/SVLS-7319

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -872,11 +872,10 @@ async fn handle_event_bus_event(
                 TelemetryRecord::PlatformReport {
                     ref request_id,
                     metrics,
-                    ref error_type,
                     ..
                 } => {
                     let mut p = invocation_processor.lock().await;
-                    p.on_platform_report(request_id, metrics, event.time.timestamp(), error_type.clone());
+                    p.on_platform_report(request_id, metrics, event.time.timestamp());
                     drop(p);
                     return Some(event);
                 }
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -302,8 +302,10 @@ impl Processor {
                 self.enhanced_metrics.increment_timeout_metric(timestamp);
             }
 
-            if status == Status::Error && error_type == Some("Runtime.OutOfMemory".to_string())  {
-                debug!("Invocation Processor | PlatformRuntimeDone | Got Runtime.OutOfMemory. Incrementing OOM metric.");
+            if status == Status::Error && error_type == Some("Runtime.OutOfMemory".to_string()) {
+                debug!(
+                    "Invocation Processor | PlatformRuntimeDone | Got Runtime.OutOfMemory. Incrementing OOM metric."
+                );
                 self.enhanced_metrics.increment_oom_metric(timestamp);
             }
         }
@@ -552,7 +554,6 @@ impl Processor {
         request_id: &String,
         metrics: ReportMetrics,
         timestamp: i64,
-        error_type: Option<String>,
     ) {
         // Set the report log metrics
         self.enhanced_metrics


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -851,13 +851,15 @@ async fn handle_event_bus_event(
                     ref request_id,
                     metrics: Some(metrics),
                     status,
+                    ref error_type,
```
duncanista, line 854: Does it need to be a ref? can't you just get the whole thing and then not clone it? doesn't seem like we need to clone if it's only used by the processor

bottlecap/src/logs/lambda/processor.rs
duncanista, line 1: Looks like we would end up with a double count if both happen? Is that possible?

bottlecap/src/bin/bottlecap/main.rs
lym953: Line 870: `return Some(event);`. Without `ref` I'd get an error:
```
use of partially moved value: `event`
partial move occurs because value has type `std::option::Option<std::string::String>`, which does not implement the `Copy` trait

main.rs(854, 21): value partially moved here
main.rs(854, 21): borrow this binding in the pattern to avoid moving the value: `ref `
```
Do you have a better solution?

bottlecap/src/logs/lambda/processor.rs
lym953: At least it didn't happen from the 6 Lambdas I deployed, but I'm sure about other scenarios I didn't cover. I'd ship it then let customers report double counts if any, so we can investigate further.

I think shipping the fix is better than not doing it because over counting OOMs (if any) should be better than under counting. For example, the customer in the ticket got a 0 for the metric (and the metric didn't even show up on the dropdown list on the UI), which may make them miss this important error.

bottlecap/src/bin/bottlecap/main.rs
duncanista: Nope, actually this makes sense, thanks for explaining it!

bottlecap/src/logs/lambda/processor.rs
duncanista: Sounds good, we'll see if this actually overcounts in self monitoring

duncanista (approved):

bottlecap/src/logs/lambda/processor.rs
lym953: Actually I'm already using self monitoring. But we can observe longer later.

PR #778, opened 2025-08-06
fix: Fix parsing provided.al runtime by reading /etc/os-release

lym953:
# Problem
We fail to parse the runtime when it is `provided.al2` or `provided.al2023`. I don't though what's the impact though, because in Datadog logs, the runtime tag is shown correctly even if I removed all the env vars related to runtime.
image

# This PR
Fix the problem by changing the path from `/etc` to `/etc/os-release`.

Thanks @shreyamalpani for the idea about the fix.

# Test
## Before
Failed to parse the runtime
image
## After
Works for both al2 and al2023
image

image

### DIFF
--- a/bottlecap/src/tags/lambda/tags.rs
+++ b/bottlecap/src/tags/lambda/tags.rs
@@ -4,7 +4,7 @@ use std::env::consts::ARCH;
 use std::fs;
 use std::sync::Arc;
 use std::time::Instant;
-use tracing::{debug, error};
+use tracing::{debug, warn};
 
 // Environment variables for the Lambda execution environment info
 const QUALIFIER_ENV_VAR: &str = "AWS_LAMBDA_FUNCTION_VERSION";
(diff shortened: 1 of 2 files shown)


bottlecap/src/tags/lambda/tags.rs
```diff
@@ -192,7 +192,10 @@ pub fn resolve_runtime_from_proc(proc_path: &str, fallback_provided_al_path: &st
                     },
                 )
         })
-        .unwrap_or_else(|| "unknown".to_string());
+        .unwrap_or_else(|| {
+            error!("Failed to read os-release file or extract runtime");
```
shreyamalpani: Maybe this should be a debug log?

lym953: This usually shouldn't happen, so I think `debug` level is too low to attract our attention to it. If you think `error` is too high due to the low impact, how about `warn`?

shreyamalpani: True, I think `warn` makes sense

lym953: Changed to `warn`.

shreyamalpani (approved):

PR #790, opened 2025-08-19
chore: Update error message when fail to send log

lym953:
## Problem
When logs fail to be sent to Datadog, the extension prints an error:
> No more retries as the access was rejected

which I think can be more clear.

Pasted Graphic

## This PR
Change the error message to:
> Failed to send logs to datadog because access was rejected. Is the API key invalid?

### DIFF
--- a/bottlecap/src/logs/flusher.rs
+++ b/bottlecap/src/logs/flusher.rs
@@ -129,7 +129,9 @@ impl Flusher {
                     _ = resp.text().await;
                     if status == StatusCode::FORBIDDEN {
                         // Access denied. Stop retrying.
-                        error!("Failed to send logs to datadog because access was rejected. Is the API key invalid?");
+                        error!(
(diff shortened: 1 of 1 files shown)


bottlecap/src/logs/flusher.rs
```diff
@@ -128,7 +128,8 @@ impl Flusher {
                     let status = resp.status();
                     _ = resp.text().await;
                     if status == StatusCode::FORBIDDEN {
-                        error!("No more retries as the access was rejected");
+                        // Access denied. Stop retrying.
+                        error!("Failed to send logs to datadog because access was rejected. Is the API key invalid?");
```
litianningdatadog: “Failed to send logs to Datadog: Access denied. Please verify that your API key is valid.”

litianningdatadog (approved):

PR #792, opened 2025-08-20
feat: Early load ustr

lym953:
## Problem
The first time `Ustr::from()` is called, it takes 10+ ms to init. Right now this is on the critical path and adds to cold start time.

## This PR
Start it early in a separate thread, so when it is needed later, it will be already initialized. 

## Result
Using `extensiononly` stack on Node 22:

### Before
- `Datadog Next-Gen Extension ready in`: **54.1** ±2.2 ms (95% confidence interval, 21 data points)
- `Metrics aggregator created in`: **10+ ms** (10,000+ us)

### After
- `Datadog Next-Gen Extension ready in`: **49.7**±1.9 ms (95% confidence level, 10 data points)
- `Metrics aggregator created in`: **277±186 us** (95% confidence level, 10 data points)

Thanks @duncanista for identifying this performance bottleneck!
Jira: https://datadoghq.atlassian.net/browse/SVLS-7456

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -498,10 +498,10 @@ async fn extension_loop_active(
         .expect("failed to create aggregator"),
     ));
     debug!(
-        "Metrics aggregator created in {:} ms",
+        "Metrics aggregator created in {:} microseconds",
         metrics_aggr_init_start_time
             .elapsed()
-            .as_millis()
+            .as_micros()
             .to_string()
     );
 


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -479,13 +489,21 @@ async fn extension_loop_active(
...
             SortedTags::parse(&tags_provider.get_tags_string()).unwrap_or(EMPTY_TAGS),
             CONTEXTS,
         )
         .expect("failed to create aggregator"),
     ));
+    debug!(
+        "Metrics aggregator created in {:} ms",
+        metrics_aggr_init_start_time
+            .elapsed()
+            .as_millis()
```
duncanista: Could this be as nanos? since it normally is below a ms?

duncanista (approved):
Amazing work Yiming 🔥 :shipit:

bottlecap/src/bin/bottlecap/main.rs
lym953: image
Good point. Changed to microseconds (1e-6 sec)

PR #794, opened 2025-08-21
Ignoring Unwanted Resources in APM

jchrostek-dd:
## Task
https://datadoghq.atlassian.net/browse/SVLS-6846

## Overview
We want to allow users to set filter tags which drops traces with root spans that match specified span tags. Specifically, users can set `DD_APM_FILTER_TAGS_REQUIRE` or `DD_APM_FILTER_TAGS_REJECT`.

More info [here](https://docs.datadoghq.com/tracing/guide/ignoring_apm_resources/?tab=datadogyaml#trace-agent-configuration-options).

## Testing
Deployed changes to Lambda. Invoked Lambda directly and through API Gateway to check with different root spans. Set the tags to either be REQUIRE or REJECT with value `name:aws.lambda`. Confirmed in logs and UI that we were dropping spans.


bottlecap/src/traces/trace_processor.rs
```diff
@@ -21,8 +21,10 @@ use datadog_trace_utils::trace_utils::{self};
 use datadog_trace_utils::tracer_header_tags;
 use datadog_trace_utils::tracer_payload::{TraceChunkProcessor, TracerPayloadCollection};
 use ddcommon::Endpoint;
+use regex::Regex;
 use std::str::FromStr;
 use std::sync::Arc;
+use tracing::debug;
 use tracing::error;
```
duncanista:
```suggestion
use tracing::{debug, error};
```

bottlecap/src/traces/trace_processor.rs
```diff
@@ -72,6 +81,176 @@ impl TraceChunkProcessor for ChunkProcessor {
...
+            if matches_reject_regex {
+                debug!(
+                    "Filtering out span '{}' - matches reject regex tags {}",
+                    span.name,
+                    reject_regex_tags.join(", ")
+                );
+                return true;
+            }
+        }
+    }
```
duncanista, line 156: Code looks quite repetitive, is there a way to simplify this?

bottlecap/src/traces/trace_processor.rs
```diff
@@ -72,6 +81,176 @@ impl TraceChunkProcessor for ChunkProcessor {
...
+        let key = parts[0].trim();
+        let value = parts[1].trim();
+        span_matches_tag_exact(span, key, value)
+    } else if parts.len() == 1 {
+        let key = parts[0].trim();
+        span_has_key(span, key)
+    } else {
+        false
+    }
+}
```
duncanista, line 174: Would a match be better here?
1 => ...
2 => ...
_ => false

bottlecap/src/traces/trace_processor.rs
```diff
@@ -72,6 +81,176 @@ impl TraceChunkProcessor for ChunkProcessor {
...
+        let key = parts[0].trim();
+        let value = parts[1].trim();
+        span_matches_tag_regex(span, key, value)
+    } else if parts.len() == 1 {
+        let key = parts[0].trim();
+        span_has_key(span, key)
+    } else {
+        false
+    }
+}
```
duncanista, line 189: Code looks very similar, cant we just pass the filter method being used?

bottlecap/src/traces/trace_processor.rs
```diff
@@ -72,6 +81,176 @@ impl TraceChunkProcessor for ChunkProcessor {
...
+    if let Some(prop_value) = span_property_value {
+        if prop_value == value {
+            return true;
+        }
+    }
+
+    false
+}
+
+fn span_matches_tag_regex(span: &Span, key: &str, value: &str) -> bool {
```
duncanista, line 215: Same for this one

bottlecap/src/traces/trace_processor.rs
jchrostek-dd: We could reduce the number of functions by passing in parameters to control for behavior like reject/require and match exact/regex. But I think that makes the code less readable, and doesn't necessarily simplify things. Or did you have something else in mind?

duncanista: Whichever makes it readable and easy to follow is my go-to always, up to you however you feel like this should be handled

jchrostek-dd: In that case, I'll think I'll leave it how it is. Same for the other places with some overlap in code =)

duncanista (approved):

PR #795, opened 2025-08-21
chore(gitlab): remove Go agent from build

duncanista:
# What?

Removes Go Agent from image and layer – therefore, the step on build pipeline is gone.

# Why?

Dropping Go Agent as we've reached feature parity

# Notes

Not deleting scripts, as some teams depend on it.

### DIFF
--- a/images/Dockerfile.build_layer
+++ b/images/Dockerfile.build_layer
@@ -24,9 +24,6 @@ WORKDIR /extensions
 # Copy Rust Agent binary
 COPY .binaries/bottlecap-$FILE_SUFFIX /extensions/datadog-agent
 
-# UPX compress
-RUN upx -1 /extensions/datadog-agent
-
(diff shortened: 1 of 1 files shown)


images/Dockerfile.build_layer
```diff
@@ -19,24 +18,21 @@ RUN ARCH=$(uname -m) && \
...
-# UPX compress
-RUN upx -1 /datadog-agent-go
-
 RUN mkdir /extensions
 WORKDIR /extensions
 
 # Copy Rust Agent binary
 COPY .binaries/bottlecap-$FILE_SUFFIX /extensions/datadog-agent
 
+# UPX compress
```
duncanista: Should we run UPX compress on Rust binary?

duncanista: Removed as this should be taken care of in cargo toml

PR #796, opened 2025-08-22
fix(lifecycle): spawn tasks on listener

duncanista:
# What?

Make Lifecycle Listener for Universal Instrumentation to spawn tasks to process data

# Motivation

There is no need to await for it to finish in order to continue, also [SVLS-7464](https://datadoghq.atlassian.net/browse/SVLS-7464)

# Notes

Still trying to figure out how to handle invocation end

[SVLS-7464]: https://datadoghq.atlassian.net/browse/SVLS-7464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/lifecycle/listener.rs
+++ b/bottlecap/src/lifecycle/listener.rs
@@ -83,7 +83,7 @@ impl Listener {
 
         let mut tasks = tasks.lock().await;
         while let Some(task) = tasks.join_next().await {
-            if let Some(e) = task.err() {
+            if let Err(e) = task {
                 error!("Lifecycle API | Shutdown error: {e}");
             }
         }
--- a/bottlecap/src/lifecycle/listener.rs
+++ b/bottlecap/src/lifecycle/listener.rs
@@ -13,8 +13,9 @@ use serde_json::json;
 use std::collections::HashMap;
 use std::net::SocketAddr;
 use std::sync::Arc;
-use tokio::net::TcpListener;
 use tokio::sync::Mutex;
+use tokio::{net::TcpListener, task::JoinSet};
+use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, warn};
 
 use crate::{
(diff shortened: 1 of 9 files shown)


bottlecap/src/lifecycle/listener.rs
```diff
@@ -33,33 +34,64 @@ const AGENT_PORT: usize = 8124;
...
+        self.shutdown_token.clone()
+    }
+
     pub async fn start(&self) -> Result<(), Box<dyn std::error::Error>> {
         let port = u16::try_from(AGENT_PORT).expect("AGENT_PORT is too large");
         let addr = SocketAddr::from(([127, 0, 0, 1], port));
         let listener = TcpListener::bind(&addr).await?;
 
-        let router = self.make_router();
+        let tasks = Arc::new(Mutex::new(JoinSet::new()));
```
astuyve, line 76: Oh we need a mutex for the joinset because we pass it between threads in the handler method tasks?

astuyve (approved):

bottlecap/src/lifecycle/listener.rs
duncanista: Yeah correct, so multiple handlers can append tasks, but it should be pretty fast, that way the graceful shutdown can finish correctly in case of errors

bottlecap/src/lifecycle/listener.rs
```diff
@@ -81,33 +113,28 @@ impl Listener {
...
+            let (parts, body) = match extract_request_body(request).await {
+                Ok(r) => r,
+                Err(e) => {
+                    error!("Failed to extract request body: {e}");
+                    return;
+                }
+            };
+
+            if let Err(e) =
+                Self::universal_instrumentation_end(&parts.headers, body, invocation_processor)
```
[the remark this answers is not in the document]
duncanista: This is fine because we don't care about the result

PR #800, opened 2025-08-25
  feat: eat: Add hierarchical configurable compression levels

litianningdatadog:
feat: Add hierarchical configurable compression levels

  - Add global compression_level config parameter (0-9, default: 6) with fallback hierarchy
  - Support 2-level compression configuration: global level first, then module-specific
  - This makes configuration more convenient - set once globally or override per module
  - Apply compression configuration to metrics flushers and trace processor
  - Add environment variable DD_COMPRESSION_LEVEL for global setting

Test
- Configuration:
image
- ([log](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fltn1-fullinstrument-bn-cold-python310-lambda/log-events/2025$252F08$252F25$252F$255B$2524LATEST$255D9c19719435bc48839f6f005d2b58b552))
Configuration:
image

### DIFF
--- a/bottlecap/Cargo.toml
+++ b/bottlecap/Cargo.toml
@@ -62,8 +62,8 @@ datadog-trace-protobuf = { git = "https://github.com/DataDog/libdatadog", rev = 
 datadog-trace-utils = { git = "https://github.com/DataDog/libdatadog", rev = "9405db9cb4ef733f3954c3ee77ce71a502e98e50" , features = ["mini_agent"] }
 datadog-trace-normalization = { git = "https://github.com/DataDog/libdatadog",  rev = "9405db9cb4ef733f3954c3ee77ce71a502e98e50" }
 datadog-trace-obfuscation = { git = "https://github.com/DataDog/libdatadog", rev = "9405db9cb4ef733f3954c3ee77ce71a502e98e50"  }
-dogstatsd = { git = "https://github.com/DataDog/serverless-components", rev = "76727c3289d65953fe9d99bd955429ee4ab60d9f", default-features = false }
-datadog-fips = { git = "https://github.com/DataDog/serverless-components", rev = "76727c3289d65953fe9d99bd955429ee4ab60d9f", default-features = false }
+dogstatsd = { git = "https://github.com/DataDog/serverless-components", rev = "b0b0cb9310d8d8f2038c00a46e3267e21dc3e287", default-features = false }
+datadog-fips = { git = "https://github.com/DataDog/serverless-components", rev = "fa1d2f4ea2c4c2596144a1f362935e56cf0cb3c7", default-features = false }
 libddwaf = { version = "1.26.0", git = "https://github.com/DataDog/libddwaf-rust", rev = "1d57bf0ca49782723e556ba327ee7f378978aaa7", default-features = false, features = ["serde", "dynamic"] }
 
 [dev-dependencies]
--- a/bottlecap/src/config/env.rs
+++ b/bottlecap/src/config/env.rs
@@ -114,13 +114,6 @@ pub struct EnvConfig {
     /// the individual component doesn't specify its own.
     pub compression_level: Option<i32>,
 
-    /// @env `DD_COMPRESSION_LEVEL`
-    ///
-    /// Global level `compression_level` parameter accepts values from 0 (no compression)
-    /// to 9 (maximum compression but higher resource usage). This value is effective only if
-    /// the individual component doesn't specify its own.
-    pub compression_level: Option<i32>,
-
     // Logs
     /// @env `DD_LOGS_CONFIG_LOGS_DD_URL`
     ///
@@ -402,8 +395,6 @@ fn merge_config(config: &mut Config, env_config: &EnvConfig) {
     merge_option!(config, env_config, version);
     merge_hashmap!(config, env_config, tags);
 
-    merge_option_to_value!(config, env_config, compression_level);
-
     // Proxy
     merge_option!(config, env_config, proxy_https);
     merge_vec!(config, env_config, proxy_no_proxy);
@@ -413,6 +404,8 @@ fn merge_config(config: &mut Config, env_config: &EnvConfig) {
     merge_string!(config, env_config, dd_url);
     merge_string!(config, env_config, url);
     merge_hashmap!(config, env_config, additional_endpoints);
+
+    merge_option_to_value!(config, env_config, compression_level);
 
     // Logs
     merge_string!(config, env_config, logs_config_logs_dd_url);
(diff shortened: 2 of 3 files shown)


bottlecap/Cargo.toml
```diff
@@ -55,8 +55,8 @@ datadog-trace-protobuf = { git = "https://github.com/DataDog/libdatadog", rev =
 datadog-trace-utils = { git = "https://github.com/DataDog/libdatadog", rev = "9405db9cb4ef733f3954c3ee77ce71a502e98e50" , features = ["mini_agent"] }
 datadog-trace-normalization = { git = "https://github.com/DataDog/libdatadog",  rev = "9405db9cb4ef733f3954c3ee77ce71a502e98e50" }
 datadog-trace-obfuscation = { git = "https://github.com/DataDog/libdatadog", rev = "9405db9cb4ef733f3954c3ee77ce71a502e98e50"  }
-dogstatsd = { git = "https://github.com/DataDog/serverless-components", rev = "d131de8419c191ce21c91bb30b5915c4d8a2cc5a", default-features = false }
-datadog-fips = { git = "https://github.com/DataDog/serverless-components", rev = "d131de8419c191ce21c91bb30b5915c4d8a2cc5a", default-features = false }
+dogstatsd = { git = "https://github.com/DataDog/serverless-components", rev = "62e7207e1b588126ca0106dc91080a4ead080f3d", default-features = false }
```
litianningdatadog, line 58: Pending on the main branch to adapt the change triggered by https://github.com/DataDog/serverless-components/pull/32/

bottlecap/src/config/env.rs
```diff
@@ -105,6 +105,13 @@ pub struct EnvConfig {
     #[serde(deserialize_with = "deserialize_key_value_pairs")]
     pub tags: HashMap<String, String>,
 
+    /// @env `DD_COMPRESSION_LEVEL`
+    ///
+    /// Global level `compression_level` parameter accepts values from 0 (no compression)
```
jchrostek-dd, line 112: We aren't enforcing this are we? I don't see anything to validate that the env variable is set in this range?

bottlecap/src/config/env.rs
```diff
@@ -105,6 +105,13 @@ pub struct EnvConfig {
...
+    /// to 9 (maximum compression but higher resource usage). Only takes effect if
+    /// `use_compression` is set to `true`.
```
jchrostek-dd: Why are we requiring 'use_compression' if the user specifies 'compression_level'?

jchrostek-dd:
Do we have a fast-follow task to update our public docs for this?

bottlecap/src/config/env.rs
litianningdatadog: This serves as an optional global setup so we don't enforce its presence.

bottlecap/src/config/env.rs
litianningdatadog: Removed this confusing and incorrect comment line.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -340,6 +340,7 @@ async fn main() -> Result<()> {
     let start_time = Instant::now();
     init_ustr();
     let (aws_config, aws_credentials, config) = load_configs(start_time);
+    println!("Using config: {config:?}");
```
shreyamalpani:
```suggestion
```

shreyamalpani (commented):
Need to check for/use global compression config, if the module-specific config is not set in flushers (edit: its being set in config, all good)

bottlecap/src/config/yaml.rs
```diff
@@ -401,6 +416,12 @@ fn merge_config(config: &mut Config, yaml_config: &YamlConfig) {
         yaml_config.logs_config,
         use_compression
     );
+    merge_option_to_value!(
+        config,
+        logs_config_compression_level,
+        yaml_config,
+        compression_level
+    );
```
shreyamalpani, line 424: I'm not sure if this one is needed since we are merging from `yaml_config.logs_config` above

bottlecap/src/config/yaml.rs
```diff
@@ -414,6 +435,20 @@ fn merge_config(config: &mut Config, yaml_config: &YamlConfig) {
         additional_endpoints
     );
 
+    merge_option_to_value!(
+        config,
+        metrics_config_compression_level,
+        yaml_config,
+        compression_level
+    );
```
shreyamalpani, line 443: I think this one might not be needed as well

bottlecap/src/config/yaml.rs
```diff
@@ -423,6 +458,12 @@ fn merge_config(config: &mut Config, yaml_config: &YamlConfig) {
         yaml_config.apm_config,
         replace_tags
     );
+    merge_option_to_value!(
+        config,
+        apm_config_compression_level,
+        yaml_config,
+        compression_level
+    );
```
shreyamalpani, line 466: Same here

bottlecap/src/config/env.rs
shreyamalpani: This was a config in the (old) Go agent, but now that we have a global compression config this could be removed?

bottlecap/src/config/yaml.rs
litianningdatadog: it is used when yaml defines compression_level but not logs_config.compression_level.

bottlecap/src/config/yaml.rs
litianningdatadog: this is for the same reason as above. Basically it allows to fall back on the newly introduced compression_level in case the individual one is not defined.

bottlecap/src/config/yaml.rs
litianningdatadog: Same as above.

bottlecap/src/config/yaml.rs
shreyamalpani: Oh I see, thank you

shreyamalpani (approved):

bottlecap/src/config/env.rs
```diff
@@ -229,6 +235,12 @@ pub struct EnvConfig {
     #[serde(deserialize_with = "deserialize_optional_bool_from_anything")]
     pub trace_propagation_http_baggage_enabled: Option<bool>,
 
+    /// @env `DD_METRICS_CONFIG_COMPRESSION_LEVEL`
```
duncanista, line 238: This env var doesn't exist, I cannot see it anywhere in our docs

bottlecap/src/config/env.rs
```diff
@@ -107,6 +107,12 @@ pub struct EnvConfig {
     /// @env `DD_TAGS`
     #[serde(deserialize_with = "deserialize_key_value_pairs")]
     pub tags: HashMap<String, String>,
+    /// @env `DD_COMPRESSION_LEVEL`
```
duncanista, line 110: Also couldn't find this in any docs from the Agent, seems like this only exist in the Lambda Forwarder?

PR #801, opened 2025-08-25
Aj/dsd channel aggregator

astuyve:
1. Upgrades dogstatsd-rs so that we use the channel based, lock-free aggregator
2. sets the context size to 5k contexts instead of 5, which is helpful for very very fast functions

TODO update git sha when we merge https://github.com/DataDog/serverless-components/pull/38

### DIFF
--- a/bottlecap/src/metrics/enhanced/lambda.rs
+++ b/bottlecap/src/metrics/enhanced/lambda.rs
@@ -387,54 +387,41 @@ impl Lambda {
         // Multiply by num_cores to report in terms of cores
         let cpu_total_utilization = cpu_total_utilization_decimal * num_cores;
 
-        let metric = Metric::new(
-            constants::CPU_TOTAL_UTILIZATION_PCT_METRIC.into(),
-            MetricValue::distribution(cpu_total_utilization_pct),
-            tags.clone(),
-            Some(now),
-        );
-        if let Err(e) = aggr.insert_batch(vec![metric]) {
-            error!("Failed to insert cpu_total_utilization_pct metric: {}", e);
-        }
+        let metrics = vec![
+            Metric::new(
+                constants::CPU_TOTAL_UTILIZATION_PCT_METRIC.into(),
+                MetricValue::distribution(cpu_total_utilization_pct),
+                tags.clone(),
+                Some(now),
+            ),
+            Metric::new(
+                constants::CPU_TOTAL_UTILIZATION_METRIC.into(),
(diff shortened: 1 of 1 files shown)


bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -134,7 +134,7 @@ impl Default for ContextBuffer {
     /// Creates a new `ContextBuffer` with a default capacity of 5.
```
duncanista:
```suggestion
    /// Creates a new `ContextBuffer` with a default capacity of 5000.
```

bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -112,12 +112,7 @@ impl Lambda {
             Some(timestamp),
         );
 
-        if let Err(e) = self
-            .aggregator
-            .lock()
-            .expect("lock poisoned")
-            .insert(metric)
-        {
+        if let Err(e) = self.aggr_handle.insert_batch(vec![metric]) {
```
duncanista, line 115: Can't we add another method where its just an insert so we don't allocate a vector all the time?

duncanista (approved):
Everything looks good, I just don't like the `vec[]` allocation on every insert, wondering if we could just add another command for `Insert` w/o a batch, are even adding batches here at all?

bottlecap/src/lifecycle/invocation/context.rs
```diff
@@ -134,7 +134,7 @@ impl Default for ContextBuffer {
...
     ///
     fn default() -> Self {
```
litianningdatadog, line 138: Could we add some comment to explain why bumping up to a much bigger value?

bottlecap/src/metrics/enhanced/lambda.rs
astuyve: Yeah but a better option in this case will be to just insert them all in one call haha

bottlecap/src/lifecycle/invocation/context.rs
astuyve: yeah

bottlecap/src/metrics/enhanced/lambda.rs
astuyve: Oh but also the metric itself is heap allocated b/c it has strings and other nonconstant struct members so the vec wrapper is likely minimal actual overhead

PR #804, opened 2025-08-26
chore: unused deps

yoshi-taka:
The crate `rmp-serde` was listed as a dependency but was not used anywhere in the codebase.  
This PR removes it.

Reference: https://github.com/DataDog/datadog-lambda-extension/pull/390

Additional note:
protobuf was also removed in [#368](https://github.com/DataDog/datadog-lambda-extension/pull/368), but it still appears in Cargo.lock due to other transitive dependencies.


litianningdatadog (commented):
mind removing protobuf as well for the same reason?

duncanista:
Looks like it built! I'm actually skeptical on the protobuf change tho, I'd need to run a manual check on it, when we introduced it, the compiler would complain about protobuf not being there 👀

PR #814, opened 2025-08-28
Fixed duplicate versions of the same crate by aligning the ddsketch-a…

yoshi-taka:
This PR fixes a dependency mismatch in Cargo.toml.
Previously, ddsketch-agent and datadog-protos were pinned to different revisions, which caused Cargo to pull in multiple versions of the same crate and inflate Cargo.lock.

By aligning the ddsketch-agent revision with datadog-protos, we avoid duplicate versions and simplify the dependency tree.

If the versions were intentionally kept different, feel free to reject this PR.

### DIFF
--- a/bottlecap/Cargo.lock
+++ b/bottlecap/Cargo.lock
@@ -513,13 +513,13 @@ dependencies = [
  "chrono",
  "cookie",
  "datadog-fips",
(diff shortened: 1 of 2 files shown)


lym953 (approved):

astuyve:
Thanks!!!

PR #818, opened 2025-08-29
chore: Add CLAUDE.md to .gitignore

litianningdatadog:
Exclude Claude Code configuration file from version control as it's a local development aid file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ bottlecap/target
 bottlecap/proptest-regressions
 
 .gitlab/pipeline*
+/CLAUDE.md


.gitignore
```diff
@@ -27,3 +27,4 @@ bottlecap/target
 bottlecap/proptest-regressions
 
 .gitlab/pipeline*
+/CLAUDE.md
```
litianningdatadog, line 30: The leading slash is to explicitly ignore the file under project root dir
https://stackoverflow.com/questions/6574116/meaning-of-leading-slash-in-gitignore-file?rq=3

PR #827, opened 2025-09-10
feat: Support "hits" trace metric

lym953:
# What
This PR enable accurate counting of `trace.aws.lambda.hits` metric (hopefully).

# How
Adds a "stats concentrator", which is a simple version of [the stats concentrator in datadog-agent](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/stats/concentrator.go)

# Architecture

image

# Testing
Tested with a modified "busycaller" self monitoring stack with Python 3.11.

## After

The "hits" graph in Datadog overall aligns with the graph on AWS. The counts are:
- Datadog: [44, 171, 171, **169**, 172, 121]
- AWS: [44, 171, 171, **170**, 172, 121]

There's an undercounting of 1 invocation. To investigate later.

image
image

## Before

The "hits" graph in Datadog was very different from the graph on AWS. The timestamp for many invocations were wrong.

image

# Next steps:
1. Support other AWS Lambda trace metrics such as `trace.aws.lambda.error`
2. Maybe support trace stats for custom traces
3. Extract the code into a component for reuse


bottlecap/src/tags/lambda/tags.rs
```diff
@@ -39,7 +39,7 @@ const SERVICE_KEY: &str = "service";
 // ComputeStatsKey is the tag key indicating whether trace stats should be computed
 const COMPUTE_STATS_KEY: &str = "_dd.compute_stats";
 // ComputeStatsValue is the tag value indicating trace stats should be computed
-const COMPUTE_STATS_VALUE: &str = "1";
+const COMPUTE_STATS_VALUE: &str = "0";
```
litianningdatadog: What is the reason for this change?

bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -0,0 +1,155 @@
...
+    // force_flush: If true, flush all stats. If false, flush stats except for the few latest
+    // buckets, which may still be getting data.
+    #[must_use]
+    pub fn get_stats(&mut self, force_flush: bool) -> Vec<pb::ClientStatsPayload> {
+        let current_timestamp: u64 = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("Failed to get current timestamp")
+            .as_nanos()
+            .try_into()
+            .expect("Failed to convert timestamp to u64");
```
litianningdatadog: Can we do below to avoid potential panic?

```
      let current_timestamp = match SystemTime::now()
          .duration_since(UNIX_EPOCH)
          .and_then(|d| d.as_nanos().try_into().map_err(|_| std::io::Error::new(std::io::ErrorKind::Other, "Timestamp overflow")))
      {
          Ok(ts) => ts,
          Err(e) => {
              error!("Failed to get current timestamp: {}, skipping stats flush", e);
              return Vec::new();
          }
      };
```

bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -0,0 +1,155 @@
...
+        for (&timestamp, bucket) in &self.buckets {
+            if force_flush || Self::should_flush_bucket(current_timestamp, timestamp) {
+                stats.push(self.construct_stats_payload(timestamp, bucket));
+                to_remove.push(timestamp);
+            }
+        }
+
+        for timestamp in to_remove {
+            self.buckets.remove(&timestamp);
+        }
```
litianningdatadog: How about using retain() to save one round of scanning?
```
      self.buckets.retain(|&timestamp, bucket| {
          if force_flush || Self::should_flush_bucket(current_timestamp, timestamp) {
              stats.push(self.construct_stats_payload(timestamp, bucket));
              false // remove this bucket
          } else {
              true // keep this bucket
          }
      });
```

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -1100,7 +1119,13 @@ fn start_trace_agent(
     tokio_util::sync::CancellationToken,
 ) {
     // Stats
-    let stats_aggregator = Arc::new(TokioMutex::new(StatsAggregator::default()));
+    let stats_concentrator = Arc::new(TokioMutex::new(StatsConcentrator::new(
+        Arc::clone(config),
+        Arc::clone(tags_provider),
+    )));
+    let stats_aggregator = Arc::new(TokioMutex::new(StatsAggregator::new_with_concentrator(
```
duncanista: Is there a way where we don't need any locks and we just follow the same pattern as dogstatsd event handling?

bottlecap/src/traces/stats_agent.rs
```diff
@@ -0,0 +1,33 @@
...
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+#[derive(Clone, Copy, Default)]
+pub struct StatsEvent {
+    pub time: u64,
+}
+
+#[allow(clippy::module_name_repetitions)]
+pub struct StatsAgent {
```
duncanista, line 17: Should this own the channel it creates and then return the tx through a public method? that way it's not created in the main binary?

bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn should_flush_bucket(current_timestamp: u64, bucket_timestamp: u64) -> bool {
+        current_timestamp - bucket_timestamp >= BUCKET_DURATION_NS * NO_FLUSH_BUCKET_COUNT
+    }
+
+    fn construct_stats_payload(&self, timestamp: u64, bucket: &Bucket) -> pb::ClientStatsPayload {
+        pb::ClientStatsPayload {
+            hostname: String::new(),
+            env: self.config.env.clone().unwrap_or_default(),
+            version: self.config.version.clone().unwrap_or_default(),
+            lang: "rust".to_string(),
```
duncanista, line 135: should this be the lambda runtime lang?

bottlecap/src/traces/stats_concentrator.rs
lym953: Good point! Will change.

bottlecap/src/tags/lambda/tags.rs
lym953: `1` means computing stats on the backend when it receives traces, instead of on our extension side. This is the current approach and doesn't work well. This PR changes traces stats to be computed on our extension side.

bottlecap/src/traces/stats_concentrator.rs
lym953: Will change, though I don't it matters a lot because u64 can represent 300+ years of time.

bottlecap/src/traces/stats_concentrator.rs
lym953: @raphaelgavache Do you know what the `lang` field means? I couldn't understand it from protobuf definition: https://github.com/DataDog/datadog-agent/blob/main/pkg/proto/datadog/trace/stats.proto#L32
Reaching out to you as the author of https://github.com/DataDog/datadog-agent/pull/7875, which adds this field.

bottlecap/src/traces/stats_agent.rs
lym953: Good point. Will do.

bottlecap/src/bin/bottlecap/main.rs
lym953: Are you referring to this? https://github.com/DataDog/serverless-components/pull/32

duncanista: Correct

lym953: Sure, will implement

PR #828, opened 2025-09-12
feat: Do not trigger the race flush if an ongoing continuous flush is…

astuyve:
In rare cases a function will begin a continuous flush then become idle for a while. If and when it wakes back up, the race flush timer may trigger causing a pretty painful resource tie-up as the race flush will begin along with a new invocation as well as potentially another flush redrive/continuous flush. I noticed this can cause occasional timeouts.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -89,7 +89,10 @@ use std::{
     os::unix::process::CommandExt,
     path::Path,
     process::Command,
-    sync::Arc,
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
     time::{Duration, Instant},
 };
 use tokio::{sync::Mutex as TokioMutex, sync::RwLock, sync::mpsc::Sender, task::JoinHandle};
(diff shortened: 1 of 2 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -652,15 +661,17 @@ async fn extension_loop_active(
                 handle_next_invocation(next_response, invocation_processor.clone()).await;
         } else {
             //Periodic flush scenario, flush at top of invocation
-            if current_flush_decision == FlushDecision::Continuous {
+            if current_flush_decision == FlushDecision::Continuous
+                && !pending_flush_handles.has_pending_handles()
```
astuyve, line 665: This prevents multiple flush tasks from piling up

bottlecap/src/traces/trace_flusher.rs
```diff
@@ -114,41 +114,39 @@ impl TraceFlusher for ServerlessTraceFlusher {
...
                 let traces_clone = traces.clone();
                 let proxy_https = self.config.proxy_https.clone();
-                join_set.spawn(async move {
+                batch_tasks.push(tokio::spawn(async move {
                     Self::send(traces_clone, Some(&endpoint), &proxy_https).await
-                });
+                }));
             }
 
-            while let Some(result) = join_set.join_next().await {
```
astuyve, line 140: Previously this code would await each request serially. At the time I didn't expect that to be an issue because it was being called from another task but ultimately it does slow things down as the final flush task was driving these serially

PR #831, opened 2025-09-12
fix: Skip log message processing when logs are disabled

litianningdatadog:
# What?

Refactors telemetry client to be a method which receives parameters and subscribes accordingly

Also gets events based on logging configuration

# Why?

We needed to ensure that we don't subscribe to logs events when customer is not enabling `DD_SERVERLESS_LOGS_ENABLED`

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -17,7 +17,7 @@ use tikv_jemallocator::Jemalloc;
 static GLOBAL: Jemalloc = Jemalloc;
 
 use bottlecap::{
-    DOGSTATSD_PORT, LAMBDA_RUNTIME_SLUG, TELEMETRY_PORT,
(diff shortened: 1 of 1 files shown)


duncanista (changes requested):
By talking to AJ I realized there's a better way to do this as opposed to how we were thinking about it, we can just not subscribe to the Function and Extension events from the Telemetry API, making it a one-line change

bottlecap/src/telemetry/client.rs
```diff
@@ -1,59 +0,0 @@
...
-
-    pub async fn subscribe(&self) -> Result<Response, Box<dyn Error>> {
-        let url = base_url(TELEMETRY_SUBSCRIPTION_ROUTE, &self.runtime_api);
-        let resp = create_reqwest_client_builder()?
-            .no_proxy()
-            .build()
-            .map_err(|e| {
-                error!("Error building reqwest client: {:?}", e);
-                e
-            })?
```
duncanista, line 37: No need to create a new client every time, that's why we pass it now

PR #833, opened 2025-09-15
chore(telemetry): fix event order handling

duncanista:
# What?

Makes events from telemetry API to be sent directly to the main event bus as opposed to logs.

# Why?

Makes more sense to directly send the events first to the main event bus, there it can async handle them


duncanista:
Drafting bc AJ pointed something important, we don't want to receive all logs in the main event bus, we don't need them in reality

PR #841, opened 2025-09-17
feat: [Trace Stats] Add feature flag DD_COMPUTE_TRACE_STATS

lym953:
## This PR

Adds a feature flag `DD_COMPUTE_TRACE_STATS`.
- If true, trace stats will be computed from the extension side. In this case, we set `_dd.compute_stats` to `0`, so trace stats won't be computed on the backend.
- If false, trace stats will NOT be computed from the extension side. In this case, we set `_dd.compute_stats` to `1`, so trace stats will be computed on the backend.
- Defaults to false for now, so `_dd.compute_stats` still defaults to `1`, i.e. default behavior is not changed.
- After we fully support computing trace stats on extension side, I will change the default to true then delete the flag.

### DIFF
--- a/bottlecap/src/config/env.rs
+++ b/bottlecap/src/config/env.rs
@@ -349,10 +349,16 @@ pub struct EnvConfig {
     /// The maximum depth of the Lambda payload to capture.
     /// Default is `10`. Requires `capture_lambda_payload` to be `true`.
     pub capture_lambda_payload_max_depth: Option<u32>,
+    /// @env `DD_COMPUTE_TRACE_STATS`
+    ///
+    /// Enable computation of trace stats for AWS Lambda.
+    #[serde(deserialize_with = "deserialize_optional_bool_from_anything")]
+    pub compute_trace_stats: Option<bool>,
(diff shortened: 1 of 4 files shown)


duncanista (approved):

bottlecap/src/config/env.rs
```diff
@@ -349,10 +349,18 @@ pub struct EnvConfig {
     /// The maximum depth of the Lambda payload to capture.
     /// Default is `10`. Requires `capture_lambda_payload` to be `true`.
     pub capture_lambda_payload_max_depth: Option<u32>,
+    /// @env `DD_COMPUTE_TRACE_STATS`
```
apiarian-datadog, line 352: looks like this does not conflict with the existing `DD_TRACE_COMPUTE_STATS` and `DD_TRACE_STATS_COMPUTATION_ENABLED` which are used by some of the tracers. bit it might be worth adding `AGENT` or something into the name to make it clear that this is not a tracer feature.

lym953: Doing it in https://github.com/DataDog/datadog-lambda-extension/pull/875

PR #842, opened 2025-09-17
feat: [Trace Stats] Add skeleton of concentrator

lym953:


bottlecap/src/traces/stats_concentrator_service.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn from(err: mpsc::error::SendError<ConcentratorCommand>) -> Self {
+        StatsError::SendError(err)
+    }
+}
+
+impl From<oneshot::error::RecvError> for StatsError {
+    fn from(err: oneshot::error::RecvError) -> Self {
+        StatsError::RecvError(err)
+    }
+}
```
duncanista: You could probably use `thiserror` to avoid the boilerplate

bottlecap/src/traces/stats_concentrator_service.rs
duncanista, line 1: Wondering if we should have a submodule called `stats` for all of this stuff, instead of having it as `stats_...`

bottlecap/src/traces/stats_concentrator_service.rs
lym953: Done

bottlecap/src/traces/stats_concentrator_service.rs
lym953: I might do it in a separate PR

PR #851, opened 2025-09-18
feat: don't spawn tmp/proc tasks if one is still running

astuyve:
tmp/proc enhanced metrics tasks open and/or read files in the /proc directory using blocking I/O. These tasks can block the async context the task is called under and thus delay other important work. In rare cases, if the tasks start taking longer than the typical invocations take, these tasks can pile up and cause timeouts.

This PR ensures no other tasks are running before respawning it.

There's likely room for improvement to this logic, which is well structured but likely not ever tested with very fast (rust/go) functions. Potentially instead of relying on locks for the invocation processor or using the invocation to signal begin points, we could detect discontinuous invocations and switch to a polling mechanism.

### DIFF
--- a/bottlecap/src/metrics/enhanced/lambda.rs
+++ b/bottlecap/src/metrics/enhanced/lambda.rs
@@ -20,7 +20,10 @@ use tokio::{
 };
 use tracing::debug;
 use tracing::error;
-
+// tmp/proc enhanced metrics tasks open and/or read files in the /proc directory using blocking I/O.
+// These tasks can block the async context the task is called under and thus delay other important work.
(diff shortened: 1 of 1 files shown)


bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -18,6 +21,9 @@ use tokio::{
 use tracing::debug;
 use tracing::error;
 
+static TMP_TASK_RUNNING: AtomicBool = AtomicBool::new(false);
```
lym953, line 27: nit: Can we document why we need this (maybe copy the PR summary here), so when people try to understand the code later, they won't need to blame the code and find this PR?

astuyve: sure

PR #852, opened 2025-09-18
fix(aap): lower severity of several warnings

### LINKED ISSUE #840: v85+ introduces spammy warning from the ASM feature
Hi,

we do see errors/warnings during our updates of the lambda extensions. This happens with v85 and v86. With v84 this is still fine. It's the only thing we changed within the update.

Do we need to update the configuration somehow?

Image

RomainMuller:
These produce excessive log spam in customer accounts, for conditions they may not be able to generally address (either because the cause is outside of their control, or because they decided the trade-off was acceptable).

The alternative would be allowing granuar configuration of log level limits for each module, which would be a significantly more involved change.

This also reverts to a behavior that is closer to that of the Go version, which was silently ignoring these conditions for the most part.

Fixes #840

### DIFF
--- a/bottlecap/src/appsec/processor/context.rs
+++ b/bottlecap/src/appsec/processor/context.rs
@@ -10,7 +10,7 @@ use libddwaf::object::{Keyed, WafMap, WafObject};
 use libddwaf::{Context as WafContext, Handle, RunError, RunOutput, RunResult, waf_map};
 use mime::Mime;
 use multipart::server::Multipart;
-use tracing::{debug, warn};
+use tracing::{debug, info, warn};
 
 use crate::appsec::processor::response::ExpectedResponseFormat;
 use crate::appsec::processor::{InvocationPayload, Processor};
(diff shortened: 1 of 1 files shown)


bottlecap/src/appsec/processor/context.rs
```diff
@@ -169,9 +169,16 @@ impl Context {
         }
 
         if self.waf_timeout.is_zero() {
-            warn!(
-                "aap: WAF execution time budget for this request is exhausted, skipping WAF ruleset evaluation (consider tweaking DD_APPSEC_WAF_TIMEOUT)"
+            // Logging as INFO and not as WARN as users may find this an acceptable trade-off in
+            // situations where extending the WAF timeout is not acceptable. Issuing a WARN here
+            // would lead to excessive log spam for these customers, who then can't do anything
+            // about it.
+            info!(
```
duncanista, line 176: Would this happen on every invocation? Or just once the timeout is zero?

Trying to figure out how spammy this would still be 🤔

bottlecap/src/appsec/processor/context.rs
```diff
@@ -783,11 +789,57 @@ fn try_parse_body_with_mime(
...
+                    f,
+                    "cannot parse {mime_type} body: missing boundary parameter"
+                )
+            }
+            Self::UnsupportedMimeType(mime_type) => {
+                write!(f, "unsupported MIME type: {mime_type}")
+            }
+            Self::IOError(e) => write!(f, "failed to read body: {e}"),
+            Self::SerdeError(e) => write!(f, "failed to parse body: {e}"),
+        }
```
duncanista: Can you use `thiserror` to cut out the boilerplate?

bottlecap/src/appsec/processor/context.rs
RomainMuller: This is once the timeout reached zero, possibly multiple times per invocation (but in practice, this is at most 1 call in a given invoke -- the 1st one is guaranteed to have a >0 timeout, and we'll make at most 2 calls through this path -- once for the request, one for the response).

The spamminess comes from calls across invocations for busy functions, where this can contribute non-trivial amounts to ingested logs (≈ cost).

bottlecap/src/appsec/processor/context.rs
RomainMuller: Duh I always forget about this crate!

PR #853, opened 2025-09-18
fix: Fix deadlock when snap start and using secret for api key

lym953:
# Problem
When a Lambda (1) uses snap start, and (2) specifies Datadog API key using `DD_API_KEY_SECRET_ARN`, the extension will encounter a deadlock. For a `RwLock`, the extension first gets a read lock:
https://github.com/DataDog/datadog-lambda-extension/blob/daf633dd003447d78261e7c371838b5af21073a1/bottlecap/src/secrets/decrypt.rs#L45
then tries to get a write lock:
https://github.com/DataDog/datadog-lambda-extension/blob/daf633dd003447d78261e7c371838b5af21073a1/bottlecap/src/secrets/decrypt.rs#L65

which never finishes. This causes the function to time out.

This bug was introduced in https://github.com/DataDog/datadog-lambda-extension/pull/717.

# This PR
Fix this bug by removing the `RwLock` usage. `AwsCredential` is only created and used once in `resolve_secrets()`, and `resolve_secrets()` is only called once, so there's no need to protect this struct with a lock.

# Testing
Tested on a Lambda with:
- Python 3.13 runtime
- snap start
- using `DD_API_KEY_SECRET_ARN`

Before:
- The function timed out.
- Data failed to be sent to Datadog.

After:
- The function finished without timeout.
- Data was sent to Datadog successfully.

# Notes
Jira: https://datadoghq.atlassian.net/browse/SLES-2482

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -317,10 +317,7 @@ fn enable_logging_subsystem(config: &Arc<Config>) {
     debug!("Logging subsystem enabled");
 }
 
-fn create_api_key_factory(
-    config: &Arc<Config>,
-    aws_config: &Arc<AwsConfig>,
-) -> Arc<ApiKeyFactory> {
+fn create_api_key_factory(config: &Arc<Config>, aws_config: &Arc<AwsConfig>) -> Arc<ApiKeyFactory> {
     let config = Arc::clone(config);
     let aws_config = Arc::clone(aws_config);
 
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -23,7 +23,7 @@ use bottlecap::{
     },
     config::{
         self, Config,
-        aws::{AwsConfig, AwsCredentials, build_lambda_function_arn},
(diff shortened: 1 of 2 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -285,14 +285,13 @@ fn init_ustr() {
     });
 }
 
-fn load_configs(start_time: Instant) -> (AwsConfig, AwsCredentials, Arc<Config>) {
+fn load_configs(start_time: Instant) -> (AwsConfig, Arc<Config>) {
     // First load the AWS configuration
     let aws_config = AwsConfig::from_env(start_time);
-    let aws_credentials = AwsCredentials::from_env();
```
lym953, line 291: No need to load credentials from `main()`

bottlecap/src/secrets/decrypt.rs
```diff
@@ -42,36 +38,32 @@ pub async fn resolve_secrets(
                 }
             };
 
-            let aws_credentials_read = aws_credentials.read().await;
+            let mut aws_credentials = AwsCredentials::from_env();
```
lym953, line 41: Load it as needed in `resolve_secrets()`, which is called lazily.

PR #855, opened 2025-09-19
feat: [Trace Stats] Move stats generation after trace obfuscation

lym953:


bottlecap/src/otlp/agent.rs
```diff
@@ -163,34 +173,46 @@ impl Agent {
...
+
+        // This needs to be after process_traces() because process_traces()
+        // performs obfuscation, and we need to compute stats on the obfuscated traces.
+        if compute_trace_stats {
+            if let Err(err) = stats_sender.send(&processed_traces) {
+                error!("OTLP | Error sending traces to the stats concentrator: {err}");
+                return (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                    json!({ "message": format!("Error sending traces to the stats concentrator: {err}") }).to_string()
+                ).into_response();
```
lym953: Core change: Add a stats generation hook in OTLP agent.

bottlecap/src/traces/trace_agent.rs
```diff
@@ -523,15 +520,6 @@ impl TraceAgent {
...
 
-        if config.compute_trace_stats {
-            if let Err(err) = stats_sender.send(&traces) {
-                return error_response(
-                    StatusCode::INTERNAL_SERVER_ERROR,
-                    format!("Error sending stats to the stats aggregator: {err}"),
-                );
-            }
-        }
-
```
lym953, line 534: Moving this into `send_processed_traces()` below

bottlecap/src/traces/trace_processor.rs
```diff
@@ -445,18 +456,28 @@ impl SendingTraceProcessor {
...
+            .map_err(SendingTraceProcessorError::SendDataBuilderInfoError)?;
+
+        // This needs to be after send_processed_traces() because send_processed_traces()
+        // performs obfuscation, and we need to compute stats on the obfuscated traces.
+        if config.compute_trace_stats {
+            if let Err(err) = self.stats_sender.send(&processed_traces) {
+                error!("TRACE_PROCESSOR | Error sending traces to the stats concentrator: {err}");
+                return Err(SendingTraceProcessorError::SendStatsError(err));
+            }
+        }
```
lym953, line 469: Core change: this is moved into `send_processed_traces()` after obfuscation is done.

bottlecap/src/traces/trace_processor.rs
```diff
@@ -307,28 +308,28 @@ fn filter_span_from_lambda_library_or_runtime(span: &Span) -> bool {
 #[allow(clippy::too_many_arguments)]
 #[async_trait]
 pub trait TraceProcessor {
-    async fn process_traces(
```
lym953, line 310: This doesn't need to be async.

bottlecap/src/traces/trace_stats_processor.rs
```diff
@@ -1,34 +1,57 @@
...
-            for span in trace {
-                let stats = StatsEvent {
-                    time: span.start.try_into().unwrap_or_default(),
-                    aggregation_key: AggregationKey {},
-                    stats: Stats {},
-                };
-                self.stats_concentrator.add(stats)?;
+    pub fn send(
+        &self,
+        traces: &TracerPayloadCollection,
```
lym953, line 31: Generating stats from processed traces instead of raw traces

bottlecap/src/traces/trace_processor.rs
```diff
@@ -360,7 +361,7 @@ impl TraceProcessor for ServerlessTraceProcessor {
             test_token: None,
         };
 
-        let builder = SendDataBuilder::new(body_size, payload, header_tags, &endpoint)
+        let builder = SendDataBuilder::new(body_size, payload.clone(), header_tags, &endpoint)
```
lym953, line 364: I have to clone the processed trace payload so I can return it to generate stats. Let me know if you have a more efficient approach.

bottlecap/src/traces/trace_stats_processor.rs
```diff
@@ -1,34 +1,57 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum SendingTraceStatsProcessorError {
+    #[error("Error sending trace stats to the stats concentrator: {0}")]
+    ConcentratorCommandError(SendError<ConcentratorCommand>),
+    #[error("Unsupported trace payload version. Failed to send trace stats.")]
+    TracePayloadVersionError,
+}
+
 // Extracts information from traces related to stats and sends it to the stats concentrator
 impl SendingTraceStatsProcessor {
```
duncanista, line 23: `SendingTraceStatsProcessor` name confuses me on what it does

bottlecap/src/traces/trace_processor.rs
duncanista: Should we allow to continue even if trace stats forwarding failed? Wouldn't it be better to keep having some data being forwarded still?

bottlecap/src/otlp/agent.rs
duncanista: Same comment as for traces

bottlecap/src/traces/trace_stats_processor.rs
lym953: It's modified from `SendingTraceProcessor`. As the PR summary says, I plan to rename it to `TraceStatsGenerator`. Does this sound good to you?

bottlecap/src/traces/trace_processor.rs
lym953: I think the fundamental question is whether trace stats are as important as traces themselves. If yes, then when stats fail to be sent, we should probably return error to let the caller handle this case. Otherwise we can return ok in this case. Seems you think stats are less important than traces, so let me swallow the error then.

duncanista: Ideally, I think this would directly impede sending traces, I wouldn't expect this error to happen a lot of times, but ideally we'd like to have some data as opposed to none

I'm not sure if the tracer would send the data back if we respond with an error, is that the case?

Because this error in theory is not related to any datadog logic, but just a failed channel forwarding

lym953: > we'd like to have some data as opposed to none

What do you mean by "have some data"? Do you mean traces should still be sent to Datadog even if stats fail to send?

> I'm not sure if the tracer would send the data back if we respond with an error, is that the case?

What do you mean by "send the data back"?
I'm not sure how tracers work, but if we think this error is critical, the extension should surface it to the caller.

duncanista: > What do you mean by "have some data"? Do you mean traces should still be sent to Datadog even if stats fail to send?

Correct, before, WDYT?

> What do you mean by "send the data back"?

As in, if you reply back to the tracer with a 400/500 status, will it re-send the tracer payload we failed to process?

> I'm not sure how tracers work, but if we think this error is critical, the extension should surface it to the caller.

I agree, but in this case, it's not a processing error, it's more of a critical error on the extension, right? So we could say it's not the tracers fault, but ours(?

lym953: > As in, if you reply back to the tracer with a 400/500 status, will it re-send the tracer payload we failed to process?

I spot checked `dd-trace-py`. It doesn't retry as long as it gets a valid response, even if the status is 400/500.

I'm okay either way. I pushed a commit to swallow and log the error. Could you review it?

PR #856, opened 2025-09-19
feat: [Trace Stats] Implement stats concentrator

lym953:


bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -1,38 +1,219 @@
...
+    #[allow(clippy::cast_possible_truncation)]
+    #[allow(clippy::cast_sign_loss)]
+    fn construct_stats_payload(
+        config: &Config,
+        timestamp: u64,
+        aggregation_key: &AggregationKey,
+        stats: Stats,
+    ) -> pb::ClientStatsPayload {
+        pb::ClientStatsPayload {
+            // TODO: handle this
```
lym953, line 133: Marking many fields with TODO, so I can keep this PR small, iterate fast and handle them in future PRs.
Some of them may need code changes, while some may need just understanding work and updating the comment.

bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -1,38 +1,217 @@
 use crate::config::Config;
 use datadog_trace_protobuf::pb;
-use std::sync::Arc;
+use std::{
+    collections::HashMap,
+    sync::Arc,
+    time::{SystemTime, UNIX_EPOCH},
```
litianningdatadog: Should we switch to the `tokio::time` module to align more closely with Tokio's scheduling?

bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -1,38 +1,217 @@
...
+struct Bucket {
+    data: HashMap<AggregationKey, Stats>,
+}
 
 #[derive(Clone, Debug, Default, Copy)]
-pub struct Stats {}
+pub struct Stats {
+    pub hits: i32,
+    pub duration: i64, // in nanoseconds
+    pub error: i32,
```
litianningdatadog: Does its int type stand for an error code or error count? Can we clarify it in comment or rename it?

bottlecap/src/traces/stats_concentrator.rs
lym953: Could you elaborate?

bottlecap/src/traces/stats_concentrator.rs
lym953: Added a comment

bottlecap/src/traces/stats_concentrator.rs
lym953: For example, what's the problem if we use `std::time`?

litianningdatadog: Per https://github.com/tokio-rs/tokio/discussions/4633, it is better to use tokio's time package if tokio scheduling is involved.

bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -1,38 +1,216 @@
...
-pub struct AggregationKey {}
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct AggregationKey {
+    pub env: String,
+    pub service: String,
+    // e.g. "aws.lambda.load", "aws.lambda.import"
+    pub name: String,
+    // e.g. "my-lambda-function-name", "datadog_lambda.handler", "urllib.request"
+    pub resource: String,
+    // e.g. "aws.lambda.load", "aws.lambda.import"
```
duncanista:
```suggestion
    // e.g. "serverless"
```

not sure if the comment is correct for `r#type`

bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -1,38 +1,216 @@
...
 }
 
+// The number of latest buckets to not flush when force_flush is false.
+// For example, if we have buckets with timestamps 10, 20, 40, the current timestamp is 45,
+// and NO_FLUSH_BUCKET_COUNT is 3, then we will flush bucket 10 but not bucket 20 or 40.
+// Note that the bucket 30 is included in the 3 latest buckets even if it has no data.
+// This is to reduce the chance of flushing stats that are still being collected to save some cost.
+const NO_FLUSH_BUCKET_COUNT: u64 = 2;
+
+const S_TO_NS: u64 = 1_000_000_000;
```
duncanista, line 55: probably already exists somewhere?

lym953: It only exists in `f64`: https://github.com/DataDog/datadog-lambda-extension/blob/ee8fd6f95a62f57f11eec8ce014f23ecdac6c827/bottlecap/src/lifecycle/invocation/processor.rs#L47
I need to define another one in `u64`.

bottlecap/src/traces/stats_concentrator.rs
lym953: You are right. Will fix.

bottlecap/src/traces/stats_concentrator.rs
lym953: Talked offline. We will keep using `std::time` because `tokio::time` is mainly used for handling time intervals, but for trace stats we need to work with absolute timestamps.

PR #862, opened 2025-09-23
feat: [Trace Stats] Handle hostname and env

lym953:
## This PR
1. For `hostname` field: pass in `tags_provider` and use it to get function arn. Refer to existing code about:
    1. how to get function arn:
https://github.com/DataDog/datadog-lambda-extension/blob/3f90d2bb09b4213e3b683dc84333e48faeddba3a/bottlecap/src/logs/lambda/processor.rs#L65
    2. use function arn as hostname:
https://github.com/DataDog/datadog-lambda-extension/blob/3f90d2bb09b4213e3b683dc84333e48faeddba3a/bottlecap/src/logs/lambda/processor.rs#L259
2. Change the default value of `env` to `unknown-env`, following this RFC: https://docs.google.com/document/d/1NmGVxNyd9o8UQ7AvFKjOVq5tBDN6GN8jeQmvyBUnkrQ

## Notes

### DIFF
--- a/bottlecap/src/traces/stats_concentrator.rs
+++ b/bottlecap/src/traces/stats_concentrator.rs
@@ -55,7 +55,7 @@ pub struct StatsConcentrator {
     config: Arc<Config>,
     tracer_metadata: TracerMetadata,
     buckets: HashMap<u64, Bucket>,
-    function_arn: String,
+    hostname: String,
 }
 
 // The number of latest buckets to not flush when force_flush is false.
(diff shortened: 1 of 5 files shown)


bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -44,6 +45,7 @@ pub struct Stats {
 pub struct StatsConcentrator {
     config: Arc<Config>,
     buckets: HashMap<u64, Bucket>,
+    function_arn: String,
```
duncanista: Should this be instead `hostname` as opposed to `function_arn`? Imagine we want to use this concentrator for other services like GCP/Azure, how could this be reused?

lym953: Good point. Will do.

bottlecap/src/traces/stats_concentrator.rs
```diff
@@ -129,10 +134,10 @@ impl StatsConcentrator {
         timestamp: u64,
         aggregation_key: &AggregationKey,
         stats: Stats,
+        function_arn: &str,
     ) -> pb::ClientStatsPayload {
         pb::ClientStatsPayload {
-            // TODO: handle this
-            hostname: String::new(),
+            hostname: function_arn.to_string(),
```
litianningdatadog: nit: rename this to hostname?

lym953: Will do.

PR #863, opened 2025-09-23
feat: [Trace Stats] Handle tracer lang, tracer version and runtime_id

lym953:
## This PR
Set these fields on trace stats:
- `lang`
- `tracer_version`
- `runtime_id`

## Notes

### DIFF
--- a/bottlecap/src/traces/stats_generator.rs
+++ b/bottlecap/src/traces/stats_generator.rs
@@ -3,12 +3,13 @@ use crate::traces::stats_concentrator_service::StatsConcentratorHandle;
 use datadog_trace_utils::tracer_payload::TracerPayloadCollection;
 use tracing::error;
 
-use tokio::sync::mpsc::error::SendError;
-
 use crate::traces::stats_concentrator_service::ConcentratorCommand;
+use std::sync::atomic::{AtomicBool, Ordering};
+use tokio::sync::mpsc::error::SendError;
 
 pub struct StatsGenerator {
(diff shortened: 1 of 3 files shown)


bottlecap/src/traces/stats_generator.rs
```diff
@@ -23,12 +25,34 @@ pub enum StatsGeneratorError {
...
+        Self {
+            stats_concentrator,
+            is_tracer_metadata_set: AtomicBool::new(false),
+        }
     }
 
     pub fn send(&self, traces: &TracerPayloadCollection) -> Result<(), StatsGeneratorError> {
         if let TracerPayloadCollection::V07(traces) = traces {
             for trace in traces {
+                // Set tracer metadata only once for the first trace because
```
litianningdatadog: If i am not mistaken, this is to ensure only set meta once. If so, can we consider internalizing this check in concentror's set_tracer_metadata()?

lym953: In that case, `StatsGenerator` will need to write an unnecessary message to the channel for each trace, which can incur extra cost. I'd not do this from the performance perspective.

litianningdatadog: What I mean by that is to move the logic into StatsConcentratorHandle::set_tracer_metadata() call and won't send it if the condition is not met. No performance impact except for better code arrangement.

duncanista (approved):

PR #865, opened 2025-09-24
fix: use joinset for trace flushing

astuyve:
Goof here - I applied the same logic from main (where we want ordered sequential futures) with this change. Reverting this to a joinset allows us to await whichever future returns first, which helps in rare cases where multiple flush requests are required.

### DIFF
--- a/bottlecap/src/traces/trace_flusher.rs
+++ b/bottlecap/src/traces/trace_flusher.rs
@@ -125,13 +125,9 @@ impl TraceFlusher for ServerlessTraceFlusher {
                 .map(SendDataBuilder::build)
                 .collect();
 
-            tokio::task::yield_now().await;
             let traces_clone = traces.clone();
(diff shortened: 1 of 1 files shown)


bottlecap/src/traces/trace_flusher.rs
```diff
@@ -164,7 +161,7 @@ impl TraceFlusher for ServerlessTraceFlusher {
         if traces.is_empty() {
             return None;
         }
-        let start = std::time::Instant::now();
+        let start = tokio::time::Instant::now();
```
litianningdatadog, line 164: Do we want to migrate entirely to tokio::time, since our project already depends heavily on Tokio?

astuyve: Yeah I actually have another PR which does that, just a WIP. It's mostly just useful for tests so we can freeze/mess with time

Issue #858, opened 2025-09-23
Lambdas crash after updating to v86


MarkusSintonen:
Our image Lambdas started to crash after updating to v86 image with `Error: fork/exec /opt/extensions/datadog-agent: no such file or directory
Extension.LaunchError`

We are using the extension image as documented.

This works:
`COPY --from=public.ecr.aws/datadog/lambda-extension:84-alpine@sha256:762ae85ea92321905d2a6198ced9daa5fafe02db79ba746c9e93781865d45a56 /opt/. /opt/`

This now breaks:
`COPY --from=public.ecr.aws/datadog/lambda-extension:86-alpine@sha256:5c63f3de14e523757f8bb391ea21d61a045c2245dd0f5097d68d20fa31018b75 /opt/. /opt/`

Above was the only change after which the Lambdas started to crash on startup.

MarkusSintonen:
@astuyve 
One example for all DD env vars in one lambda:
```
DD_API_KEY_SECRET_ARN | <omitted>
DD_CAPTURE_LAMBDA_PAYLOAD | false
DD_FLUSH_TO_LOG | false
DD_INSTRUMENTATION_TELEMETRY_ENABLED | false
DD_LOGS_ENABLED | false
DD_LOGS_INJECTION | false
DD_MERGE_XRAY_TRACES | false
DD_OTLP_CONFIG_TRACES_ENABLED | false
DD_PROFILING_ENABLED | false
DD_SERVERLESS_APPSEC_ENABLED | false
DD_SERVERLESS_FLUSH_STRATEGY | end,5000
DD_SERVERLESS_LOGS_ENABLED | false
DD_SITE | datadoghq.com
DD_TRACE_ENABLED | false
DD_TRACE_OTEL_ENABLED | false
```

RomainMuller:
This is most certainly caused by the extension having un undesirable dependency on `libgcc_s.so.1`. #868 ought to remove this issue by removing the use of the `-Ctarget-feature=-crt-static` build option, which causes this dependency.

MarkusSintonen:
> you might be able to work around

I’m not in a hurry to upgrade so I can wait for the fix in the next image version, thanks!

PR #868, opened 2025-09-25
fix(bottlecap): run-time dependency on libgcc_s.so

RomainMuller:
An unexpected dependency on `libgcc_s.so` is implied by building on `musl` platforms with the `-Ctarget-feature=-crt-runtime`, which causes breakage in some customer functions. Instead of disabling the static C runtime for the entire build, use a `RUSTC_WRAPPER` script to disable it only specifically for the build-script of `libddwaf-sys` which is where `bindgen` needs to be able to dynamically load `libclang.so`.

Additionally, present the `libcompiler-rt.builtins` object to the linker where necessary because `rust` passes `-nodefaultlibs` to the linker, but some objects in `libddwaf` depend on compiler intrinsics.

Fixes: https://github.com/DataDog/datadog-lambda-extension/issues/858
JIRA: APPSEC-59319

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,17 +1,3 @@
 [alias]
 license = "dd-rust-license-tool check"
 format = "fmt --all && clippy --workspace --all-features --fix"
-
-# libddwaf's bindgen uses `clang-sys` which by default dynamically opens
-# `libclang.so`; however this is not supported on Alpine Linux (the usual musl
-# target).
-# Enabling the `static` feature of `bindgen`/`clang-sys` creates additional
-# build requirements (`libLLVM.a`), which Alpine's `llvm-static` does not
-# provide (it has its content presented as many `libLLVM*.a` files). Disabling
-# statically linking to the C runtime alleviates this issue, but can result in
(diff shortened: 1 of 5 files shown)


images/Dockerfile.bottlecap.compile
```diff
@@ -37,7 +37,7 @@ RUN --mount=type=cache,target=/usr/local/cargo/git \
     fi; \
     # The `libddwaf` crate links against static objects that require `libclang_rt.builtins`, but
     # this is not presented to the linker by default on this platform, so we force it in.
-    export RUSTFLAGS="${RUSTFLAGS:-} -Clinker=clang -L$(dirname $(clang --print-file-name="libclang_rt.builtins-$(uname -m).a")) -lclang_rt.builtins-$(uname -m)"; \
+    export RUSTFLAGS="${RUSTFLAGS:-} -Clinker=clang -Clinker-args=-l$(clang --print-file-name="libclang_rt.builtins-$(uname -m).a")"; \
```
RomainMuller: This retains the same behavior, but is a simplified/more compact way to do it.

duncanista (approved):

PR #879, opened 2025-09-30
fix(logs): use channel-based aggregator

shreyamalpani:
Use a channel-based aggregator for logs to avoid locking on aggregation. Adds an aggregator service which receives incoming commands on a channel. Commands can be flush requests, inserts, or shutdowns.

Follows a similar pattern as https://github.com/DataDog/serverless-components/pull/32

Tested that dual shipping still works as well.

[SVLS-7646](https://datadoghq.atlassian.net/browse/SVLS-7646)

[SVLS-7646]: https://datadoghq.atlassian.net/browse/SVLS-7646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/logs/lambda/processor.rs
+++ b/bottlecap/src/logs/lambda/processor.rs
@@ -741,7 +741,6 @@ mod tests {
         let (tx, _rx) = tokio::sync::mpsc::channel(2);
         let (aggregator_service, aggregator_handle) = AggregatorService::default();
 
-        // Spawn the aggregator service
         let service_handle = tokio::spawn(async move {
             aggregator_service.run().await;
         });


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -376,11 +382,15 @@ async fn extension_loop_active(
         .to_string();
     let tags_provider = setup_tag_provider(&Arc::clone(&aws_config), config, &account_id);
 
+    let (logs_aggr_service, logs_aggr_handle) = LogsAggregatorService::default();
+    start_logs_aggregator(logs_aggr_service);
+
```
duncanista: I think this logic can live in the `start_logs_agent` as no one else uses this but logs related stuff

bottlecap/src/logs/aggregator_service.rs
```diff
@@ -0,0 +1,151 @@
+use log::error;
```
duncanista: Don't use `log::error`, use the one from the `tracing` crate

PR #882, opened 2025-10-01
feat: remove failover to go

astuyve:
Removes the failover to Go. If we can't parse any of the config options we log the failing value and move on with the default specified.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -252,12 +252,14 @@ async fn main() -> anyhow::Result<()> {
 
     let cloned_client = client.clone();
     let runtime_api = aws_config.runtime_api.clone();
-    let response =
-        tokio::task::spawn(async move { extension::register(&cloned_client, &runtime_api).await });
+    let response = tokio::task::spawn(async move {
+        extension::register(&cloned_client, &runtime_api, extension::EXTENSION_NAME).await
+    });
     // First load the AWS configuration
     let lambda_directory: String =
         env::var("LAMBDA_TASK_ROOT").unwrap_or_else(|_| "/var/task".to_string());
     let config = Arc::new(config::get_config(Path::new(&lambda_directory)));
+
     let aws_config = Arc::new(aws_config);
     let api_key_factory = create_api_key_factory(&config, &aws_config);
 
--- a/.gitlab/templates/pipeline.yaml.tpl
+++ b/.gitlab/templates/pipeline.yaml.tpl
@@ -46,32 +46,10 @@ cargo clippy:
 
 {{ range $flavor := (ds "flavors").flavors }}
 
-go agent ({{ $flavor.name }}):
-  stage: compile
-  image: registry.ddbuild.io/images/docker:20.10
-  tags: ["arch:amd64"]
-  needs: []
-  artifacts:
-    expire_in: 1 week
-    paths:
-      - .binaries/datadog-agent-{{ $flavor.suffix }}
-  variables:
-    ARCHITECTURE: {{ $flavor.arch }}
-    ALPINE: {{ $flavor.alpine }}
-    FILE_SUFFIX: {{ $flavor.suffix }}
-    FIPS: {{ $flavor.fips }}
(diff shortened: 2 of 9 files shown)


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -289,26 +290,16 @@ fn init_ustr() {
...
-            let err = Command::new("/opt/datadog-agent-go").exec();
-            panic!("Error starting the extension: {err:?}");
-        }
-    };
-
-    (aws_config, config)
-}
+fn enable_logging_subsystem() {
+    let log_level = LogLevel::from_str(
+        std::env::var("DD_LOG_LEVEL")
```
duncanista, line 302: what if set on yaml

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -253,9 +249,14 @@ async fn main() -> anyhow::Result<()> {
...
 
-    let r = extension::register(&client, &aws_config.runtime_api, extension::EXTENSION_NAME)
+    let r = extension::register(&client, &aws_config.runtime_api)
         .await
         .map_err(|e| anyhow::anyhow!("Failed to register extension: {e:?}"))?;
+    // First load the AWS configuration
+    let lambda_directory: String =
+        env::var("LAMBDA_TASK_ROOT").unwrap_or_else(|_| "/var/task".to_string());
+    let config = Arc::new(config::get_config(Path::new(&lambda_directory)));
+    init_ustr();
```
duncanista: Why this moved from being at the top to being after? I remember where it was positioned matter to lower init

bottlecap/src/config/mod.rs
```diff
@@ -453,90 +450,20 @@ impl Default for Config {
...
-            .otlp_config_metrics_sums_cumulative_monotonic_mode
-            .is_some()
-        || config
-            .otlp_config_metrics_sums_initial_cumulativ_monotonic_value
-            .is_some()
-        || config.otlp_config_metrics_summaries_mode.is_some()
-        || config
-            .otlp_config_traces_probabilistic_sampler_sampling_percentage
-            .is_some()
-        || config.otlp_config_logs_enabled;
```
duncanista, line 518: We still don't support any of this, should we warn at least? It has never been supported, but just wondering

bottlecap/src/config/mod.rs
```diff
@@ -453,90 +450,20 @@ impl Default for Config {
...
     let mut config_builder = ConfigBuilder::default()
         .add_source(Box::new(YamlConfigSource { path }))
         .add_source(Box::new(EnvConfigSource));
 
-    let config = config_builder.build();
-
-    fallback(&config)?;
-
-    Ok(config)
+    config_builder.build()
```
duncanista:
```suggestion
ConfigBuilder::default()
.add_source(Box::new(YamlConfigSource { path }))
.add_source(Box::new(EnvConfigSource))
.build()
```

duncanista (approved):
Left some comments, but LGTM

bottlecap/src/bin/bottlecap/main.rs
astuyve: I'm not actually sure we honor that? Even the serverless flare tools use the env vars instead of a yaml file. IMO this is a safe thing to assume.

bottlecap/src/bin/bottlecap/main.rs
astuyve: I can move it back up, I have another change which modifies this anyway to make the registration call a task

bottlecap/src/bin/bottlecap/main.rs
duncanista: https://github.com/DataDog/datadog-agent/blob/989b241375676b4fdf3ca9b8523f9083e75380cb/pkg/config/config_template.yaml#L2701-L2707

Up to you, happy to not care about it

bottlecap/src/config/mod.rs
astuyve: I think it's ok for now.

PR #886, opened 2025-10-02
fix: tombstone event to be sent properly

duncanista:
# What?

The Telemetry API needs to signal the main event loop that it stopped processing events. The order of how this operations occurred could have been busted before. This fixes it and moves the code around so the cancel token can trigger the Tombstone signal.

# Why?

Reduces time spent on shutdown when processing

# Notes

Also makes the Tombstone event to be a main event as opposed to a Telemetry API event

### Before

image

### After

Screenshot 2025-10-03 at 3 46 34 PM


lym953:
nit: Should we keep the format consistent? For example, `PROXY_FLUSHER` is in all caps, but `Trace Agent` is not.

Issue #834, opened 2025-09-15
Cached Datadog api key doesn't play well with snapstart in V84+


kirkrd:
We are using .NET lambdas with snapstart and rotate our DD key every day, after upgrading to the latest version we noticed that the inflow of data and logs stop after a certain period of time (the keys are stuck and expire).

Suspected commit causing this:
https://github.com/DataDog/datadog-lambda-extension/pull/717

Comment found that might be causing the issue
https://github.com/DataDog/datadog-lambda-extension/pull/717#discussion_r2183150362

Our solution now is to downgrade, but would love to have keep the same behaviour as pre V84.

lym953:
Thanks for the report! Looking.

kirkrd:
But perhaps its better to use the arn directly instead of reading the secret value directly, will try that and report back

lym953:
@kirkrd Thanks for the context. Using `apiKeySecret` of the `DatadogLambda` construct results in `DD_API_KEY_SECRET_ARN` env var being set, which is affected by this bug. You don't need to try using the arn directly. The bug will be fixed in the next release.

lym953:
@kirkrd Could you try v87?

kirkrd:
Hello again, we tried v87 and using apiKeySecretArn instead of apiKeySecret, same issue @lym953

lym953:
@kirkrd Thanks for the update. Let me reopen this issue and take a look.

lym953:
Hey @kirkrd. I reproduced the issue on v87. After the secret value is rotated, data stops to be sent to Datadog, but invocations can complete without timeout.
But v83 has a different problem. After the secret value is rotated, the next invocation times out (2 minutes in my case). Then the extension restarts, which loads the updated DD API Key, and logs for subsequent invocations can be sent to Datadog. In short, one invocation experiences (1) timeout (2) missing data in Datadog. Is that what you see?

lym953:
@kirkrd Could you keep a grace period for the old key? i.e. after you update the secret to the new key, keep the old key for a while (e.g. 20 minutes). Then, I can make the extension support refreshing the key every 10 minutes, so the key in use will be always valid.

kirkrd:
@lym953 The issue is now resolved, no logs missing after the weekend! 🥇

PR #893, opened 2025-10-14
feat: Support periodic reload for api key secret

lym953:
# This PR
Supports the env var `DD_API_KEY_SECRET_RELOAD_INTERVAL`, in seconds. It applies when Datadog API Key is set using `DD_API_KEY_SECRET_ARN`. For example:
- if it's `120`, then api key will be reloaded about every 120 seconds. Note that reload can only be triggered when api key is used, usually when data is being flushed. If there is no invocation and no data needs to be flushed, then reload won't happen.
- If it's not set or set to `0`, then api key will only be loaded once the first time it is used, and won't be reloaded.

# Motivation
Some customers regularly rotate their api key in a secret. We need to provide a way for them to update our cached key.
https://github.com/DataDog/datadog-lambda-extension/issues/834

# Testing
## Steps
1. Set the env var `DD_API_KEY_SECRET_RELOAD_INTERVAL` to `120`

2. Invoke the Lambda every minute

## Result
The reload interval is passed to the `ApiKeyFactory`
image

Reload happens roughly every 120 seconds. It's sometimes longer than 120 seconds due to the reason explained above.
image

# Notes to Users
When you use this env var, please also keep a grace period for the old api key after you update the secret to the new key, and make the grace period longer than the reload interval to give the extension sufficient time to reload the secret.

# Internal Notes
Jira: https://datadoghq.atlassian.net/browse/SVLS-7572

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -343,7 +343,7 @@ pub struct Config {
     pub capture_lambda_payload: bool,
     pub capture_lambda_payload_max_depth: u32,
     pub compute_trace_stats_on_extension: bool,
-    pub api_key_reload_interval: Option<Duration>,
+    pub api_key_secret_reload_interval: Option<Duration>,
 
     pub serverless_appsec_enabled: bool,
     pub appsec_rules: Option<String>,
@@ -445,7 +445,7 @@ impl Default for Config {
             capture_lambda_payload: false,
             capture_lambda_payload_max_depth: 10,
             compute_trace_stats_on_extension: false,
-            api_key_reload_interval: None,
+            api_key_secret_reload_interval: None,
 
             serverless_appsec_enabled: false,
             appsec_rules: None,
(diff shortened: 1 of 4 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -745,6 +747,17 @@ pub fn deserialize_optional_duration_from_seconds<'de, D: Deserializer<'de>>(
     deserializer.deserialize_any(DurationVisitor)
 }
 
+// Like deserialize_optional_duration_from_seconds(), but return None if the value is 0
+pub fn deserialize_optional_duration_from_seconds_ignore_zero<'de, D: Deserializer<'de>>(
+    deserializer: D,
+) -> Result<Option<Duration>, D::Error> {
+    let duration: Option<Duration> = deserialize_optional_duration_from_seconds(deserializer)?;
+    if duration.is_some_and(|d| d.as_secs() == 0) {
```
litianningdatadog, line 755: As deserialize_optional_duration_from_seconds may return Ok(None) in case of negative value, is this logic safe to return Ok(None) accordingly?

lym953: I think it's safe. If value is negative, then it's invalid. In this case, the function returns `Ok(None)`, and api key secret won't be reloaded, as if the env var is not set.

litianningdatadog: Resolved offline. It will lead to Ok(None) if in line 758 for negative value.

PR #894, opened 2025-10-14
remove invocation lock

jchrostek-dd:
## Task
https://datadoghq.atlassian.net/browse/SVLS-7465?atlOrigin=eyJpIjoiYzI2ZWQ3NmU1M2U4NGFiMGI1ZjE3MWY5MTM2MzlkN2MiLCJwIjoiaiJ9

## Overview
Remove InvocationProcessor lock and replace with an aggregator service.

### DIFF
--- a/bottlecap/src/traces/trace_agent.rs
+++ b/bottlecap/src/traces/trace_agent.rs
@@ -485,10 +485,8 @@ impl TraceAgent {
         let mut reparenting_info = match invocation_processor_handle.get_reparenting_info().await {
             Ok(info) => info,
             Err(e) => {
-                return error_response(
-                    StatusCode::INTERNAL_SERVER_ERROR,
-                    format!("Failed to get reparenting info: {e}"),
-                );
+                error!("Failed to get reparenting info: {e}");
+                VecDeque::new()
             }
         };
 


bottlecap/src/lifecycle/invocation/processor_service.rs
```diff
@@ -0,0 +1,504 @@
...
+            .await
+            .map_err(|e| ProcessorError::ChannelReceive(e.to_string()))?
+    }
+
+    pub async fn update_reparenting(
+        &self,
+        reparenting_info: std::collections::VecDeque<ReparentingInfo>,
+    ) -> Result<Vec<crate::lifecycle::invocation::context::Context>, ProcessorError> {
+        let (response_tx, response_rx) = oneshot::channel();
+
```
litianningdatadog, line 265: would be great to refactor these.

bottlecap/src/lifecycle/invocation/processor_service.rs
```diff
@@ -0,0 +1,504 @@
...
+        Processor::extract_span_context(headers, payload_value, propagator)
+    }
+
+    pub async fn set_cold_start_span_trace_id(
+        &self,
+        trace_id: u64,
+    ) -> Result<Option<u64>, ProcessorError> {
+        let (response_tx, response_rx) = oneshot::channel();
+
+        self.sender
```
litianningdatadog, line 294: Same here

bottlecap/src/lifecycle/invocation/processor_service.rs
jchrostek-dd: what did you have in mind?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -813,35 +822,42 @@ async fn handle_event_bus_event(
...
-                        event.time.timestamp(),
-                    )
-                    .await;
-                    drop(p);
+                    if let Err(e) = invocation_processor_handle
+                        .on_platform_runtime_done(
+                            request_id.clone(),
+                            metrics,
+                            status,
+                            error_type.clone(),
```
litianningdatadog, line 839: if wait_for_platform_runtime_done_completion=false, cycle could be wasted due to async. Given that it is a more frequent scenario, would we consider to use async only if necessary?

bottlecap/src/lifecycle/invocation/processor_service.rs
```diff
@@ -0,0 +1,504 @@
...
+                    self.processor.on_shutdown_event();
+                }
+                ProcessorCommand::SendCtxSpans {
+                    tags_provider,
+                    trace_sender,
+                    context,
+                } => {
+                    self.processor
+                        .send_ctx_spans(&tags_provider, &trace_sender, context)
+                        .await;
```
litianningdatadog, line 495: why do we wait here, unlike other processor calls?

bottlecap/src/bin/bottlecap/main.rs
jchrostek-dd: What do you mean by cycle wasted due to async? If wait_for_platform_runtime_done_completion=false, then invocation_processor_handle.on_platform_runtime_done() should pretty much immediately return.

bottlecap/src/lifecycle/invocation/processor_service.rs
jchrostek-dd: oops, that is a typo, will remove 'await'

jchrostek-dd: Actually .await is needed since send_ctx_spans() is async.

bottlecap/src/lifecycle/invocation/processor_service.rs
```diff
@@ -0,0 +1,504 @@
...
+impl InvocationProcessorService {
+    #[must_use]
+    pub fn new(
+        tags_provider: Arc<provider::Provider>,
+        config: Arc<config::Config>,
+        aws_config: Arc<AwsConfig>,
+        metrics_aggregator_handle: AggregatorHandle,
+        propagator: Arc<DatadogCompositePropagator>,
+    ) -> (InvocationProcessorHandle, Self) {
+        let (sender, receiver) = mpsc::unbounded_channel();
```
duncanista: Maybe we want to bound it to a 1000? @astuyve what would be a good number?

jchrostek-dd: Yeah, good point.

bottlecap/src/traces/trace_agent.rs
```diff
@@ -482,9 +482,14 @@ impl TraceAgent {
...
-            invocation_processor.get_reparenting_info()
+        let mut reparenting_info = match invocation_processor_handle.get_reparenting_info().await {
+            Ok(info) => info,
+            Err(e) => {
+                return error_response(
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    format!("Failed to get reparenting info: {e}"),
+                );
+            }
         };
```
duncanista, line 491: Not sure if we should be erroring here, maybe just logging the error is enough? What happens if it failed before?

duncanista: I guess this error is mostly when failing to send/receive from the channel?

bottlecap/src/lifecycle/invocation/processor_service.rs
```diff
@@ -0,0 +1,504 @@
...
+        request_id: String,
+        metrics: RuntimeDoneMetrics,
+        status: Status,
+        error_type: Option<String>,
+        tags_provider: Arc<provider::Provider>,
+        trace_sender: Arc<SendingTraceProcessor>,
+        timestamp: i64,
+        wait_for_completion: bool,
+    ) -> Result<(), ProcessorError> {
+        // Only when Flush decision is 'End' do we want to wait to finish processing PlatformRuntimeDone
```
duncanista: Why?

jchrostek-dd: Otherwise aren't we going to start flushing events before potentially processing all of them?

bottlecap/src/lifecycle/invocation/processor_service.rs
litianningdatadog: something like 

    async fn send_and_await<T, F>(
        &self,
        command_builder: F,
    ) -> Result<T, ProcessorError>
    where
        F: FnOnce(oneshot::Sender<Result<T, ProcessorError>>) -> ProcessorCommand,
    {
        let (response_tx, response_rx) = oneshot::channel();

        self.sender
            .send(command_builder(response_tx))
            .await
            .map_err(|e| ProcessorError::ChannelSend(e.to_string()))?;

        response_rx
            .await
            .map_err(|e| ProcessorError::ChannelReceive(e.to_string()))?
    }
    pub async fn get_reparenting_info(
        &self,
    ) -> Result<std::collections::VecDeque<ReparentingInfo>, ProcessorError> {
        self.send_and_await(|response| ProcessorCommand::GetReparentingInfo { response })
            .await
    }
    pub async fn update_reparenting(
        &self,
        reparenting_info: std::collections::VecDeque<ReparentingInfo>,
    ) -> Result<Vec<Context>, ProcessorError> {
        self.send_and_await(|response| ProcessorCommand::UpdateReparenting {
            reparenting_info,
            response,
        })
        .await
    }

Same for other places. WDYT?

bottlecap/src/bin/bottlecap/main.rs
litianningdatadog: while it could return right away, since it is an async call, I would imagine it will incur some overhead which doesn't happen in non-async fashion.

PR #901, opened 2025-10-21
refactor enhanced metrics monitoring

shreyamalpani:
Use a long-running task that resumes at the start of an invocation, instead of spawning a new task on each invocation. Uses a channel-based service to pause and resume monitoring.

Tested with a function that opens a large number of file descriptors to make sure it still captures data accurately. Tested with self-monitoring as well (new implementation, previous implementation):
Screenshot 2025-10-20 at 4 16 47 PM

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor_service.rs
+++ b/bottlecap/src/lifecycle/invocation/processor_service.rs
@@ -100,7 +100,7 @@ pub enum ProcessorCommand {
     SendCtxSpans {
         tags_provider: Arc<provider::Provider>,
         trace_sender: Arc<SendingTraceProcessor>,
-        context: Context,
+        context: Box<Context>,
     },
     Shutdown,
 }
@@ -338,7 +338,7 @@ impl InvocationProcessorHandle {
             .send(ProcessorCommand::SendCtxSpans {
                 tags_provider: Arc::clone(tags_provider),
                 trace_sender: Arc::clone(trace_sender),
-                context,
+                context: Box::new(context),
             })
             .await
     }
@@ -491,7 +491,7 @@ impl InvocationProcessorService {
                     context,
                 } => {
                     self.processor
-                        .send_ctx_spans(&tags_provider, &trace_sender, context)
+                        .send_ctx_spans(&tags_provider, &trace_sender, *context)
                         .await;
                 }
                 ProcessorCommand::Shutdown => {


bottlecap/src/metrics/enhanced/usage_metrics.rs
```diff
@@ -0,0 +1,238 @@
...
+impl Default for UsageMetrics {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[derive(Debug)]
+pub enum EnhancedMetricsCommand {
+    UpdateMetrics(Option<f64>, Option<f64>, Option<f64>),
+    ResetMetrics(),
```
litianningdatadog: nit: ResetMetrics

shreyamalpani: What should it be changed to?

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -103,11 +102,14 @@ impl Processor {
             config.trace_aws_service_representation_enabled,
         );
 
+        let enhanced_metrics = EnhancedMetrics::new(metrics_aggregator, Arc::clone(&config));
+        enhanced_metrics.start_enhanced_metrics_task(); // starts the long-running task that monitors usage metrics (fd_use, threads_use, tmp_used)
```
duncanista: I'd rename this to something like `start_usage_metrics_task` since it's already enhanced metrics(?)

bottlecap/src/metrics/enhanced/usage_metrics.rs
```diff
@@ -0,0 +1,238 @@
...
+    GetMetrics(oneshot::Sender<UsageMetrics>),
+    Shutdown,
+}
+
```
duncanista:
```suggestion
pub enum Command {
    Update(Option<f64>, Option<f64>, Option<f64>),
    Reset(),
    Get(oneshot::Sender<UsageMetrics>),
    Shutdown,
}

```

bottlecap/src/metrics/enhanced/usage_metrics.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            tmp_used: 0.0,
+            fd_use: 0.0,
+            threads_use: 0.0,
+        }
+    }
+
+    pub fn update_metrics(
```
duncanista:
```suggestion
    pub fn update(
```

duncanista (approved):

Issue #899, opened 2025-10-17
DD_EXTENSION | ERROR | Failed to buffer the request body: length limit exceeded


ex-nerd:
I recently upgraded from v65 to v87 (and now to v88) and am suddenly seeing this error in some of my lambda logs:

`DD_EXTENSION | ERROR | Failed to buffer the request body: length limit exceeded`

Additionally, about 5s after this an uncaught exception is logged by the lambda layer:

```
Error: write EPIPE
at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:95:16)
at WriteWrap.callbackTrampoline (node:internal/async_hooks:130:17)
```

Followed almost immediately by:

```
LAMBDA_RUNTIME Failed to post handler success response. Http response code: 403. {"errorMessage":"State transition from Ready to InvocationErrorResponse failed for runtime. Error: State transition is not allowed","errorType":"InvalidStateTransition"}
```

This ultimately results in my API calls receiving `502` errors from the server.

Unfortunately, due to the error itself I don't know much about the request itself other than the response payload is rather large.

Even if DD can't send logs to its own server, that should _never_ prevent a response from being returned to my API clients.

lym953:
@ex-nerd Thanks for reporting. To help us understand the issue: does your Lambda use a tracing Layer such as [datadog-lambda-js](https://github.com/DataDog/datadog-lambda-js) like `arn:aws:lambda:us-east-1:464622532012:layer:Datadog-Node22-x:129`? If so, what's the version?

ex-nerd:
Yes, I upgraded to `129` at the same time. I have since reverted the lambda extension to `65` but kept the lambda js at `129` and the errors are no longer happening.

lym953:
Thanks for the info. I reproduced the issue. Will take further look next week.

mdnorman:
We've been using it for a couple of weeks and haven't run into it so far.

PR #902, opened 2025-10-21
feat: Increase trace payload limit from 2 MiB to 50 MiB

lym953:
# This PR
Change the trace request limit from 2 MiB to 50 MiB.

# Motivation
When the Node.js tracer layer sends a request to Lambda extension that's between 2 MiB and 50 MiB, the extension closes the HTTP connection, the tracer gets an `EPIPE` error and breaks. (Maybe the tracer should handle the error better, but that's out of scope of this PR.)

According to @rochdev:
> the agent is supposed to have a limit of 50mb

So let's change the limit on agent side to match the expectation.

# Testing
Tested with Node.js 22 Lambda with this handler:
```
import tracer from 'dd-trace';
import crypto from 'crypto';
tracer.init();

function randomGarbage(len) {
  // low-compressibility payload (random bytes -> base64)
  return crypto.randomBytes(len).toString('base64');
}

export const handler = async (event) => {
  const SPANS = 3000;
  const TAG_BYTES_PER_SPAN = 20_000; // ~20 KB per span tag (base64 expands a bit)

  const root = tracer.startSpan('repro.root');
  root.setTag('dd.repro', 'true');

  for (let i = 0; i < SPANS; i++) {
    console.log(`Sending the ${i}-th span`);
    const span = tracer.startSpan('repro.child', { childOf: root });
    span.setTag('blob', randomGarbage(TAG_BYTES_PER_SPAN));
    span.finish();
  }
  root.finish();

  const response = {
    statusCode: 200,
    body: JSON.stringify('Hello from Lambda!'),
  };
  return response;
};
```

### Before:
There are errors like:
```
Error: write EPIPE
at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:95:16)
at WriteWrap.callbackTrampoline (node:internal/async_hooks:130:17)
```
```
LAMBDA_RUNTIME Failed to post handler success response. Http response code: 403. {"errorMessage":"State transition from Ready to InvocationErrorResponse failed for runtime. Error: State transition is not allowed","errorType":"InvalidStateTransition"}
```

### After
When Lambda's memory is 1024 MB, the error no longer happens.
When Lambda's memory is 512 MB, the invocation can fail due to OOM. But I think that's a legit error. We can ask customers to increase memory limit for high-volume workload like this.

# Notes
cc @astuyve who set a `MAX_CONTENT_LENGTH` of 10 MiB in https://github.com/DataDog/datadog-lambda-extension/pull/294. This PR increases it to 50 MiB as well.

Thanks @dougqh @duncanista @lucaspimentel @rochdev for discussion.

https://github.com/DataDog/datadog-lambda-extension/issues/899
Jira: https://datadoghq.atlassian.net/browse/SVLS-7777

### DIFF
--- a/bottlecap/src/traces/trace_agent.rs
+++ b/bottlecap/src/traces/trace_agent.rs
@@ -262,7 +262,7 @@ impl TraceAgent {
             .layer(RequestBodyLimitLayer::new(DEFAULT_REQUEST_BODY_LIMIT))
             .with_state(proxy_state);
 
-            let info_router = Router::new().route(INFO_ENDPOINT_PATH, any(Self::info));
+        let info_router = Router::new().route(INFO_ENDPOINT_PATH, any(Self::info));
 
         Router::new()
             .merge(trace_router)


bottlecap/src/traces/trace_agent.rs
```diff
@@ -254,16 +259,21 @@ impl TraceAgent {
                 INSTRUMENTATION_ENDPOINT_PATH,
                 post(Self::instrumentation_proxy),
             )
+            .layer(RequestBodyLimitLayer::new(DEFAULT_REQUEST_BODY_LIMIT))
             .with_state(proxy_state);
 
-        let info_router = Router::new().route(INFO_ENDPOINT_PATH, any(Self::info));
+        let info_router = Router::new()
+            .route(INFO_ENDPOINT_PATH, any(Self::info))
+            .layer(RequestBodyLimitLayer::new(DEFAULT_REQUEST_BODY_LIMIT));
```
duncanista: this endpoint probably doesn't need it

PR #909, opened 2025-10-27
fix(traces): use channel-based aggregator

shreyamalpani:
Use a channel-based aggregator for traces to avoid locking on aggregation. Adds an aggregator service which receives incoming commands on a channel. Commands can be to insert payloads, get batches, and shutdown.

Follows a similar pattern as #879. Tested with self-monitoring and tested that dual shipping still works as well.

[SVLS-7647](https://datadoghq.atlassian.net/browse/SVLS-7647)

[SVLS-7647]: https://datadoghq.atlassian.net/browse/SVLS-7647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -45,7 +45,10 @@ use bottlecap::{
     },
     logger,
     logs::{
-        agent::LogsAgent, aggregator_service::AggregatorService as LogsAggregatorService,
+        agent::LogsAgent,
+        aggregator_service::{
+            AggregatorHandle as LogsAggregatorHandle, AggregatorService as LogsAggregatorService,
+        },
         flusher::LogsFlusher,
     },
     otlp::{agent::Agent as OtlpAgent, should_enable_otlp_agent},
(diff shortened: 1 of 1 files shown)


bottlecap/src/traces/trace_aggregator_service.rs
```diff
@@ -0,0 +1,155 @@
...
+                AggregatorCommand::GetBatches(response_tx) => {
+                    let mut batches = Vec::new();
+                    let mut current_batch = self.aggregator.get_batch();
+                    while !current_batch.is_empty() {
+                        batches.push(current_batch);
+                        current_batch = self.aggregator.get_batch();
+                    }
+                    if response_tx.send(batches).is_err() {
+                        error!("Failed to send trace flush response - receiver dropped");
+                    }
```
litianningdatadog, line 89: would be great to log the error details.

litianningdatadog (commented):
Overall LGTM

bottlecap/src/traces/trace_aggregator_service.rs
shreyamalpani: I think the error here just contains the data that failed to send

litianningdatadog: I am thinking of sth like below. It will give us more context in case of errors. 
```
if let Err(e) = response_tx.send(batches) {
   error!("Failed to send trace flush response : {}", e);
}
```

shreyamalpani: I tried this, but here the error is the `Vec<Vec<SendDataBuilder>>` of data that failed to send and doesn't have additional context I believe

PR #912, opened 2025-10-30
fix: Fix log processing rule on orphan log

lym953:
# Problem
When a user follows [our doc](https://docs.datadoghq.com/serverless/aws_lambda/logs/?tab=serverlessframework#filter-or-scrub-information-from-logs) to set the env var: `DD_LOGS_CONFIG_PROCESSING_RULES `: `[{"type": "exclude_at_match", "name": "exclude_start_and_end_logs", "pattern": "(START|END|REPORT) RequestId"}]`, we not only excludes `START`, `END` and `REPORT` logs, but some other logs as well, such as (1) extension log and (2) error log from the Lambda handler. As a result, errors can be left unobserved.

# Bug
This is because we first apply the log processing rules to the log we receive from telemetry API, to decide whether to send it to Datadog, then use the same result for "orphan log", which may include extension log and user error log.

# This PR
For each orphan log, evaluate log processing rules again, instead of reusing the existing result.

# Testing
## Setup
- Lambda runtime: Node 22
- Handler:
```
const region = process.env["NO_REGION"];
if (!region) {
  throw new Error("NO_REGION environment variable not set");
}


export const handler = async (event) => {
  const response = {
    statusCode: 200,
    body: JSON.stringify('Hello from Lambda!'),
  };
  return response;
};
```
- Env var: `DD_LOGS_CONFIG_PROCESSING_RULES `: `[{"type": "exclude_at_match", "name": "exclude_start_and_end_logs", "pattern": "(START|END|REPORT) RequestId"}]`

## Result before:
No log showed up in Datadog's Log Explorer or Live Tail

## Result after:
Custom error log and extension log showed up.
image

# Notes
Thanks @duncanista @astuyve @litianningdatadog for debugging and discussion.

Jira: https://datadoghq.atlassian.net/browse/SLES-2469

### DIFF
--- a/bottlecap/src/logs/lambda/processor.rs
+++ b/bottlecap/src/logs/lambda/processor.rs
@@ -324,31 +324,30 @@ impl LambdaProcessor {
         }
     }
 
-    pub async fn process(&mut self, event: TelemetryEvent, aggregator_handle: &AggregatorHandle) {
-        if let Ok(mut log) = self.make_log(event).await {
-            let should_send_log = self.logs_enabled
-                && LambdaProcessor::apply_rules(&self.rules, &mut log.message.message);
-            if should_send_log {
-                if let Ok(serialized_log) = serde_json::to_string(&log) {
-                    // explicitly drop log so we don't accidentally re-use it and push
-                    // duplicate logs to the aggregator
-                    drop(log);
-                    self.ready_logs.push(serialized_log);
-                }
+    /// Processes a log, applies filtering rules, serializes it, and queues it for aggregation
+    fn process_and_queue_log(&mut self, mut log: IntakeLog) {
+        let should_send_log = self.logs_enabled
+            && LambdaProcessor::apply_rules(&self.rules, &mut log.message.message);
+        if should_send_log {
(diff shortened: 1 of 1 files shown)


bottlecap/src/logs/lambda/processor.rs
```diff
@@ -341,6 +341,8 @@ impl LambdaProcessor {
             for mut orphan_log in self.orphan_logs.drain(..) {
                 orphan_log.message.lambda.request_id =
                     Some(self.invocation_context.request_id.clone());
+                let should_send_log = self.logs_enabled
```
astuyve: Oh god, this failed because we shadowed the `should_send_log` variable? 

This reassignment works, but can we refactor this into an instance method which just borrows the message itself?

If that is too complicated to pass the borrow checker, can we instead rename this variable to `orphan_should_send_log` so we don't reshadow this?

lym953: It failed because we reused the result of `should_send_log` from the first message for all the log messages.
Will try to refactor this.

lym953: Done!

PR #914, opened 2025-11-04
add template

jchrostek-dd:
Add PR template.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,5 @@
 **Please include Jira ticket in title.**
 
 ## Overview
-<!-- What are we changing? Why are we changing this? -->
 
 ## Testing 
-<!-- How did you verify your changes? Unit tests? Manual testing? -->


.github/pull_request_template.md
```diff
@@ -0,0 +1,7 @@
+**Please include Jira ticket in title.**
```
apiarian-datadog, line 1: is there a way to turn this into a comment as well? what happens to comments anyway? when do they disappear? or do they persist in the markdown and the internals of the description? do they end up getting included in the squashed commit if the repo is configured to use the pr description as its body?

jchrostek-dd: What do you mean? You want a comment on the PR with the link to Jira?

litianningdatadog:
If the ultimate goal is to enable JIRA linkage, we can use this [this BP](https://datadoghq.atlassian.net/wiki/spaces/ENGPROG/pages/3172010447/Linking+PRs+to+Jira+Tickets). We can also add a Git hook to enforce this as a minimal merge requirement.

PR #916, opened 2025-11-06
[SVLS-7885] update tag splitting to allow for ',' and ' '

jchrostek-dd:
## Overview
We currently split the`DD_TAGS` only by `,`.  Customer is asking if we can also split by spaces since that is common for container images and lambda lets you deploy images. (https://docs.datadoghq.com/getting_started/tagging/assigning_tags/?tab=noncontainerizedenvironments)

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -556,7 +556,7 @@ where
                     continue;
                 }
                 let parts = tag.split(':').collect::<Vec<&str>>();
-                if parts.len() == 2 {
+                if parts.len() == 2 && !parts[0].is_empty() && !parts[1].is_empty() {
                     map.insert(parts[0].to_string(), parts[1].to_string());
                 } else {
                     error!(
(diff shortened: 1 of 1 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -551,15 +551,17 @@ where
             E: serde::de::Error,
         {
             let mut map = HashMap::new();
-
-            for tag in value.split(',') {
+            for tag in value.split(&[',', ' ']) {
+                if tag.is_empty() {
```
litianningdatadog, line 555: let tag = tag.trim();
 if tag.is_empty() {

jchrostek-dd: Since we are splitting by a space, it will already be trimmed. Confirmed with test `test_tags_space_separated_with_extra_spaces`

bottlecap/src/config/mod.rs
```diff
@@ -551,15 +551,17 @@ where
...
+                    continue;
+                }
                 let parts = tag.split(':').collect::<Vec<&str>>();
                 if parts.len() == 2 {
                     map.insert(parts[0].to_string(), parts[1].to_string());
```
litianningdatadog, line 560: In fact I meant to change below to skip empty key or value
`if parts.len() == 2 && !parts[0].is_empty() && !parts[1].is_empty() `

jchrostek-dd: Added.

PR #924, opened 2025-11-12
[SVLS-7704] add support for SSM Parameter API key

jchrostek-dd:
## Overview
* Add support for customers storing Datadog API Key in SSM Parameter Store.

## Testing
* Deployed changes and confirmed this work with Parameter Store String and SecureString.

### DIFF
--- a/bottlecap/src/secrets/decrypt.rs
+++ b/bottlecap/src/secrets/decrypt.rs
@@ -18,87 +18,97 @@ use tracing::debug;
 use tracing::error;
 
 pub async fn resolve_secrets(config: Arc<Config>, aws_config: Arc<AwsConfig>) -> Option<String> {
-    let api_key_candidate =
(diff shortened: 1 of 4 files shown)


bottlecap/src/secrets/decrypt.rs
```diff
@@ -18,87 +18,97 @@ use tracing::debug;
...
+                error!("Error creating reqwest client: {}", err);
+                return None;
+            }
+        };
 
-            let client = match builder.build() {
-                Ok(client) => client,
+        let mut aws_credentials = AwsCredentials::from_env();
+
+        if aws_credentials.aws_secret_access_key.is_empty()
```
duncanista, line 45: Maybe change it to a method in `aws_credentials`?

`aws_credentials.is_snap_start()`

duncanista (approved):
LGTM – might need a public docs update

PR #930, opened 2025-11-14
[SVLS-8002] fix: collect CPU metric offsets on PlatformStart

shreyamalpani:
## Overview

Collect CPU metric offsets on receiving the PlatformStart instead of the Invoke event to avoid accounting for additional idle time while calculating utilization. Also, guard uptime and idle time to be non-negative for edge-cases.

Maintains collecting network offsets on the Invoke event to match with Lambda Insights.

## Testing 

Tested with self-monitoring

image

### DIFF
--- a/bottlecap/src/metrics/enhanced/lambda.rs
+++ b/bottlecap/src/metrics/enhanced/lambda.rs
@@ -369,7 +369,7 @@ impl Lambda {
         let num_cores = cpu_data_end.individual_cpu_idle_times.len() as f64;
         let uptime = uptime_data_end - uptime_data_offset;
         let total_idle_time = cpu_data_end.total_idle_time_ms - cpu_data_offset.total_idle_time_ms;
-        
+
         // Change in uptime should be positive and greater than total idle time across all cores
         if uptime <= 0.0 || (uptime * num_cores) < total_idle_time {
             debug!("Invalid uptime, skipping CPU utilization metrics");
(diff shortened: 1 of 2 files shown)


bottlecap/src/metrics/enhanced/lambda.rs
```diff
@@ -370,6 +370,12 @@ impl Lambda {
         let uptime = uptime_data_end - uptime_data_offset;
         let total_idle_time = cpu_data_end.total_idle_time_ms - cpu_data_offset.total_idle_time_ms;
 
+        // Change in uptime should be positive and greater than total idle time across all cores
+        if uptime <= 0.0 || (uptime * num_cores) < total_idle_time {
```
litianningdatadog, line 374: Can we just check for non-negative uptime here? The other case is already handled by line 413 anyway.

Also, since we’ve changed when the offset is captured, is it still possible to see negative values? If not, that might eliminate the need to modify this file.

shreyamalpani: The additional check here means we don't send any CPU util metrics in this case, so I think we might want to keep to ensure we don't send invalid metrics.

I think I saw the occasional negative value without the guards so keeping that for edge-cases

litianningdatadog: ok, I am ok to keep this as long as we still see negative cases with the main change.

Issue #688, opened 2025-06-02
Decrease in trace metrics after updating to `>v68`


HiDAl:
Hello, 

some days ago my team tried to update the currently used layer to the latest version from `v65`. First, we tried with `v68` and noticed a huge decrease in the `hits` metrics generated through the tracing tool: for example when using `tracer.trace('name', ...)`, we saw a decrease in the metrics `trace.name.hits`. 
Checking documentation and such, we noticed that the env var `DD_TRACE_SAMPLE_RATE` was deprecated so we migrated to use the new one `DD_TRACE_SAMPLING_RULES` (moving from `DD_TRACE_SAMPLE_RATE=0.4` to `DD_TRACE_SAMPLING_RULES='[{"sample_rate": 0.4}]'`. Sadly, the issue was still there. We decided to give a try to `v78`, but the problem persisted. 
When enabling the debugging option, I see the following output (shortened for simplicity):

```json
// v65 using DD_TRACE_SAMPLE_RATE
{
  "os_name": "Linux",
  "os_version": "5.10.236-249.934.amzn2.aarch64",
  "architecture": "arm64",
  "version": "4.44.0",
  "lang": "nodejs",
  "lang_version": "20.19.1",
  "sample_rate": 0.4,
  "sampling_rules": [
    {
      "matchers": [],
      "_sampler": {
        "_rate": 0.4
      },
      "_limiter": {
        "_rateLimit": 100,
        "_limiter": {
          "tokenBucket": {
            "bucketSize": 100,
            "tokensPerInterval": 100,
            "interval": 1000,
            "parentBucket": null,
            "content": 99,
            "lastDrip": 1748592474968
          },
          "curIntervalStart": 1558141,
          "tokensThisInterval": 2
        },
        "_tokensRequested": 2,
        "_prevIntervalTokens": 0,
        "_prevTokensRequested": 0
      }
    }
  ],
  ...
}

// v78 using DD_TRACE_SAMPLING_RULES
{
  "os_name": "Linux",
  "os_version": "5.10.236-249.934.amzn2.aarch64",
  "architecture": "arm64",
  "version": "4.44.0",
  "lang": "nodejs",
  "lang_version": "20.19.1",
  "sampling_rules": [
    {
      "matchers": [],
      "_sampler": {
        "_rate": 0.4
      }
    }
  ],
  ...
}
```

[...]

PR #939, opened 2025-11-21
[SVLS-8169] feat: Support span dedup

lym953:
## TL;DR
- Support deduping spans by span_id
- which fixes trace stats overcount issue for Node.js

## Problem
Node.js tracer sometimes sends duplicate traces to the extension, causing over-count of trace stats. In our tests, the over-count is usually 1–4 for 50, 500 or 5000 invocations. This only happens when the "default" flushing strategy (which uses the "continuous" strategy) is used, and doesn't happen when the "end" strategy is used.

## Cause of problem
I think it's similar to a known problem of continuous flushing.
- The known problem is, when Lambda runtime is frozen, the connection between **extension** and **DD endpoint** can time out, causing data flush failure.
- In the trace stats case, the problem is, when Lambda runtime is frozen, the connection between **tracer** and **extension** can time out. Extension receives the request from tracer, then freezes before sending a response, causing the tracer's request to time out, which makes the tracer resend the trace. This doesn't happen with the END flush strategy because in that case, after the Lambda handler finishes, extension still needs to flush the data and doesn't freeze so fast, and it has enough time to respond to tracer.

### Steps
- Build a test extension layer
- Run e2e tests, including:
  - Install it on various Lambda functions
  - Invoke these functions with various traffic pattern
  - Check the trace stats result

### Result
**Before:**
- Over-count happens for (1) Node.js runtime + (2) "Sampling" test, which uses the default flush strategy
- (vs expected 50)

Screenshot 2025-11-20 at 9 30 00 PM

**After:**
- No over-count (vs expected 5000)
Screenshot 2025-11-20 at 9 25 32 PM

## Options considered
1. **At most once**: Disable retry in tracer, at least for Lambda.
2. **At least once**
    - 2.1 Do nothing. Call out this as a known limitation.
    - 2.2 Treat tracer as VIP. Before calling /next, make sure tracer's requests have been responded. This may cause regression on invocation duration, especially when volume is high.
3. **Exactly once**: Implement dedup in extension, by trace_id.

This PR chooses 3 because it's the easiest.

## Notes
Thanks @astuyve @rochdev @purple4reina for discussion.
Thanks Cursor for writing most of the code.
The under-count issue will be addressed separately.

Related issue: https://github.com/DataDog/datadog-lambda-extension/issues/688

### DIFF
--- a/bottlecap/Cargo.lock
+++ b/bottlecap/Cargo.lock
@@ -795,8 +795,8 @@ dependencies = [
 
 [[package]]
 name = "datadog-trace-obfuscation"
-version = "24.0.0"
-source = "git+https://github.com/DataDog/libdatadog?rev=422fae9093ec0b3b1b2638458597a12cc373d00a#422fae9093ec0b3b1b2638458597a12cc373d00a"
+version = "24.0.1"
+source = "git+https://github.com/DataDog/libdatadog?rev=7540423559548ce049c5424599d28ee1731378e8#7540423559548ce049c5424599d28ee1731378e8"
 dependencies = [
  "anyhow",
  "libdd-common",
@@ -1828,7 +1828,7 @@ checksum = "1171693293099992e19cddea4e8b849964e9846f4acee11b3948bcc337be8776"
 [[package]]
 name = "libdd-common"
 version = "1.0.0"
-source = "git+https://github.com/DataDog/libdatadog?rev=422fae9093ec0b3b1b2638458597a12cc373d00a#422fae9093ec0b3b1b2638458597a12cc373d00a"
+source = "git+https://github.com/DataDog/libdatadog?rev=7540423559548ce049c5424599d28ee1731378e8#7540423559548ce049c5424599d28ee1731378e8"
 dependencies = [
  "anyhow",
  "cc",
@@ -1861,7 +1861,7 @@ dependencies = [
 [[package]]
 name = "libdd-ddsketch"
 version = "1.0.0"
-source = "git+https://github.com/DataDog/libdatadog?rev=422fae9093ec0b3b1b2638458597a12cc373d00a#422fae9093ec0b3b1b2638458597a12cc373d00a"
+source = "git+https://github.com/DataDog/libdatadog?rev=7540423559548ce049c5424599d28ee1731378e8#7540423559548ce049c5424599d28ee1731378e8"
 dependencies = [
  "prost",
 ]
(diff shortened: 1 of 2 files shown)


bottlecap/src/traces/dedup.rs
```diff
@@ -0,0 +1,134 @@
+// Copyright 2023-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::{HashSet, VecDeque};
+
+const DEFAULT_CAPACITY: usize = 50;
+
+/// A deduplicator that maintains a fixed-size cache of recently seen IDs.
+/// When the capacity is reached, the oldest ID is evicted (FIFO).
+pub struct Deduper {
```
litianningdatadog: this deduper is based on span id. Is it guaranteed to be globally unique? ie. what if between 2 traces they happen to have the same span id?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -1062,6 +1063,10 @@ fn start_trace_agent(
 
     let stats_processor = Arc::new(stats_processor::ServerlessStatsProcessor {});
 
+    // Deduper
```
jchrostek-dd: These 1-line comments that just describe a single line of code feel a bit noisy. Can we remove them.

bottlecap/src/traces/span_dedup_service.rs
```diff
@@ -0,0 +1,147 @@
...
+    deduper: Deduper,
+    rx: mpsc::UnboundedReceiver<DedupCommand>,
+}
+
+/// A service that handles `check_and_add()` requests without using mutex.
+impl DedupService {
+    /// Creates a new dedup service with the default capacity (50).
+    #[must_use]
+    pub fn new() -> (Self, DedupHandle) {
+        Self::with_capacity(50)
```
jchrostek-dd: Don't think we need to let this be configurable, just always use the same size. Also, maybe worth bumping to 100? (non-blocking comment)

lym953: I want to keep it for tests so I can set capacity to 3 to test eviction. But I'll remove `pub`. Let me know if you have a better alternative.
I don't have a strong idea on capacity. Will bump to 100.

bottlecap/src/traces/dedup.rs
lym953: Good point. I didn't know two traces can have the same span id. Added trace id to dedup key.

bottlecap/src/bin/bottlecap/main.rs
lym953: Removed.

jchrostek-dd (approved):

PR #946, opened 2025-11-29
[SVLS-8054] add integration testing

jchrostek-dd:
## Overview
Setup integration tests for the lambda extension. These integration tests will run on every PR.

### Details
This PR includes:
* CDK stacks for deploying lambda integration tests. This is for the lambda, and any related resources, we want to test against.
* Integration tests, setup with Jest. These invoke lambda functions, wait, then get Datadog telemetry data to test/verify against.
* Gitlab Integration Test Step (info below).
* `README.md` for how to run tests locally.

Note:
* For simplicity, this is setup to just test against the ARM variant (not AMD). This also doesn't include FIPS or AppSec builds. I think this should be a reasonable starting point for our integration tests and we can evaluate adding additional configuration support as needed.

### Gitlab Integration Test Step
The integration tests step in Gitlab will:
1. Publish the lambda extension.
2. Deploy CDK stacks, using the newly published lambda extension.
3. Run a test suite.
4. Destroy the CDK stacks.
5. Delete the lambda extension.

### Executing the integration tests
The integration tests will automatically run on every PR. Developers can also run the integration tests locally by running `npm run test`. Full information is included in `README.md`.

### Example Integration Tests
I added a 2 basic tests, one for node and one for python. These lambda function logs 'Hello World' and are setup with the extension and tracer library. The integration test gets the logs and traces from Datadog. It confirms that we have a log with the message 'Hello World!'. It also confirms we have spans with names `aws.lambda.cold_start`, `aws.lambda.load` and `aws.lambda`. Note that this isn't actually working correct for python for `aws.lambda.load` and `aws.lambda.cold_start`. Those spans are created, but with a different traceId so they aren't getting linked to `aws.lambda`. I will follow up and investigate.

I plan on having a follow up PR with other runtimes.

## Testing 
This PR triggered the integration tests, can see the [corresponding gitlab pipeline](https://gitlab.ddbuild.io/DataDog/datadog-lambda-extension/-/pipelines/84401218) with the newly added step 'integration-tests'. (Or see the 'dd-gitlab/integration-test' in the checks for this PR)

The results from the integration test can be obtained by going to [integration step](https://gitlab.ddbuild.io/DataDog/datadog-lambda-extension/-/jobs/1262527100) and downloading the artifacts. Screenshot attached below. 

Screenshot 2025-12-01 at 9 14 03 AM

### DIFF
--- a/integration-tests/lib/stacks/base-python-stack.ts
+++ b/integration-tests/lib/stacks/base-python-stack.ts
@@ -21,7 +21,6 @@ export class BasePythonStack extends cdk.Stack {
         DD_SERVICE: pythonFunctionName,
         DD_TRACE_ENABLED: 'true',
         DD_LAMBDA_HANDLER: 'lambda_function.handler',
-        DD_LOG_LEVEL: 'debug',
         DD_TRACE_AGENT_URL: 'http://127.0.0.1:8126',
         DD_COLD_START_TRACING: 'true',
         DD_MIN_COLD_START_DURATION: '0',
--- a/integration-tests/lib/util.ts
+++ b/integration-tests/lib/util.ts
@@ -9,7 +9,6 @@ export const extensionLayerArn = process.env.EXTENSION_LAYER_ARN!;
 
 export const node20LayerArn = 'arn:aws:lambda:us-east-1:464622532012:layer:Datadog-Node20-x:130';
 export const python313LayerArn = 'arn:aws:lambda:us-east-1:464622532012:layer:Datadog-Python313-ARM:117';
-export const pythonLayerArn = 'arn:aws:lambda:us-east-1:425362996713:layer:John-Datadog-Python312-ARM:20';
 
 export const defaultDatadogEnvVariables = {
     DD_API_KEY_SECRET_ARN: datadogSecretArn,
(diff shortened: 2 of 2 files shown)


integration-tests/README.md
```diff
@@ -0,0 +1,70 @@
+# Datadog Lambda Extension Integration Tests
+
+This directory contains integration tests for the Datadog Lambda Extension.
```
apiarian-datadog, line 3: might be worth adding some notes here about the sorts of tests we should be including in this suite (vs the e2e test suite.)

integration-tests/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+This directory contains integration tests for the Datadog Lambda Extension. 
+
+The general flow is:
+1. Deploy test setup using CDK.
+2. Invoke lambda functions.
+3. Wait for data to propagate to Datadog.
+4. Call Datadog to get telemetry data and check the data based on test requirements.
+
+For simplicity, integraiton tests are setup to only test against ARM runtimes.
```
apiarian-datadog, line 11: i'd think that if this is intended to test the extension, then making sure that it is built correctly for arm vs. amd runtimes becomes important. or have we never seen architecture-dependent differences in our extension builds?

jchrostek-dd: I think just having arm for now is a sufficient starting point and we can expand on this later as needed/as we have time.

integration-tests/README.md
jchrostek-dd: I don't think there is too much to note - This is for the extension, the other e2e test suite is what the apm serverless team uses. There is overlap, but will let engineers use their judgement. Can re-evaluate this in the future if we are having issues.

PR #959, opened 2025-12-09
APPSEC-60188: downgrade integration response parsing log to debug

RomainMuller:
The message was incorrectly kept as `warn!` when we donwgraded all these log messages to `debug!` and this is very noisy in customer accounts.

JJ-Change-Id: zkowts
ZD: 2375557
Jira: APPSEC-60188

### DIFF
--- a/.github/workflows/serverless-init-vulnerability-scan.yml
+++ b/.github/workflows/serverless-init-vulnerability-scan.yml
@@ -35,7 +35,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Scan latest serverless-init image with grype
-        uses: anchore/scan-action@40a61b52209e9d50e87917c5b901783d546b12d0 # v7.2.1
+        uses: anchore/scan-action@3c9a191a0fbab285ca6b8530b5de5a642cba332f # v7.2.2
         with:
           image: "datadog/serverless-init:latest"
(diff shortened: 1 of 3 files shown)


bottlecap/src/appsec/processor/mod.rs
```diff
@@ -131,7 +131,7 @@ impl Processor {
                 ctx.run(payload.as_ref());
             }
             Ok(None) => debug!("aap: no response payload available"),
-            Err(e) => warn!("aap: failed to parse invocation result payload: {e}"),
+            Err(e) => debug!("aap: failed to parse invocation result payload: {e}"),
```
jchrostek-dd, line 134: Why is this a noisy log message for customers? Shouldn't this not be happening in most cases?

RomainMuller: Basically, if this happens on a given Lambda function, then it is likely to happen for _all_ invocations, as what we are parsing is the "core" structure of the API GW "Lambda Proxy" response... which is likely to be structured identically for all responses.

Ideally, we would produce _telemetry_ logs for these conditions, so that we can be made aware of these issues (as you said, it should not happen unless we have a bug -- as was the case here and it's been fixed via #960) without causing a logs surcharge to the customer.

PR #961, opened 2025-12-09
[SVLS-7934] feat: Support TLS certificate for trace/stats flusher

lym953:
## Problem
A customer reported that their Lambda is behind a proxy, and the Rust-based extension can't send traces to Datadog via the proxy, while the previous go-based extension worked.

## This PR
Supports the env var `DD_TLS_CERT_FILE`: The path to a file of concatenated CA certificates in PEM format.
Example: `DD_TLS_CERT_FILE=/opt/ca-cert.pem`, so the when the extension flushes traces/stats to Datadog, the HTTP client created can load and use this cert, and connect the proxy properly.

### Steps
1. Create a Lambda in a VPC with an NGINX proxy.
2. Add a layer to the Lambda, which includes the CA certificate `ca-cert.pem`
3. Set env vars:
    - `DD_TLS_CERT_FILE=/opt/ca-cert.pem`
    - `DD_PROXY_HTTPS=http://10.0.0.30:3128`, where `10.0.0.30` is the private IP of the proxy EC2 instance
    - `DD_LOG_LEVEL=debug`
4. Update routing rules of security groups so the Lambda can reach `http://10.0.0.30:3128`
5. Invoke the Lambda
### Result
**Before**
Trace flush failed with error logs:
> DD_EXTENSION | ERROR | Max retries exceeded, returning request error error=Network error: client error (Connect) attempts=1
DD_EXTENSION | ERROR | TRACES | Request failed: No requests sent

**After**
Trace flush is successful:
> DD_EXTENSION | DEBUG | TRACES | Flushing 1 traces
DD_EXTENSION | DEBUG | TRACES | Added root certificate from /opt/ca-cert.pem
DD_EXTENSION | DEBUG | TRACES | Proxy connector created with proxy: Some("http://10.0.0.30:3128")
DD_EXTENSION | DEBUG | Sending with retry url=https://trace.agent.datadoghq.com/api/v0.2/traces payload_size=1120 max_retries=1
DD_EXTENSION | DEBUG | Received response status=202 Accepted attempt=1
DD_EXTENSION | DEBUG | Request succeeded status=202 Accepted attempts=1
DD_EXTENSION | DEBUG | TRACES | Flushing took 1609 ms

## Notes
This fix only covers trace flusher and stats flusher, which use `ServerlessTraceFlusher::get_http_client()` to create the HTTP client. It doesn't cover logs flusher and proxy flusher, which use a different function (http.rs:get_client()) to create the HTTP client. However, logs flushing was successful in my tests, even if no certificate was added. We can come back to logs/proxy flusher if someone reports an error.

### DIFF
--- a/bottlecap/src/config/yaml.rs
+++ b/bottlecap/src/config/yaml.rs
@@ -54,7 +54,7 @@ pub struct YamlConfig {
     #[serde(deserialize_with = "deserialize_optional_string")]
     pub http_protocol: Option<String>,
     #[serde(deserialize_with = "deserialize_optional_string")]
-    pub cls_ca_cert: Option<String>,
+    pub tls_cert_file: Option<String>,
 
     // Endpoints
     #[serde(deserialize_with = "deserialize_additional_endpoints")]
@@ -419,7 +419,7 @@ fn merge_config(config: &mut Config, yaml_config: &YamlConfig) {
     merge_option!(config, proxy_https, yaml_config.proxy, https);
     merge_option_to_value!(config, proxy_no_proxy, yaml_config.proxy, no_proxy);
     merge_option!(config, yaml_config, http_protocol);
-    merge_option!(config, yaml_config, cls_ca_cert);
+    merge_option!(config, yaml_config, tls_cert_file);
 
     // Endpoints
     merge_hashmap!(config, yaml_config, additional_endpoints);
@@ -750,7 +750,7 @@ proxy:
   no_proxy: ["localhost", "127.0.0.1"]
 dd_url: "https://metrics.datadoghq.com"
 http_protocol: "http1"
-cls_ca_cert: "/opt/ca-cert.pem"
+tls_cert_file: "/opt/ca-cert.pem"
 
 # Endpoints
 additional_endpoints:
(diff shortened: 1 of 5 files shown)


bottlecap/src/config/yaml.rs
```diff
@@ -53,6 +53,8 @@ pub struct YamlConfig {
     pub dd_url: Option<String>,
     #[serde(deserialize_with = "deserialize_optional_string")]
     pub http_protocol: Option<String>,
+    #[serde(deserialize_with = "deserialize_optional_string")]
+    pub ssl_ca_cert: Option<String>,
```
duncanista: Where does the DD_SSL_CA_CERT environment variable comes from?

lym953: https://github.com/DataDog/datadog-lambda-extension/blob/7619068445fe431e2adb9a57919d81e6fd7956b5/bottlecap/src/config/env.rs#L653
It's handled by this code, which puts the value of env var `DD_SSL_CA_CERT` to the field `ssl_ca_cert`. Is this what you are asking?

duncanista: No, I'm asking where does `DD_SSL_CA_CERT` name convention comes from, does this come from the Datadog Agent config? Does it come from the docs? Did you came up with it?

lym953: Good question. I came up with it. Upon further search, I found `tls_ca_cert` here:
https://github.com/DataDog/integrations-core/blob/master/http_check/datadog_checks/http_check/data/conf.yaml.example#L477
so I'll rename the env var to `DD_SSL_CA_CERT`. Does this sound good to you?

duncanista: Yeah, although I'd confirm with the Agent team!

lym953: Sure, will do!

dustmop: Env var name LGTM from the perspective of agent-configuration team!

lym953: Discussed offline. Changing to `DD_TLS_CERT_FILE` to be consistent with:
https://github.com/DataDog/datadog-agent/blob/0638dfce1e1f3a9ae336334d4df01cb2a5e35120/pkg/config/setup/config.go#L1410

The same config option has different names in different places. This PR just picks one of them.

PR #967, opened 2025-12-17
[APPSEC-60218] Fix AWS API Gateway endpoints correlation HTTP span tags

e-n-0:
## Overview

Implementing changes from [RFC 1081](https://docs.google.com/document/d/1lH9RCZLR6SyJmPAIRYBpv4sKRy54kUAWOMPuOAOpS2Q/edit?tab=t.0#heading=h.4kapbym3f1x2).

- API Gateway REST/HTTP inferred spans now emit `span.type` as `web`
- droped `operation_name` and `apiname` (in api gateway REST, HTTP and websocket)
- Inferred spans now carry the API Gateway ARN in `dd_resource_key`
- Appsec enablement and the json data from the service entry span are propagated into inferred and wrapped spans

## Testing

- Tested if dd_resource_key are correctly set for api gateway (HTTP/REST)
- Tested if inferred spans correctly have the appsec data propagated (appsec enabled via entry span or via config)
- Updated tests to remove `operation_name` occurence

### DIFF
--- a/bottlecap/src/lifecycle/invocation/span_inferrer.rs
+++ b/bottlecap/src/lifecycle/invocation/span_inferrer.rs
@@ -628,29 +628,6 @@ mod tests {
     }
 
     #[test]
-    fn test_api_gateway_sets_dd_resource_key_for_rest_event() {
-        let payload = api_gateway_rest_payload();
-        let aws_config = aws_config("us-east-1");
-        let mut inferrer = SpanInferrer::new(Arc::new(Config::default()));
-
-        inferrer.infer_span(&payload, &aws_config);
-
-        let inferred_span = inferrer
-            .inferred_span
-            .as_ref()
-            .expect("Should have inferred API Gateway span");
-
-        assert_eq!(
-            inferred_span
-                .meta
-                .get("dd_resource_key")
-                .cloned()
-                .unwrap_or_default(),
-            "arn:aws:apigateway:us-east-1::/restapis/id"
-        );
-    }
-
-    #[test]
     fn test_complete_inferred_spans_propagates_appsec_from_invocation() {
         let payload = api_gateway_rest_payload();
         let aws_config = aws_config("us-east-1");
--- a/bottlecap/src/lifecycle/invocation/triggers/api_gateway_http_event.rs
+++ b/bottlecap/src/lifecycle/invocation/triggers/api_gateway_http_event.rs
@@ -442,6 +442,18 @@ mod tests {
     }
 
     #[test]
+    fn test_get_dd_resource_key() {
+        let json = read_json_file("api_gateway_http_event.json");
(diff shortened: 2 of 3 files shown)


duncanista:
Is the `dd_resource_key ` exclusive to API Gateway or all inferred spans?

e-n-0:
No, it’s not exclusive to API Gateway. `dd_resource_key` is the CCRID, which represents the underlying cloud resource. For API Gateway inferred spans we want it to be the gateway ARN, and for Lambda spans it should be the Lambda function ARN, so it applies to all inferred spans that represent a cloud resource, not just API Gateway.

Right now, the tag already exists and appears to be set by the backend (since it is already always present and the value we set for API Gateway is being overwrited by the Lambda function ARN instead). The backend team will need to investigate this behavior later, but it shouldn’t block the current work.

bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -209,6 +210,8 @@ impl SpanInferrer {
         };
 
         let identified_trigger = IdentifiedTrigger::from_value(payload_value);
+        let dd_resource_key =
+            Self::get_api_gateway_resource_key(&identified_trigger, &aws_config.region);
```
duncanista: Could you make this a trait method for Triggers as opposed to make it API Gateway specific?

You don't need to implement it for every inferred span, but I want to make sure the SpanInferrer class doesn't do any trigger-specific logic unless required

bottlecap/src/lifecycle/invocation/span_inferrer.rs
```diff
@@ -278,12 +287,15 @@ impl SpanInferrer {
                 invocation_span.service.clone(),
             );
             s.meta.insert("span.kind".to_string(), "server".to_string());
+            let appsec_enabled = self.config.serverless_appsec_enabled;
+            propagate_appsec(appsec_enabled, invocation_span, s);
```
duncanista, line 288: Why does the json need to be propagated up for the inferred spans?

duncanista (commented):
Left a comment and a question! 

Thanks for contributing 💯

bottlecap/src/lifecycle/invocation/span_inferrer.rs
e-n-0: Applied in 9acc70f and cddf7fc

duncanista (approved):

bottlecap/src/lifecycle/invocation/span_inferrer.rs
RomainMuller: This is to address an issue raised by a customer who wants to be able to link attacks to API Gateway endpoints... Initially we wanted to do it backend-side, but we cannot guarantee the spans are in the same chunk so that was problematic... Copying the attack data to the inferred span makes this problem go away.

PR #973, opened 2025-12-29
[SVLS-8231] fixing log requestId for LMI

jchrostek-dd:
https://datadoghq.atlassian.net/browse/SVLS-8231

## Overview
* Bug fix - When running a LMI function, function logs don't get attributed with `lambda.requestId`. As a result, no function logs appear for an invocation. They still show up for the function, just not not on the invocation.
* We could have alternatively updated the UI to search for `lambda.requestId` and `requestId` instead of just searching for `lambda.requestId`. However, I would prefer for On Demand and Lambda Managed Instance Functions to be consistent. So, both using lambda.requestId is preferred.

## Testing 
* Added integration tests that checks for logs per invocation request.
* Also confirmed in the UI we are seeing the logs for an invocation (before/after screenshot below).

Screenshot 2025-12-30 at 8 27 34 AM

Screenshot 2025-12-30 at 8 20 34 AM

### DIFF
--- a/bottlecap/src/logs/lambda/processor.rs
+++ b/bottlecap/src/logs/lambda/processor.rs
@@ -96,7 +96,7 @@ impl LambdaProcessor {
                             obj.get("requestId")
                                 .or_else(|| obj.get("AWSRequestId"))
                                 .and_then(|v| v.as_str())
-                                .map(|s| s.to_string())
+                                .map(ToString::to_string)
                         } else {
                             None
                         };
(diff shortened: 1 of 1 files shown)


bottlecap/src/logs/lambda/processor.rs
```diff
@@ -89,7 +89,44 @@ impl LambdaProcessor {
...
+                    _ => (None, None),
+                };
+
+                if let Some(message) = message {
+                    if is_oom_error(&message) {
+                        debug!("LOGS | Got a runtime-specific OOM error. Incrementing OOM metric.");
+                        if let Err(e) = self.event_bus.send(Event::OutOfMemory(event.time.timestamp())).await {
+                            error!("LOGS | Failed to send OOM event to the main event bus: {e}");
+                        }
+                    }
```
lym953, line 116: I think we can delete the `if is_oom_error(&message) {}` block in the `TelemetryRecord::Extension` case below. I think runtime-specific OOM errors can only come from the function.

lym953 (approved):

bottlecap/src/logs/lambda/processor.rs
jchrostek-dd: Are we able to validate that? I would have though that extensions can also have an OOM error.

jchrostek-dd: Will leave for now to be safe.

PR #974, opened 2025-12-31
[SVLS-8179] add debug log for flush strategy

jchrostek-dd:
## Overview
* Add debug log for flush strategy.
* Add debug log for incoming telemetry records. This excludes Extension telemetry since that would lead to an infinite loop.

## Testing 
* Deployed and invoked function. Confirm we see flush strategy and incoming telemetry records.

### DIFF
--- a/bottlecap/src/logs/lambda/processor.rs
+++ b/bottlecap/src/logs/lambda/processor.rs
@@ -88,6 +88,12 @@ impl LambdaProcessor {
     #[allow(clippy::too_many_lines)]
     async fn get_message(&mut self, event: TelemetryEvent) -> Result<Message, Box<dyn Error>> {
         let copy = event.clone();
+
+        // Log all non-Extension telemetry events to avoid infinite loop
+        if !matches!(event.record, TelemetryRecord::Extension(_)) {
+            debug!("LOGS | Incoming telemetry event: {:?}", event.record);
+        }
+
         match event.record {
             TelemetryRecord::Function(v) => {
                 let (request_id, message) = match v {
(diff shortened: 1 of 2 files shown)


bottlecap/src/logs/lambda/processor.rs
```diff
@@ -88,6 +88,12 @@ impl LambdaProcessor {
     #[allow(clippy::too_many_lines)]
     async fn get_message(&mut self, event: TelemetryEvent) -> Result<Message, Box<dyn Error>> {
         let copy = event.clone();
+
+        // Log all non-Extension telemetry events to avoid infinite loop
+        if !matches!(event.record, TelemetryRecord::Extension(_)) {
+            debug!("LOGS | Incoming telemetry event: {:?}", event.record);
+        }
+
```
lym953, line 96: We are already logging telemetry events here:
https://github.com/DataDog/datadog-lambda-extension/blob/57315b87a0a8765835602629f69faaa58bf81980/bottlecap/src/bin/bottlecap/main.rs#L1213

Is the new log added to help debug what happens in between?

jchrostek-dd: Yeah, helpful for what happens in between. When I was looking at logs for Durable Functions, it seemed like we were missing some things.

astuyve: We should probably revert this. If we’re missing something for durable functions then let’s find out why it’s not being captured and sent to Datadog or via a Datadog flare. This will cause us to double log everything the user sends if debug logs are enabled, which will more than 2x the cost for no value.

Issue #919, opened 2025-11-11
How to disable SSL verification


Dudssource:
## What problem we are facing?

I've been trying to disable SSL verification on v88, as I suspect that this is the problem I'm facing (since we have datadog behind an internal proxy with intermediate TLS certificates).

## Describe the issue

After upgrading one of our golang lambda functions extension from v68 to v88, we noticed that APM data was missing.

After inspecting logs, we noticed one of the issues with HTTP2 protocol for logs (which we solved by adding the undocumented `DD_HTTP_PROTOCOL=http1` env var).

But still we can't get traces and metrics to work. (rollbacking to v68 does fix the issue)

## Error logs

Unfortunately error logs don't return any useful information.

```plaintext
346 | 2025-11-11T10:23:09.789-03:00 | DD_EXTENSION \| ERROR \| TRACES \| Request failed: No requests sent 
347 | 2025-11-11T10:23:09.789-03:00 | DD_EXTENSION \| ERROR \| Max retries exceeded, returning request error error=Network error: client error (Connect) attempts=1

```

Any help would be highly appreciated.

Thanks!

Dudssource:
After a few attempts I was able to collect more info about the error (indeed it seems like SSL check is not being disabled, even with the right env vars)

```plaintext
kind: InvalidData, error: InvalidCertificate(UnknownIssuer) 
```

Dudssource:
I'm a bit worried about this one, after digging the extension code, I really believe that the flags `DD_SKIP_SSL_VALIDATION` and `DD_LOGS_CONFIG_LOGS_NO_SSL` were not implemented at all on the new version of the extension

lym953:
@Dudssource Thanks for the test! Seems some changes are needed on our side. Let me track this work internally.
To help us reproduce the issue: What kind of proxy are you using, e.g. NGINX or squid?

PR #979, opened 2026-01-07
[SVLS-7945] feat: Support TLS certificate for logs/proxy flusher

lym953:
## Problem
A customer reported that their Lambda is behind a proxy, and the Rust-based extension encounters an error when sending logs and metrics to Datadog via the proxy.

A previous PR https://github.com/DataDog/datadog-lambda-extension/pull/961 fixed this for traces and stats, but not for other things because the customer and I didn't see any error with them at that time.

## This PR

Applies the env var `DD_TLS_CERT_FILE` to logs flusher and proxy flusher as well.
Example: `DD_TLS_CERT_FILE=/opt/ca-cert.pem`, so the when the extension flushes logs or proxied data to Datadog, the HTTP client created can load and use this cert, and connect the proxy properly.

## Testing
1. Create a Lambda in a VPC with a proxy EC2 instance.
2. Connect to the proxy instance. With the help of ChatGPT, set up a custom-build nginx with `ngx_http_proxy_connect_module`
3. Save the CA certificate from the proxy server to local machine
4. In the CDK stack, add a layer to the Lambda, which includes the CA certificate `ca-cert.pem`
5. Set env vars:
    - `DD_TLS_CERT_FILE=/opt/ca-cert.pem`
    - `DD_PROXY_HTTPS=http://10.0.0.30:3128`, where `10.0.0.30` is the private IP of the proxy EC2 instance
    - `DD_LOG_LEVEL=debug`
6. Invoke the Lambda

## Result
**Before:**
Log flushing failed:
> DD_EXTENSION | ERROR | LOGS | Failed to send request after 97 ms and 3 attempts: reqwest::Error { kind: Request, url: "https://http-intake.logs.datadoghq.com/api/v2/logs", source: hyper_util::client::legacy::Error(Connect, ConnectFailed(Custom { kind: Other, error: Custom { kind: InvalidData, error: InvalidCertificate(UnknownIssuer) } })) }

**After:**
No such error

## Next steps
Do the same thing for dogstatsd metric flusher. Metric flusher is in a separate repo https://github.com/DataDog/serverless-components, so let's create separate PRs for that change.

## Notes
Customer report issue: https://github.com/DataDog/datadog-lambda-extension/issues/919

### DIFF
--- a/bottlecap/src/http.rs
+++ b/bottlecap/src/http.rs
@@ -90,12 +90,10 @@ fn load_custom_cert(cert_path: &str) -> Result<Vec<reqwest::Certificate>, Box<dy
     }
 
     // Convert all certificates found in the file
-    let mut reqwest_certs = Vec::new();
-    for cert in certs {
-        reqwest_certs.push(reqwest::Certificate::from_der(&cert)?);
-    }
-
-    Ok(reqwest_certs)
+    certs
+        .into_iter()
+        .map(|cert| reqwest::Certificate::from_der(&cert).map_err(Into::into))
+        .collect()
 }
 
 pub async fn handler_not_found() -> Response {


bottlecap/src/http.rs
```diff
@@ -47,6 +47,28 @@ fn build_client(config: &Arc<config::Config>) -> Result<reqwest::Client, Box<dyn
             .http2_keep_alive_timeout(Duration::from_secs(1000));
     }
 
+    // Load custom TLS certificate if configured
+    if let Some(cert_path) = &config.tls_cert_file {
```
duncanista, line 51: file and path are confusing, can a customer load more than 1 certificate or define a folder?

duncanista: nvm, ignore this, a path can be to a file, but again, my question persist, can a customer load more than 1 cert?

bottlecap/src/http.rs
```diff
@@ -56,6 +78,26 @@ fn build_client(config: &Arc<config::Config>) -> Result<reqwest::Client, Box<dyn
...
+    let mut reader = BufReader::new(file);
+
+    // Parse PEM certificates
+    let certs = rustls_pemfile::certs(&mut reader).collect::<Result<Vec<_>, _>>()?;
+
+    if certs.is_empty() {
+        return Err("No certificates found in file".into());
+    }
+
+    // Convert all certificates found in the file
```
duncanista, line 92: Can a file contain multiple certificates? This confuses me

bottlecap/src/http.rs
```diff
@@ -56,6 +78,26 @@ fn build_client(config: &Arc<config::Config>) -> Result<reqwest::Client, Box<dyn
...
+    let mut reqwest_certs = Vec::new();
+    for cert in certs {
+        reqwest_certs.push(reqwest::Certificate::from_der(&cert)?);
+    }
```
duncanista: Is there a more idiomatic way of doing this?

duncanista (approved):
Great job on fixing this! 🙇🏽

bottlecap/src/http.rs
lym953: Good point. Changed to `map()`.

bottlecap/src/http.rs
lym953: A pem file can contain multiple certificates:
https://stackoverflow.com/questions/68340665/pem-file-has-two-certificates-what-does-it-mean

And we supported multiple certs for trace and stats flusher: https://github.com/DataDog/datadog-lambda-extension/blob/main/bottlecap/src/traces/trace_flusher.rs#L247-L250

Not sure if there's a real use case, but since it's not hard, I'm supporting it in this PR.

yues7: Hi! Appreciate the fix, our customer reached out in the support ticket that they do have many CA bundled in one .crt so the use case is valid. They also share that the PR has fixed the issue for them so really appreciate it, thanks a lot!

lym953: @yues7 Nice!

PR #986, opened 2026-01-09
[SVLS-8211] feat: Add timeout for requests to span_dedup_service

lym953:
## Problem
Span dedup service sometimes fails to return the result and thus logs the error:
> DD_EXTENSION | ERROR | Failed to send check_and_add response: true

I see this error in our Self Monitoring and a customer's account.
Also I believe it causes extension to fail to receive traces from the tracer, causing missing traces. This is because the caller of span dedup is in `process_traces()`, which is the function that handles the tracer's HTTP request to send traces. If this function fails to get span dedup result and gets stuck, the HTTP request will time out.

## This PR
While I don't yet know what causes the error, this PR adds a patch to mitigate the impact:
1. Change log level from `error` to `warn`
2. Add a timeout of 5 seconds to the span dedup check, so that if the caller doesn't get an answer soon, it defaults to treating the trace as not a duplicate, which is the most common case.

## Testing
To merge this PR then check log in self monitoring, as it's hard to run high-volume tests in self monitoring from a non-main branch.

### DIFF
--- a/bottlecap/src/traces/span_dedup_service.rs
+++ b/bottlecap/src/traces/span_dedup_service.rs
@@ -130,17 +130,17 @@ mod tests {
         let key2 = DedupKey::new(100, 456);
 
         // First call should return true (key was added)
-        assert!(handle.check_and_add(key1).await.unwrap());
+        assert!(handle.check_and_add(key1, None).await.unwrap());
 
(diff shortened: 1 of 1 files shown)


litianningdatadog (approved):
LGTM

PR #1006, opened 2026-01-23
integration test cleanup function

jchrostek-dd:
## Overview
* Personal stacks used for integ testing don't always get cleaned up. We had a lot of LMI stacks that was causing us to hit our Capacity Provider limit and not deploy new integ stacks for testing.
* This PR creates a lambda function that runs daily and deletes integration stacks that haven't been updated in 7 days.

## Testing 
* Deployed function and invoked it. Confirmed that it deleted old stacks.


duncanista (commented):
Feels like an unnecessary project when we could just have a lambda that cleans all resources in our account which might not be needed – or projects which failed to delete

PR #1011, opened 2026-01-28
[APMSVLS-197] Add check to extract trace context present within event.request.headers

Dogbu-cyber:
# [APMSVLS-197] feat: Add check to extract trace context from event.request.headers

## Overview

This PR adds automatic trace context extraction from `event.request.headers` for AppSync integration scenarios. This eliminates the need for customers to use customized extractors for RUM → AppSync → Lambda resolver flows where RUM-injected trace context is nested under `event["request"]["headers"]`.

## Problem

Previously, trace context from RUM was not propagated through AppSync to Lambda resolvers because the extension only checked:
1. Trigger-specific headers (via span inferrer)
2. `event.headers` (top-level event headers)
3. Request headers (HTTP request headers)

For AppSync, the trace context is nested under `event.request.headers`, which was not checked, causing traces to break at the AppSync boundary.

## Solution

Added a step that checks `event.request.headers` after all other trace context sources have been attempted.

### Before 
Without this change, trace context is lost at the AppSync boundary:

[Trace showing broken propagation through AppSync](https://ddserverless.datadoghq.com/apm/trace/5616290163800553628?graphType=waterfall&shouldShowLegend=true&spanID=5616290163800553628&timeHint=1769698078685.9998&traceQuery=)

[Log showing the passed header value for x-datadog-trace-id does not align with the given trace_id](https://ddserverless.datadoghq.com/logs?query=%40lambda.request_id%3A%28cba225c5-7b76-45d8-9a43-86df9031687f%29%20OR%20trace_id%3A5616290163800553628&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AwAAAZwKObH6tpUUmwAAABhBWndLT2JqNEFBQWpWdUZHSkF5YzFRQXgAAAAkZjE5YzBhNDQtYjE3Zi00YjljLTk2NzMtNWY2NzI2OTc2MDlmAA2mrg&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=flex_tier&stream_sort=desc&viz=stream&from_ts=1769697078981&to_ts=1769699079405&live=false)

Flow: `RUM → AppSync → Lambda` (trace ID not propagated)

### After
With this change, trace context is properly extracted and propagated:

[Trace showing successful propagation through AppSync](https://ddserverless.datadoghq.com/apm/trace/12345?graphType=waterfall&shouldShowLegend=true&spanID=7582932037133874185&timeHint=1769698399447&traceQuery=)

Flow: `RUM → AppSync → Lambda resolver` (trace continues using value within x-datadog-trace-id)

### Unit Tests

Added 10 unit tests covering:

- Extraction from `event.request.headers` with Datadog headers
- Extraction from `event.request.headers` with W3C TraceContext headers
- Edge cases (missing headers, empty objects, invalid data)

Run tests with:
```bash
cargo test test_extract_span_context -- --nocapture
```

[APMSVLS-197]: https://datadoghq.atlassian.net/browse/APMSVLS-197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -1076,13 +1076,13 @@ impl Processor {
             return Some(sc);
         }
 
-        if let Some(request_obj) = payload_value.get("request") {
-            if let Some(request_headers) = request_obj.get("headers") {
-                if let Some(sc) = propagator.extract(request_headers) {
-                    debug!("Extracted trace context from event.request.headers");
-                    return Some(sc);
-                }
-            }
+        if let Some(sc) = payload_value
+            .get("request")
+            .and_then(|req| req.get("headers"))
+            .and_then(|headers| propagator.extract(headers))
+        {
+            debug!("Extracted trace context from event.request.headers");
+            return Some(sc);
         }
 
         if let Some(payload_headers) = payload_value.get("headers") {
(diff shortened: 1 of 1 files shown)


bottlecap/src/lifecycle/invocation/processor.rs
duncanista, line 1: Could you add some unit testing and show an example trace in the PR description? Just to see what were we missing

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -1074,6 +1074,15 @@ impl Processor {
...
 
+        if let Some(request_obj) = payload_value.get("request") {
+            if let Some(request_headers) = request_obj.get("headers") {
+                if let Some(sc) = propagator.extract(request_headers) {
+                    debug!("Extracted trace context from event.request.headers");
+                    return Some(sc);
+                }
+            }
+        }
+
```
duncanista: Why did you decided that this should be last priority?

Dogbu-cyber: I assumed there was some reason for the original order, so I didn't take a stance on where exactly it should go and added it as the last case..

duncanista: Based on other projects which support AppSync extraction, it seems that they are considered in the same order as inferred spans would do it – I'd suggest to follow that same order

https://github.com/DataDog/datadog-lambda-js/blob/main/src/trace/context/extractor.ts#L92

Dogbu-cyber: Thanks! That's a good point.

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -1838,4 +1847,298 @@ mod tests {
...
+        });
+
+        let result = Processor::extract_span_context(&headers, &payload, propagator);
+
+        assert!(result.is_some());
+        let context = result.unwrap();
+        assert_eq!(
+            context.trace_id, 555,
+            "Should prioritize event.request.headers as service-specific extraction"
+        );
```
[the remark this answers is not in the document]
Dogbu-cyber: This is expected behavior? 
Extraction from managed services (including Appsync) is higher priority than trace context within event.headers.

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -1063,6 +1063,15 @@ impl Processor {
             return Some(sc);
         }
 
+        if let Some(request_obj) = payload_value.get("request") {
+            if let Some(request_headers) = request_obj.get("headers") {
```
duncanista: Can be simplified instead of having two nested if let somes to and_then

duncanista: Maybe even with the third one, depending on what clippy says

bottlecap/src/lifecycle/invocation/processor.rs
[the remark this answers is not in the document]
duncanista: Maybe it just bugged due to the changes

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -1838,4 +1847,298 @@ mod tests {
...
+        let result = Processor::extract_span_context(&headers, &payload, propagator);
+
+        assert!(
+            result.is_none(),
+            "Should return None when headers are invalid"
+        );
+    }
+
+    #[test]
+    fn test_extract_span_context_appsync_real_world_example() {
```
duncanista: bad naming with "real_world_example" instead, either set all trace context values to be something useful

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -1838,4 +1847,298 @@ mod tests {
...
+                    "x-datadog-tags": "_dd.p.dm=-0",
+                    "user-agent": "Mozilla/5.0",
+                    "content-type": "application/json"
+                }
+            },
+            "info": {
+                "fieldName": "getUser",
+                "parentTypeName": "Query"
+            }
+        });
```
duncanista: Could be part of the payloads we have in testing as a file

PR #1016, opened 2026-02-04
chore(flushing): refactor flushing

duncanista:
## Overview

Refactors flushing to become its own service

## Motivation

Making it easier for us to handle the flushing code in one place and ensure consistency
[SVLS-8505](https://datadoghq.atlassian.net/browse/SVLS-8505)

## Notes

Not using dynamic typings for the flushers we accepts so far, in another PR I will standardize it and ensure we can have capacity-based flushing too.

[SVLS-8505]: https://datadoghq.atlassian.net/browse/SVLS-8505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -750,6 +750,9 @@ async fn extension_loop_active(
                 .flush_blocking(true, &mut locked_metrics)
                 .await;
 
+            // Even though we're shutting down, we need to reset the flush interval to prevent any future flushes
+            race_flush_interval.reset();
+
             // Shutdown aggregator services
             if let Err(e) = logs_aggregator_handle.shutdown() {
                 error!("Failed to shutdown logs aggregator: {e}");


bottlecap/src/flushing/service.rs
```diff
@@ -0,0 +1,366 @@
...
+                }
+            });
+            self.handles.metric_flush_handles.push(handle);
+        }
+
+        // Spawn stats flush (fire-and-forget, no retry)
+        let sf = Arc::clone(&self.stats_flusher);
+        self.handles
+            .stats_flush_handles
+            .push(tokio::spawn(async move { sf.flush(false).await }));
```
duncanista, line 142: Need your review here @lym953

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -1046,19 +744,11 @@ async fn extension_loop_active(
...
-                &*stats_flusher,
-                &proxy_flusher,
-                Some(&mut race_flush_interval),
-                &metrics_aggregator_handle,
-                true, // force_flush_trace_stats
-            )
-            .await;
+            flushing_service
+                .flush_blocking(true, &mut locked_metrics)
+                .await;
```
litianningdatadog, line 751: Do we need to reset race_flush_interval as well?

duncanista: It's the last flush, technically we don't need it, but I'll add it nonetheless

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -817,48 +589,12 @@ async fn extension_loop_active(
             &lifecycle_listener_shutdown_token,
         );
 
-        // Wait for background flusher to complete gracefully
+        // Wait for background flusher to complete gracefully.
+        // The background task performs the final flush before exiting, so we just need to wait.
         if let Err(e) = flush_task_handle.await {
             error!("Error waiting for background flush task: {e:?}");
         }
```
litianningdatadog, line 597: Is there a particular reason we no longer flush on shutdown? If so, would you mind adding a brief comment in the commit to capture the rationale?

bottlecap/src/bin/bottlecap/main.rs
duncanista: Done

bottlecap/src/bin/bottlecap/main.rs
duncanista: It got moved, once we await the flush_task_handle, since it's the last task we wait for, inside of it, it just does its final flush on token cancellation

https://github.com/DataDog/datadog-lambda-extension/pull/1016/changes#diff-0b46a0bd1733b354850b895657f37a662f9a30cc2fa0ae684dfa1655520a53aaR451-R457

easier and cleaner than having it after we already awaited for our flushing task

PR #1021, opened 2026-02-06
[SVLS-8150] fix(config): ensure logs intake URL is correctly prefixed

shreyamalpani:
## Overview

Ensures `DD_LOGS_CONFIG_LOGS_DD_URL` is correctly prefixed with `https://`

## Testing 

Manually tested that logs get sent to alternate logs intake


duncanista (approved):
Approving to unblock, ensure to get latest main rebased so audit doesn't complain

PR #1022, opened 2026-02-09
publish image to correct repos

jchrostek-dd:
## Overview
We don't publish to the correct repos.

https://dd.slack.com/archives/CBWDFKWV8/p1770042159728849


duncanista (approved):
LGTM – might be good to wait on Ali to check whether we want to publish to Azure or not

PR #1034, opened 2026-02-19
chore(deps): upgrade rust to `v1.93.1`

duncanista:
## What?

Upgrade rust to latest stable 1.93.1

## Why?

`time` vulnerability fix is only available on rust >= 1.88.0


bottlecap/src/traces/propagation/text_map_propagator.rs
duncanista, line 1: Eventually want to replace this with the propagator in dd-trace-rs

Issue #807, opened 2025-08-27
Support for AWS JSON log level format


artkhod:
According to https://docs.datadoghq.com/logs/guide/logs-show-info-status-for-warnings-or-errors/ we can remap status from let's say "level" or include status attribute ourselves.

https://github.com/DataDog/datadog-lambda-extension/blob/0e9fc6fed652fa129d2a826604b87307ffda15a4/bottlecap/src/logs/lambda/mod.rs#L55

However, since it's a lambda extension, it would make sense to support AWS JSON Log format

```json 
{
  "timestamp": "2025-08-27T10:25:22.244Z",
  "level": "WARN", <--- level is uppercase, and it drives the log level setting on the lambda configuration itself
  "message": "AWS Lambda plans to remove support for callback-based function handlers starting with Node.js 24. You will need to update this function to use an async handler to use Node.js 24 or later. For more information and to provide feedback on this change, see https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/issues/137. To disable this warning, set the AWS_LAMBDA_NODEJS_DISABLE_CALLBACK_WARNING environment variable."
}
```
Here's the list of levels https://github.com/aws-powertools/powertools-lambda-typescript/blob/801333deff04b09d4adb88860c60f497295685d9/packages/logger/src/constants.ts#L17 

JSON log format can be set here

<img width="1599" height="552" alt="Image" src="https://github.com/user-attachments/assets/c1e3369b-7291-40ad-ae0d-4825f78818bf" />

and it's described by this section https://docs.aws.amazon.com/lambda/latest/dg/monitoring-cloudwatchlogs-logformat.html#monitoring-cloudwatchlogs-JSON-application

PR #1036, opened 2026-02-20
fix(logs): format log level into correct status

duncanista:
## Overview

Log level was not properly formatted for AWS JSON logs – same goes for our own logs where the level was not truly respected.

#807 and [SVLS-8590](https://datadoghq.atlassian.net/browse/SVLS-8590)

## Testing 

Unit tests and on actual resources.

Before logs with `debug` appeared as `info`
Screenshot 2026-02-19 at 7 17 12 PM

Now, all appear with their right level:

Screenshot 2026-02-19 at 7 17 55 PM

[SVLS-8590]: https://datadoghq.atlassian.net/browse/SVLS-8590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/logger.rs
+++ b/bottlecap/src/logger.rs
@@ -180,7 +180,7 @@ mod tests {
             "quotes should be escaped in raw JSON"
         );
         assert!(
-            output.contains(r#"\n"#),
+            output.contains(r"\n"),
             "newline should be escaped in raw JSON"
         );
 


bottlecap/src/logger.rs
```diff
@@ -20,36 +37,147 @@ where
...
+        let output = capture_log(|| {
+            tracing::info!("message with \"quotes\" and a\nnewline");
+        });
 
-        writeln!(writer)
+        let parsed: serde_json::Value =
+            serde_json::from_str(output.trim()).expect("special chars should be escaped");
+        let msg = parsed["message"].as_str().unwrap();
+        assert!(msg.contains("quotes"));
+        assert!(msg.contains("newline"));
```
lym953: Is the test missing that ensures formatter really escapes special characters?

bottlecap/src/logs/lambda/processor.rs
```diff
@@ -55,6 +55,21 @@ fn is_oom_error(error_msg: &str) -> bool {
...
+/// Case-insensitive and accepts both short and long forms
+/// (e.g. "WARN"/"WARNING", "INFO"/"INFORMATION", "ERR"/"ERROR").
+/// Returns `None` for unrecognized levels so callers can fall back to a default.
+fn map_log_level_to_status(level: &str) -> Option<&'static str> {
+    match level.to_uppercase().as_str() {
+        "FATAL" | "CRITICAL" => Some("critical"),
+        "ERROR" | "ERR" => Some("error"),
+        "WARN" | "WARNING" => Some("warn"),
+        "INFO" | "INFORMATION" => Some("info"),
+        "DEBUG" | "TRACE" => Some("debug"),
```
lym953: Don't we support "trace" level?

duncanista: Seems like we do! Let me update this, at first I thought we didn't

bottlecap/src/logger.rs
duncanista: I will change this so its more clear and shows that the text doesn't include the `\"` or `\n`

duncanista: Updated!

lym953 (approved):

PR #1039, opened 2026-02-23
fix(deps): upgrade libdatadog 

duncanista:
## Overview

- Upgrade libdatadog dependencies from 158b594 to c8121f4 (v28), adapting to upstream API changes (hyper_migration → http_common, HyperClient → HttpClient, new use_system_resolver field on Endpoint, SendData no
  longer Clone)
  - Rename hyper_client module to http_client to match the upstream naming change
  - Fix dual-shipping to additional endpoints: previously TracerHeaderTags::default() was used (losing all tracer metadata headers like datadog-meta-lang, datadog-meta-tracer-version, etc.). Now the real header
  tags are preserved through the aggregation pipeline and used for additional endpoints
  - Avoid unnecessary full TracerPayloadCollection clone in trace_processor by only cloning inner V07 payloads for stats generation

## Changes

  - Bump all libdd-* crates to rev c8121f4 in Cargo.toml / Cargo.lock
  - Adapt to hyper_migration → http_common rename across trace_agent, stats_processor, stats_flusher, and http_client
  - Add use_system_resolver: false to all Endpoint constructions (new required field)

  hyper_client.rs → http_client.rs

  - Rename module and type alias (HyperClient → HttpClient)
  - Update all consumers (trace_flusher, stats_flusher, mod.rs)

  Trace flusher refactor

  - Adapt to SendData no longer being Clone: construct per-endpoint SendData objects by cloning inner V07 payloads instead of cloning the whole SendData
  - Remove override_endpoint parameter from send_traces — each SendData now carries its own target endpoint
  - Simplify retry logic comments (endpoint tracking is no longer relevant)

  Preserve TracerHeaderTags for additional endpoints

  - Introduce OwnedTracerHeaderTags — an owned version of TracerHeaderTags<'a> that can be stored across async boundaries
  - Add header_tags field to SendDataBuilderInfo and thread it through the aggregation pipeline
  - Use real header tags (via to_tracer_header_tags()) when constructing SendData for additional endpoints instead of TracerHeaderTags::default()

  Trace processor optimization

  - Clone only inner V07 payloads for stats generation instead of cloning the entire TracerPayloadCollection before building SendData

##  Tests
  - cargo check passes
  - cargo test --lib passes (481 tests)
  - Verify dual-shipped traces to additional endpoints include correct datadog-meta-* headers with DD_APM_ADDITIONAL_ENDPOINTS configured

### DIFF
--- a/bottlecap/src/traces/trace_aggregator.rs
+++ b/bottlecap/src/traces/trace_aggregator.rs
@@ -41,7 +41,7 @@ impl From<TracerHeaderTags<'_>> for OwnedTracerHeaderTags {
 }
 
 impl OwnedTracerHeaderTags {
-    #[must_use] 
+    #[must_use]
     pub fn to_tracer_header_tags(&self) -> TracerHeaderTags<'_> {
         TracerHeaderTags {
             lang: &self.lang,
@@ -174,7 +174,11 @@ mod tests {
             tracer_header_tags.clone(),
             &Endpoint::from_slice("localhost"),
         );
-        SendDataBuilderInfo::new(builder, size, OwnedTracerHeaderTags::from(tracer_header_tags))
+        SendDataBuilderInfo::new(
+            builder,
+            size,
+            OwnedTracerHeaderTags::from(tracer_header_tags),
+        )
     }
 
     #[test]


bottlecap/src/traces/http_client.rs
```diff
@@ -20,7 +20,7 @@ use tracing::debug;
 /// Type alias for the HTTP client used by trace and stats flushers.
 ///
 /// This is the client type expected by `libdd_trace_utils::SendData::send()`.
-pub type HyperClient =
```
jchrostek-dd, line 23: This isn't an upgrade is it? This is a change in libraries we are using?

duncanista: libdatadog switches away from having to having a dependency on hyper clients as its expecting to use reqwest instead, what we normally use everywhere else here, this just renames the previous client which was hyper tied to just be a "generic" type `HttpClient` because we are not concerned of what it is, and they're exporting it as an http helper library

bottlecap/src/traces/trace_aggregator.rs
```diff
@@ -110,42 +108,29 @@ mod tests {
...
+        let size = 1;
+
+        aggregator.add(SendDataBuilderInfo::new(make_builder(size), size));
         assert_eq!(aggregator.queue.len(), 1);
     }
 
     #[test]
     fn test_get_batch() {
         let mut aggregator = TraceAggregator::default();
-        let tracer_header_tags = TracerHeaderTags {
```
jchrostek-dd, line 128: Could you include more details in the PR description. Not clear why we are having to update test cases like this is we are just doing an upgrade to dependencies. Seems like there are other changes like this as well which I wouldn't expect if we are just upgrade dependencies.

duncanista: Seems like on this test update, it just simplified data creation for easier testing, seems like we were repeating the same code over and over – although some of these deps indeed changed as they're breaking so we might see some test updates just due to the nature of it (not this one tho)

bottlecap/src/traces/trace_flusher.rs
```diff
@@ -126,26 +124,32 @@ impl TraceFlusher {
...
-                let traces_clone = traces.clone();
+                let additional_traces: Vec<_> = traces
+                    .iter()
+                    .filter_map(|trace| match trace.get_payloads() {
+                        TracerPayloadCollection::V07(payloads) => Some(SendData::new(
+                            trace.len(),
+                            TracerPayloadCollection::V07(payloads.clone()),
+                            TracerHeaderTags::default(),
+                            &endpoint,
+                        )),
```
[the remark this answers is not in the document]
duncanista: Resolved

PR #1042, opened 2026-02-24
perf(logger): replace serde_json::json! with zero-allocation write! in logger

litianningdatadog:
## Summary 
Root cause: commit 3b533a0 introduced `serde_json::json!()` for JSON formatting, adding per-call heap pressure and a new crate dependency that contributes to Lambda init duration regression

JIRA: https://datadoghq.atlassian.net/browse/SVLS-8619

- Replaces `serde_json::json!()` in `logger.rs` with direct `write!`/`writeln!` calls to the tracing `Writer`, eliminating two heap allocations (a `Map` + `Value`) per log call
- Adds a `write_json_escaped` helper that handles all 6 mandatory JSON escape sequences inline (`"`, `\`, `\n`, `\r`, `\t`, U+0000–U+001F)
- Removes the `"json"` feature from `tracing-subscriber`, dropping `tracing-serde` as a transitive dependency and reducing binary size
- The JSON output format is **identical** — no observable behavior change

## Test plan
- unit tests
- self-monitoring tests

### DIFF
--- a/bottlecap/src/logger.rs
+++ b/bottlecap/src/logger.rs
@@ -6,6 +6,23 @@ use tracing_subscriber::fmt::{
 };
 use tracing_subscriber::registry::LookupSpan;
 
+/// Writes `s` to `w` with the 6 mandatory JSON string escape sequences applied.
+/// Handles: `"`, `\`, `\n`, `\r`, `\t`, and U+0000–U+001F control characters.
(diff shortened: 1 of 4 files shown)


bottlecap/src/logger.rs
```diff
@@ -116,6 +129,55 @@ mod tests {
         }
     }
 
+    fn escaped(s: &str) -> String {
```
lym953, line 132: Where is this used? Only in tests?

Issue #1041, opened 2026-02-23
DD_EXTENSION | ERROR | Failed to extract request body: Failed to buffer the request body: length limit exceeded for large invocation payloads (>3 MB)


oulusoy:
### Description

Lambda functions that receive large invocation payloads (e.g. KafkaEvents >3 MB via MSK Event Source Mappings) produce the following error:

```
DD_EXTENSION | ERROR | Failed to extract request body: Failed to buffer the request body: length limit exceeded
```

The error originates from the Rust-based next-gen extension agent. The older Go-based agent does not have this issue.

### Environment

- **Runtime:** Java 21 (`public.ecr.aws/lambda/java:21`)
- **Extension version:** v92 (`public.ecr.aws/datadog/lambda-extension:92`)
- **Java tracer:** `dd-java-agent.jar` (latest)
- **Event source:** MSK (Kafka) Event Source Mapping delivering KafkaEvent payloads of 3-6 MB

### Reproduction

Invoke any Lambda function with a payload larger than ~3 MB while `DD_TRACE_ENABLED=true`:

```bash
aws lambda invoke --function-name my-function --payload file://large-kafka-event.json out.json
```

The function executes successfully, but the extension logs:
```
DD_EXTENSION | ERROR | Failed to extract request body: Failed to buffer the request body: length limit exceeded
```

### Distinction from #899

Issue #899 was fixed in v89 by increasing the **trace payload limit** from 2 MiB to 50 MiB. That limit controls the communication between the language tracer and the extension's internal trace agent.

This issue is about a **different limit**: the extension's internal buffer for the **Lambda invocation event body**. This limit remains at ~3 MB and was not addressed by the v89 fix.

### Investigation

**Environment variables — none had any effect:**

| Variable | Value | Result |
| --- | --- | --- |
| `DD_UNIVERSAL_INSTRUMENTATION` | `false` | Error persists |
| `DD_CAPTURE_LAMBDA_PAYLOAD` | `false` | Error persists |
| `DD_EXPERIMENTAL_ENABLE_PROXY` | `true` | Error persists |
| `DD_SERVERLESS_APPSEC_ENABLED` | `true` | Error persists |

**Extension versions — only Go-based agents are unaffected:**

| Version | Agent | Mode | 5 MB payload |
| --- | --- | --- | --- |
| v65 | Go | default | No error |
| v87 | Go+Rust | `DD_EXTENSION_VERSION=compatibility` | No error |
[...]

PR #1044, opened 2026-02-24
fix(lifecycle): raise axum body limit to 6 MB for large Lambda payloads

litianningdatadog:
https://datadoghq.atlassian.net/browse/SVLS-8609
Issue reported: https://github.com/DataDog/datadog-lambda-extension/issues/1041
## Summary

- axum 0.7+ applies a 2 MB default body limit globally; the `/lambda/start-invocation` endpoint was rejecting payloads larger than 2 MB with `length limit exceeded`
- Raises the limit to 6 MB (`DefaultBodyLimit::max(6 * 1024 * 1024)`) to match [Lambda's maximum synchronous invocation payload size](https://docs.aws.amazon.com/lambda/latest/api/API_Invoke.html)
- Adds three unit tests covering: accept above the old 2 MB default, accept at boundary (6 MB − 1 byte), reject above 6 MB (413)

## Test plan
- Unit tests
- Run `bash local_tests/repro-large-payload.sh` which sets up the local test env and sends a 3 MB payload to `/lambda/start-invocation` and verifies HTTP 200 with no `length limit exceeded` error

| Scenario|Log| 
|--------|--------|
| Without the fix|```{"level":"ERROR","message":"DD_EXTENSION \| ERROR \| Failed to extract request body: Failed to buffer the request body: length limit exceeded"}``` |
| With the fix| ```{"level":"DEBUG","message":"DD_EXTENSION \| DEBUG \| Received start invocation request from headers:{\"datadog-meta-lang\": \"java\", \"user-agent\": \"curl/8.3.0\", \"host\": \"localhost:8124\", \"lambda-runtime-aws-request-id\": \"test-large-payload-request\", \"expect\": \"100-continue\", \"accept\": \"*/*\", \"content-length\": \"3200074\", \"content-type\": \"application/json\"}"}``` |

### DIFF
--- a/bottlecap/Cargo.lock
+++ b/bottlecap/Cargo.lock
@@ -3786,16 +3786,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "tracing-serde"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "704b1aeb7be0d0a84fc9828cae51dab5970fee5088f83d1dd7ee6f6246fc6ff1"
-dependencies = [
- "serde",
- "tracing-core",
-]
-
-[[package]]
 name = "tracing-subscriber"
 version = "0.3.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3804,14 +3794,11 @@ dependencies = [
  "matchers",
  "once_cell",
  "regex-automata",
- "serde",
- "serde_json",
  "sharded-slab",
  "thread_local",
  "tracing",
  "tracing-core",
  "tracing-log",
- "tracing-serde",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


bottlecap/src/lifecycle/listener.rs
```diff
@@ -37,6 +37,9 @@ const HELLO_PATH: &str = "/lambda/hello";
 const START_INVOCATION_PATH: &str = "/lambda/start-invocation";
 const END_INVOCATION_PATH: &str = "/lambda/end-invocation";
 const AGENT_PORT: usize = 8124;
+// Lambda's maximum synchronous invocation payload size
+// reference: https://docs.aws.amazon.com/lambda/latest/api/API_Invoke.html
+const LAMBDA_INVOCATION_MAX_PAYLOAD: usize = 6 * 1024 * 1024;
```
duncanista, line 42: Wondering how would this look like with streaming lambda payload... do we have guards to not send the payload if it goes over the limit?

duncanista: Because in streaming the max is 200MB

duncanista: I mean good for now, but we might want to leave a comment about streaming

PR #1048, opened 2026-02-25
[SVLS-8581] feat: Add durable_function:true tag to enhanced metrics

lym953:
## Overview

When the Lambda function's runtime version contains "RuntimeVersion", add the tag `durable_function:true` to enhanced metrics generated by the extension, so the facet `durable_function` appears on Serverless View.

Sample value of runtime version:
- Python: `python:3.14.DurableFunction.v9`
- Node.js: `nodejs:24.DurableFunction.v10`

### Steps
1. Build a layer and install it on a durable function
2. Invoke it
3. Check Serverless view
### Result
1. When I search by `durable_function:true`, the function appears in the result.
image

2. A facet `durable_function` appears on the sidebar.
image

3. When I search by `-durable_function:true`, other functions appear in the result.
image

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -285,11 +285,8 @@ impl Processor {
     /// This is used to create a cold start span, since this telemetry event does not
     /// provide a `request_id`, we try to guess which invocation is the cold start.
     pub fn on_platform_init_start(&mut self, time: DateTime<Utc>, runtime_version: Option<String>) {
-        if let Some(ref rv) = runtime_version {
-            debug!("Runtime version: {}", rv);
-            if rv.contains("DurableFunction") {
-                self.enhanced_metrics.set_durable_function_tag();
-            }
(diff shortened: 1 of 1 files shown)


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -284,7 +284,13 @@ impl Processor {
     ///
     /// This is used to create a cold start span, since this telemetry event does not
     /// provide a `request_id`, we try to guess which invocation is the cold start.
-    pub fn on_platform_init_start(&mut self, time: DateTime<Utc>) {
+    pub fn on_platform_init_start(&mut self, time: DateTime<Utc>, runtime_version: Option<String>) {
+        if let Some(ref rv) = runtime_version {
+            debug!("Runtime version: {}", rv);
```
duncanista: Is this debug log required? It always appear at the top of the platform init start

lym953: I don't have a strong opinion. I added this in case AWS changes the format one day. But that should be able to be captured by e2e tests when they are ready. Let me remove it then.

PR #1190, opened 2026-04-17
fix: update Cargo.lock to fix cargo audit vulnerabilities

lym953:
## Summary

- Update `rustls-webpki` 0.103.10 → 0.103.12 (fixes RUSTSEC-2026-0098 and RUSTSEC-2026-0099: name constraint bypass vulnerabilities in TLS certificate validation)
- Update `rand` 0.8.5 → 0.8.6 and 0.9.2 → 0.9.4 (fixes RUSTSEC-2026-0097: unsound aliased mutable reference when using custom loggers)

These were causing the [cargo audit CI job to fail](https://github.com/DataDog/datadog-lambda-extension/actions/runs/24472752003/job/71517081151).

Only `Cargo.lock` is updated (no `Cargo.toml` changes needed — these are transitive dependency version bumps).

### DIFF
--- a/bottlecap/Cargo.lock
+++ b/bottlecap/Cargo.lock
@@ -392,7 +392,7 @@ dependencies = [
  "bitflags 2.11.0",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.11.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -515,7 +515,7 @@ dependencies = [
  "ordered_hash_map",
  "proptest",
  "prost 0.14.3",
- "rand 0.8.5",
+ "rand 0.8.6",
  "regex",
  "reqwest",
  "rustls",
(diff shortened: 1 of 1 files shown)


bottlecap/Cargo.lock
```diff
@@ -392,7 +392,7 @@ dependencies = [
  "bitflags 2.11.0",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.11.0",
```
litianningdatadog, line 395: it is downgraded?

litianningdatadog (commented):
Left a comment.

bottlecap/Cargo.lock
lym953: It's a consequence of running `cargo update rustls-webpki`

From Claude:
> The chain is:                                                                                
  rustls-webpki → aws-lc-rs → aws-lc-fips-sys → bindgen → itertools                                      
  When cargo updates rustls-webpki, it re-resolves this entire chain. Bindgen's constraint is            
  >=0.10,<0.14, and itertools 0.11.0 is already in the graph (for lalrpop). So cargo consolidates — using
   the already-present 0.11.0 rather than keeping a separate 0.13.0 copy. It's the resolver deduplicating
   as a side effect of re-solving the chain.                                                             
  The change is valid: bindgen 0.72.1 explicitly declares itertools = ">=0.10,<0.14", so 0.11.0 is within
   its stated range.

litianningdatadog (approved):

PR #1195, opened 2026-04-21
chore(codeowners): add `apm-serverless` as co-owner of `bottlecap/src/traces/`

lucaspimentel:
## Overview

- Add `@DataDog/apm-serverless` as co-owner of `bottlecap/src/traces/` alongside `@DataDog/serverless-aws` in `.github/CODEOWNERS`.
- Delete the root `CODEOWNERS` file; the one in `.github/` takes precedence, so the root file was dead weight.

## Testing

- Verified via GitHub UI that the `.github/CODEOWNERS` rules resolve correctly for files under `bottlecap/src/traces/`.

> *"Two teams, one codepath — may your review notifications be ever balanced."* — Claude 🤖

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,0 @@
-* @DataDog/serverless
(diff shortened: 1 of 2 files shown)


duncanista (approved):

CODEOWNERS
```diff
@@ -1 +0,0 @@
-* @DataDog/serverless
```
lucaspimentel, line 1: This file was being ignored because `.github/CODEOWNERS` takes precedence.

PR #1214, opened 2026-04-27
Fix E2E test trigger

shreyamalpani:
## Overview
Change how the E2E tests are triggered. Launches pipelines for `amd64` and `amd64, fips` flavors, will add tests for `arm64` and `arm64, fips` as well if these work as expected.

## Testing 
E2E tests pass on this PR


.gitlab/templates/pipeline.yaml.tpl
```diff
@@ -231,7 +231,7 @@ publish layer [self-monitoring] ({{ $flavor.name }}):
 {{ end }}  # end flavors
 
 {{ range $f := (ds "flavors").flavors }}
-{{ if $f.needs_layer_publish }}
+{{ if and $f.needs_layer_publish (eq $f.arch "arm64") }}
```
purple4reina: I would prefer if you were to choose only one, to choose arm64, since that's what we use in our nightly runs.

shreyamalpani: Sounds good, updated this

purple4reina (approved):

PR #1225, opened 2026-05-13
feat(bottlecap): make flush retry/DLQ/deadline-margin configurable

purple4reina:
## Summary

Adds three new config knobs and migrates the `FLUSH_RETRY_COUNT` constant to a per-config value. **No runtime behavior change yet on its own** — this is the foundation for upcoming deadline-aware per-request timeouts and an in-memory dead letter queue that will ensure extension flushes never cause a customer's Lambda to time out.

## New config fields

All read once at startup into `Arc<Config>` via env-var (`DD_*`) and YAML bindings, alongside the existing `flush_timeout` plumbing.

| Env var | Default | Effect |
|---|---|---|
| `DD_FLUSH_DEADLINE_MARGIN_MS` | **100** | Safety margin (ms) reserved at the tail of the Lambda budget so the extension can finish its `/extension/event/next` call after flushing. Will gate the upcoming per-request deadline math. |
| `DD_FLUSH_DLQ_MAX_BYTES` | **52_428_800** (50 MiB) | Cap on combined in-memory size of the upcoming dead letter queue across all telemetry types. |
| `DD_FLUSH_RETRY_ATTEMPTS` | **3** | Max attempts each flusher will make per `flush()` call before returning the payload for redrive / DLQ. Previously the `FLUSH_RETRY_COUNT: usize = 3` constant in `lib.rs`. |

`DD_FLUSH_TIMEOUT` default is **unchanged at 30 s**.

## What changed

- `bottlecap/src/config/{mod,env,yaml}.rs`: new fields + bindings + merge logic + test fixtures.
- `bottlecap/src/lib.rs`: removed the `FLUSH_RETRY_COUNT` constant.
- `bottlecap/src/logs/flusher.rs`, `traces/proxy_flusher.rs`: `send()` associated functions now take `max_attempts: u32`; callers pass `self.config.flush_retry_attempts`.
- `bottlecap/src/traces/stats_flusher.rs`: retry loop reads `self.config.flush_retry_attempts`.
- `bottlecap/src/traces/trace_flusher.rs`: `trace_retry_strategy()` now takes `&Config` and reads the value at runtime.

## Test plan

- [x] \`cargo test --lib\` — all 528 existing unit tests pass.
- [x] \`cargo check\` clean.
- [ ] Follow-up PR will land deadline-aware per-request timeout and dead letter queue (this is the foundation).

## Why draft

This is part 1 of a multi-PR effort to fix a class of Lambda timeouts caused by stalled telemetry intake. Draft until the follow-up PRs land so review can evaluate the full picture; this piece alone is intentionally inert.

### DIFF
--- a/bottlecap/src/config/env.rs
+++ b/bottlecap/src/config/env.rs
@@ -59,6 +59,29 @@ pub struct EnvConfig {
     #[serde(deserialize_with = "deserialize_option_lossless")]
     pub flush_timeout: Option<u64>,
 
+    /// @env `DD_FLUSH_DEADLINE_MARGIN_MS`
+    ///
+    /// Safety margin (ms) reserved at the tail of the Lambda budget so the
+    /// extension can finish `/extension/event/next` after flushing. Per-request
+    /// timeout = `min(flush_timeout, deadline_ms - now_ms - flush_deadline_margin_ms)`.
+    #[serde(deserialize_with = "deserialize_option_lossless")]
+    pub flush_deadline_margin_ms: Option<u64>,
+
+    /// @env `DD_FLUSH_DLQ_MAX_BYTES`
+    ///
+    /// Maximum combined size (bytes) of the in-memory dead letter queue across
+    /// all telemetry types. When pushing a still-failed payload would exceed
+    /// this cap, the incoming payload is dropped.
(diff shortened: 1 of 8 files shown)


purple4reina:

PR #1226, opened 2026-05-13
feat(bottlecap): thread invocation deadline through flushers

purple4reina:
## Summary

Plumbing-only change. Adds an `Arc<AtomicU64>` shared cell holding the current Lambda invocation's `deadline_ms`, and threads it from `handle_next_invocation` (where it gets published on every INVOKE event) into each bottlecap-owned flusher's struct.

**No runtime behavior change yet** — the field is held but not yet consumed. The follow-up PR will read it inside `tokio::time::timeout(...)` wrappers around each HTTP `send()` so a single hung intake call can never overrun the customer's Lambda budget.

> [!NOTE]
> Stacked on #1225 — base set to `rey.abolofia/flush-config-knobs`.

### `bottlecap/src/flushing/mod.rs`

- `InvocationDeadline` type alias for `Arc<AtomicU64>` (epoch ms; `0` means "no deadline known")
- `new_invocation_deadline()` constructor
- `compute_flush_cap(deadline_ms, flush_timeout_secs, margin_ms) -> Duration` — the cap formula:
  ```
  min(flush_timeout, deadline_ms − now_ms − margin_ms)
  ```
  saturating to `Duration::ZERO` when there is no budget left (or when no deadline is known, in which case it returns the static ceiling).
- `compute_flush_cap_from(&InvocationDeadline, ...)` convenience that loads from the atomic.
- 4 new unit tests covering: no-deadline case, tighter-than-ceiling case, deadline-already-passed, and remaining-less-than-margin.

### Each flusher

`LogsFlusher` (inner `Flusher`), `TraceFlusher`, `StatsFlusher`, `ProxyFlusher` (`Flusher` in `traces/proxy_flusher.rs`) gain an `invocation_deadline: InvocationDeadline` field, threaded through `new()` at construction. Fields carry `#[allow(dead_code)]` until the follow-up PR consumes them.

### `bottlecap/src/bin/bottlecap/main.rs`

- `extension_loop_active` builds a single `InvocationDeadline` and clones it into `start_logs_agent` and `start_trace_agent`, which forward it to each flusher constructor.
- `handle_next_invocation` takes `&InvocationDeadline` and:
  - On `NextEventResponse::Invoke`: stores the event's `deadline_ms`
  - On `NextEventResponse::Shutdown`: resets to `0`

## What's intentionally out of scope

- Metrics (`dogstatsd` in `serverless-components`) — the per-request timeout for `ship_series` / `ship_distributions` lands via an upstream PR in `serverless-components`.
- Actually wrapping `send().await` calls in `tokio::time::timeout(...)` — next PR in the stack.
- DLQ persistence of failed payloads across invocations — also next.

## Test plan

- [x] `cargo test --lib` — all **532 tests pass** (4 new + 528 existing).
- [x] `cargo check` clean — no warnings.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -38,7 +38,7 @@ use bottlecap::{
         },
     },
     fips::{log_fips_status, prepare_client_provider},
-    flushing::FlushingService,
+    flushing::{FlushingService, InvocationDeadline, new_invocation_deadline},
     lifecycle::{
         flush_control::{DEFAULT_CONTINUOUS_FLUSH_INTERVAL, FlushControl, FlushDecision},
         invocation::processor_service::{InvocationProcessorHandle, InvocationProcessorService},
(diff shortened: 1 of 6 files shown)


purple4reina:

PR #1227, opened 2026-05-13
feat(bottlecap): adaptive per-request flush timeout from invocation deadline

purple4reina:
## Summary

This is the first behavior-changing PR in the stack. It wraps every flusher HTTP send in a per-attempt `tokio::time::timeout(cap, ...)` where `cap` is derived from the current Lambda invocation's `deadline_ms`. When the remaining invocation budget is exhausted, the flusher bails immediately and the payload is returned for redrive on the next invocation instead of risking a Lambda timeout.

> [!NOTE]
> Stacked on #1226 — base set to `rey.abolofia/flush-deadline-plumbing`.

## The cap formula

```
cap = min(DD_FLUSH_TIMEOUT, deadline_ms - now_ms - DD_FLUSH_DEADLINE_MARGIN_MS)
```

Defaults: `DD_FLUSH_TIMEOUT=30s` (unchanged), `DD_FLUSH_DEADLINE_MARGIN_MS=100ms`. Each retry attempt recomputes `cap` fresh so timeouts tighten as the deadline approaches. `cap == 0` (or no deadline known, e.g. before the first INVOKE) returns the payload for redrive **without** making the HTTP call.

## What's wrapped

| Flusher | Mechanism |
|---|---|
| **Logs** (`bottlecap/src/logs/flusher.rs`) | `Flusher::send` wraps `cloned_req.send()` in `tokio::time::timeout(cap, ...)` |
| **EVP-proxy / llmobs / profile / etc.** (`bottlecap/src/traces/proxy_flusher.rs`) | Same wrap pattern on the proxy `send()` |
| **Stats** (`bottlecap/src/traces/stats_flusher.rs`) | Builds a fresh `Endpoint` per attempt with `timeout_ms = cap`; `libdd_trace_utils::send_with_retry` enforces it internally via `tokio::time::timeout`. `OnceCell<Endpoint>` changed to `OnceCell<hyper::Uri>` (only the URL parsing is cached). |
| **Traces** — **dual-shipping endpoints** (`bottlecap/src/traces/trace_flusher.rs`) | Mutates `endpoint.timeout_ms = cap_ms` on each cloned additional endpoint before constructing `SendData` |
| **Traces** — **primary endpoint** | ⚠ **Not yet adaptive.** Stays on the static `config.flush_timeout * S_TO_MS` set at trace-process time. Needs an upstream `SendDataBuilder::with_timeout_ms` in `libdatadog`. Follow-up PR will bump the rev. |

## Failure modes

[...]

PR #1228, opened 2026-05-13
feat(bottlecap): add in-memory dead letter queue infrastructure

purple4reina:
## Summary

Adds the storage layer for the dead letter queue (DLQ) that will hold payloads which exhaust both the per-flusher retry loop and the cross-flusher redrive. **Groundwork only** — the queues are constructed and integrated into `FlushingService`, but nothing pushes into them yet. That lands in the follow-up PR that captures redrive returns and drains DLQs at the start of each flush.

> [!NOTE]
> Stacked on #1227 — base set to `rey.abolofia/flush-adaptive-per-request-timeout`.

### `bottlecap/src/flushing/dlq.rs` (new file)

- `Dlq` struct holding one `Mutex<VecDeque<DlqItem<T>>>` per telemetry type (logs, traces, stats, proxy, metrics) plus a shared `AtomicU64` byte counter and a `max_bytes` ceiling.
- Constructed via `Dlq::new(max_bytes)` which returns `Arc<Dlq>` so spawned redrive tasks can push without restructuring `&self` into `&mut self`.
- `try_reserve` / `release` use `compare_exchange` so concurrent pushers can't race past the cap.
- Per-type `try_push_*` helpers that on overflow:
  - **Drop the incoming item** (matches dd-trace-py's `BufferedEncoder::put` `BufferFull` behavior at `internal/_encoding.pyx:411-416`).
  - Log at **error level**: `DROP | type=... bytes=... reason=dlq_full dlq_bytes=N/MAX` — visible at default log levels.
  - Log a trace-level marker note (full payload contents will be logged at the call site by the follow-up PR that uses these helpers).
- Size estimators:
  - `estimate_stats_size()` uses prost's `encoded_len()` (exact).
  - `estimate_metrics_batch_size()` uses per-item heuristics (512 B/series, 2 KiB/sketch) so sizing is cheap; the DLQ budget is a soft cap, not exact accounting.
- 5 unit tests covering reserve under/over cap, release semantics, u64 overflow guard, and the stats size estimator.

### `bottlecap/src/flushing/service.rs`

- `FlushingService` gains a `pub(crate) dlq: Arc<Dlq>` field.
- `FlushingService::new` now takes `dlq_max_bytes: u64` (sourced from `config.flush_dlq_max_bytes`).
- `dlq()` accessor returns `Arc<Dlq>` for test/inspection.

### `bottlecap/src/flushing/mod.rs`

- Re-exports `Dlq`, `DlqItem`, `DropReason`, `TelemetryType`, and the two size-estimator functions.

### `bottlecap/src/bin/bottlecap/main.rs`

- Both `FlushingService::new` call sites now pass `config.flush_dlq_max_bytes`.

## What's intentionally out of scope

- **Wiring the DLQ into `await_handles` and `flush_blocking_inner`** — next PR in the stack. That's where redrive failures get captured and queued payloads are drained back into the flusher.
- **SHUTDOWN final-drain + drop log summary** — also next.
- **Trace-level logging of dropped payload contents** at the call site (the helper here logs only metadata; full contents are typed and live with the caller).

## Test plan

- [x] `cargo test --lib` — all **537 tests pass** (5 new DLQ + 532 existing).
- [x] `cargo check` clean — no warnings.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -469,6 +469,7 @@ async fn extension_loop_active(
         // data is sent to Datadog even while concurrent invocations are being processed.
         // The flushing happens independently of invocation lifecycle events.
         // This background task runs until shutdown is signaled via cancel_token_clone.
+        let flush_dlq_max_bytes = config.flush_dlq_max_bytes;
         let flush_task_handle = tokio::spawn(async move {
             let mut flushing_service = FlushingService::new(
                 logs_flusher_clone,
@@ -477,6 +478,7 @@ async fn extension_loop_active(
                 proxy_flusher_clone,
                 metrics_flushers_clone,
                 metrics_aggr_handle_clone,
+                flush_dlq_max_bytes,
             );
 
             loop {
@@ -651,6 +653,7 @@ async fn extension_loop_active(
         proxy_flusher.clone(),
         Arc::clone(&metrics_flushers),
         metrics_aggregator_handle.clone(),
+        config.flush_dlq_max_bytes,
     );
     handle_next_invocation(next_lambda_response, &invocation_processor_handle, &invocation_deadline).await;
     loop {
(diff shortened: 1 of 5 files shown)


purple4reina:

PR #1233, opened 2026-05-21
fix: index dd_resource_key on inferred API Gateway spans via span.tags

Vesuri:
## Summary

- `dd_resource_key` on inferred API Gateway spans was written to `span.meta` instead of `span.tags`
- `span.meta` is not indexed as a searchable APM facet, so the Event Platform `dd_resource_key:*` filter could not find the API GW span by its own ARN
- The span was only matched via the Lambda ARN propagated into its indexed tags, causing both the API GW span and the Lambda span to yield the same `dd_resource_key` value in the query — the `source <> target` deduplication condition then eliminated the edge entirely
- Writing to `span.tags` makes the API Gateway ARN indexed and discoverable, enabling cloud architecture map APM trace overlays to render API Gateway → Lambda connections

## Test plan

- [ ] Existing unit tests for `get_dd_resource_key()` in `api_gateway_rest_event.rs` and `api_gateway_http_event.rs` continue to pass (they test the value, not the field it's written to)
- [ ] Deploy to a Lambda behind an API Gateway with the Datadog extension; verify `dd_resource_key` appears as an indexed span tag (searchable via `@dd_resource_key:*` in APM) on the inferred `aws.apigateway` span with value `arn:aws:apigateway:{region}::/restapis/{api_id}`
- [ ] Confirm cloud architecture map datadogtrace overlay renders API Gateway → Lambda connections for the account/region

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bottlecap/src/lifecycle/invocation/span_inferrer.rs
+++ b/bottlecap/src/lifecycle/invocation/span_inferrer.rs
@@ -250,7 +250,7 @@ impl SpanInferrer {
 
             if let Some(dd_resource_key) = t.get_dd_resource_key(&aws_config.region) {
                 inferred_span
-                    .tags
+                    .meta
                     .insert("dd_resource_key".to_string(), dd_resource_key);
             }
 
--- a/bottlecap/src/lifecycle/invocation/span_inferrer.rs
+++ b/bottlecap/src/lifecycle/invocation/span_inferrer.rs
@@ -250,7 +250,7 @@ impl SpanInferrer {
 
             if let Some(dd_resource_key) = t.get_dd_resource_key(&aws_config.region) {
                 inferred_span
-                    .meta
+                    .tags
                     .insert("dd_resource_key".to_string(), dd_resource_key);
             }
 


Vesuri:
Thanks for the review. Reverted the  change — you're right that  doesn't exist on  (CI confirmed: compile error E0609). Only  is the correct map for string attributes.

After further investigation we also believe the  placement is fine: other span attributes written to  (e.g. EC2  values) ARE searchable in the Event Platform trace track. Our earlier assumption that `meta` fields weren't indexed was based on a misread of APM UI search behaviour (which uses indexed facets, not the same as the Event Platform query).

The substantive change in this PR is the use of `get_dd_resource_key()` (returning `/restapis/{id}` / `/apis/{id}` without stage suffix) rather than `get_arn()` (which includes `/stages/{stage}`). This aligns the value with what cloud architecture map uses as its resource node IDs.

We're in the process of verifying end-to-end that the Event Platform `dd_resource_key:*` filter correctly finds API GW inferred spans with this format — staging infrastructure was unavailable today. Happy to close this PR if the team confirms the format was already correct in an earlier version, or if there's a different preferred fix.

Vesuri:
Reopening with concrete evidence of the bug.

**Verified reproduction in org2 staging** using cloudcraftdev account (123734022431), extension v97, Node.js tracer, API Gateway v2 HTTP API (`vpt0wftm5g`) → Lambda (`myFunction1`).

Running the exact `sourceTraces` CTE query from cloud-maps `datadogtraceplugin.go` against org2:

```
service: vpt0wftm5g.execute-api.us-east-1.amazonaws.com  (inferred API GW span)
dd_resource_key: arn:aws:lambda:us-east-1:123734022431:function:myfunction1  ← Lambda ARN, not API GW ARN
```

The API GW ARN (`arn:aws:apigateway:us-east-1::/apis/vpt0wftm5g`) is correctly set in `meta.dd_resource_key` by the extension, but the Event Platform `dd_resource_key` **column** returns the Lambda ARN — propagated from the child Lambda span at trace level. Both the API GW span and Lambda span resolve to the same Lambda ARN, so the `source <> target` JOIN condition eliminates the edge and the API GW → Lambda connection is never produced.

**Root cause:** The Event Platform column reads from the indexed facet, not raw `meta`. The indexed `@dd_resource_key` on the inferred span is the Lambda ARN from trace-level propagation, shadowing the API GW ARN written to `meta`.

**What is needed:** A way to set the API GW ARN as the indexed `dd_resource_key` on the inferred span, overriding the propagated value. Asking for guidance on the correct field or API to use for this in the extension — `span.meta` is clearly not sufficient, but `span.tags` does not exist on `libdd_trace_protobuf::pb::Span`.

Vesuri:
Closing — the fix direction is confirmed but the exact mechanism (how to set an indexed span tag from the extension's Rust protobuf code) needs input from the team. See the updated message in #serverless-aws for full context. Will reopen with the correct implementation once we have guidance.

PR #1239, opened 2026-05-27
[SVLS-9189] feat(logs): add config param DD_LAMBDA_DURABLE_FUNCTION_LOG_BUFFER_SIZE

lym953:
## Background
For durable functions, the extension holds logs and waits to get the invocation context from the tracer so the extension can enrich the logs. 

## Problem
When tracing is off (which is not recommended for durable functions), logs are flushed when:
1. the buffer is full (buffer size is all logs for 50 invocations). This increases delay. Or
2. when extension shuts down. However, this is not guaranteed to happen due to race conditions, causing missing logs.

## This PR

- Adds `lambda_durable_function_log_buffer_size` config field (env var `DD_LAMBDA_DURABLE_FUNCTION_LOG_BUFFER_SIZE`, default: `0`) controlling the max number of request ID keys held in `held_logs`.
- When set to `0`, logs bypass the hold mechanism and are sent immediately without durable execution context enrichment.
- Setting the default to 0 for now because the tracer changes for durable functions are not released. Later, after tracer changes are released, we can change the default back to 50 and ask customers to install the tracer for durable functions.

### Steps
- Build an extension layer
- install it on a durable function
- execute the function
- wait for a few minutes until the environment shuts down

### Result

Before:
- No log appears in Datadog
- and the invocation doesn't appear either

After:
- The invocation appears
image

- Logs also appear
image

### DIFF
--- a/bottlecap/src/config/env.rs
+++ b/bottlecap/src/config/env.rs
@@ -489,14 +489,14 @@ pub struct EnvConfig {
     #[serde(deserialize_with = "deserialize_string_or_int")]
     pub org_uuid: Option<String>,
 
-    /// @env `DD_DURABLE_FUNCTION_LOG_BUFFER_SIZE`
+    /// @env `DD_LAMBDA_DURABLE_FUNCTION_LOG_BUFFER_SIZE`
     ///
     /// Maximum number of request IDs whose logs are held waiting for durable execution
     /// context. Set to 0 to disable log holding; logs will be sent immediately without
     /// durable execution context enrichment. Useful when the tracer is not installed.
     /// Default is `0`.
     #[serde(deserialize_with = "deserialize_option_lossless")]
-    pub durable_function_log_buffer_size: Option<usize>,
+    pub lambda_durable_function_log_buffer_size: Option<usize>,
 }
 
 #[allow(clippy::too_many_lines)]
@@ -701,7 +701,7 @@ fn merge_config(config: &mut Config, env_config: &EnvConfig) {
     merge_option_to_value!(config, env_config, api_security_sample_delay);
 
     merge_string!(config, dd_org_uuid, env_config, org_uuid);
-    merge_option_to_value!(config, env_config, durable_function_log_buffer_size);
+    merge_option_to_value!(config, env_config, lambda_durable_function_log_buffer_size);
 }
 
 #[derive(Debug, PartialEq, Clone, Copy)]
@@ -1064,7 +1064,7 @@ mod tests {
                 api_security_sample_delay: Duration::from_secs(60),
 
                 dd_org_uuid: String::default(),
-                durable_function_log_buffer_size: 0,
+                lambda_durable_function_log_buffer_size: 0,
             };
 
             assert_eq!(config, expected_config);
--- a/bottlecap/src/config/env.rs
+++ b/bottlecap/src/config/env.rs
@@ -488,6 +488,15 @@ pub struct EnvConfig {
     /// The Datadog organization UUID. When set, delegated auth is auto-enabled.
     #[serde(deserialize_with = "deserialize_string_or_int")]
     pub org_uuid: Option<String>,
+
+    /// @env `DD_DURABLE_FUNCTION_LOG_BUFFER_SIZE`
+    ///
+    /// Maximum number of request IDs whose logs are held waiting for durable execution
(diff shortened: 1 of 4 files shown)


bottlecap/src/logs/lambda/processor.rs
```diff
@@ -143,6 +141,7 @@ impl LambdaProcessor {
 
         let processing_rules = &datadog_config.logs_config_processing_rules;
         let logs_enabled = datadog_config.serverless_logs_enabled;
+        let held_logs_max_keys = datadog_config.durable_function_log_buffer_size;
```
litianningdatadog: nit: rename held_logs_max_keys to durable_function_log_buffer_size or sth. similar for better readability.

bottlecap/src/config/env.rs
```diff
@@ -488,6 +488,15 @@ pub struct EnvConfig {
...
     pub org_uuid: Option<String>,
+
+    /// @env `DD_DURABLE_FUNCTION_LOG_BUFFER_SIZE`
+    ///
+    /// Maximum number of request IDs whose logs are held waiting for durable execution
+    /// context. Set to 0 to disable log holding; logs will be sent immediately without
+    /// durable execution context enrichment. Useful when the tracer is not installed.
+    /// Default is `0`.
+    #[serde(deserialize_with = "deserialize_option_lossless")]
+    pub durable_function_log_buffer_size: Option<usize>,
```
duncanista: Wondering if this should include the `LAMBDA` prefix, since we're gonna migrate to the config package from serverless-components eventually,

that way it makes it more descriptive that its for the lambda product

bottlecap/src/logs/lambda/processor.rs
lym953: Renamed it as `lambda_durable_function_log_buffer_size`

bottlecap/src/config/env.rs
lym953: Renamed it as `lambda_durable_function_log_buffer_size`

duncanista: I don't see the rename?

duncanista: I see it only in the processor, not the config

lym953: Good catch. Fixed.

bottlecap/src/config/env.rs
```diff
@@ -488,6 +488,15 @@ pub struct EnvConfig {
     /// The Datadog organization UUID. When set, delegated auth is auto-enabled.
     #[serde(deserialize_with = "deserialize_string_or_int")]
     pub org_uuid: Option<String>,
+
+    /// @env `DD_LAMBDA_DURABLE_FUNCTION_LOG_BUFFER_SIZE`
```
litianningdatadog, line 500: The PR title and message both mentioned DD_DURABLE_FUNCTION_LOG_BUFFER_SIZE which is different from above comment. Did I miss anything?

bottlecap/src/logs/lambda/processor.rs
```diff
@@ -684,7 +685,7 @@ impl LambdaProcessor {
     /// arrives.
     fn hold_log(&mut self, request_id: String, log: IntakeLog) {
         if !self.held_logs.contains_key(&request_id) {
-            while self.held_logs.len() >= HELD_LOGS_MAX_KEYS {
+            while self.held_logs.len() >= self.lambda_durable_function_log_buffer_size {
```
litianningdatadog, line 720: in case self.lambda_durable_function_log_buffer_size  == 0, this check will always be true, while could lead to infinite loop. Is this an expected behavior?

litianningdatadog (commented):
left some comments

bottlecap/src/config/env.rs
lym953: Good catch. PR title and message were outdated. Updated.

bottlecap/src/logs/lambda/processor.rs
lym953: The caller already ensures `self.lambda_durable_function_log_buffer_size > 0` before calling `held_logs()`.
For clarity:
1. add a comment: Callers must ensure `self.lambda_durable_function_log_buffer_size > 0`
2. In side `held_logs()`, if `lambda_durable_function_log_buffer_size == 0`, return early to avoid infinite loop, as a safety net.

litianningdatadog (approved):

Issue #1237, opened 2026-05-22
aws.lambda.enhanced.out_of_memory not emitted when Lambda reaches memory limit but AWS reports as timeout


Hamdi-Hassan-seccl:
Hi Team

We are investigating an incident where a Lambda reached its memory limit but our Datadog monitor based on `aws.lambda.enhanced.out_of_memory` did not trigger.

We tried reproducing a similar case in sandbox Lambda. Enhanced metrics are working generally as expected as we can see that `aws.lambda.enhanced.invocations` and `aws.lambda.enhanced.max_memory_used`.
However,  the `out_of_memory one remains no-data even when Lambda memory usage reaches the configured limit. ( see screenshot attached below)

<img width="1475" height="524" alt="Image" src="https://github.com/user-attachments/assets/356ff785-57c5-47c8-a0b4-c8bbd2e868a6" />

Example CloudWatch output we've observed in when reproduced the issue:
```
INIT_REPORT Init Duration: 180000.45 ms Phase: invoke Status: timeout
REPORT RequestId: 514789fd-5e24-49c5-a41c-24c049a36e2a
Duration: 180000.00 ms
Billed Duration: 180000 ms
Memory Size: 192 MB
Max Memory Used: 192 MB
Status: timeout
```

We also saw similar tests where memory was very close to the limit:
```
Memory Size: 128 MB
Max Memory Used: 127 MB
Status: timeout
and:
Memory Size: 256 MB
Max Memory Used: 252–254 MB
```
In these cases, Datadog emitted `aws.lambda.enhanced.max_memory_used`, but did not emit `aws.lambda.enhanced.out_of_memory.`
The Datadog extension appears to be running and subscribed to telemetry:
```
TELEMETRY Name: datadog-agent State: Already subscribed Types: [Platform, Extension, Function]
EXTENSION Name: datadog-agent State: Ready Events: [INVOKE, SHUTDOWN]
```
**### Question**
Is `aws.lambda.enhanced.out_of_memory `metric expected to emit only for explicit runtime OOM signatures, such as:
```
Runtime exited with error: signal: killed
Runtime.ExitError
OutOfMemoryError
JavaScript heap out of memory
```
Or should it also emit when the Lambda REPORT line shows memory saturation, for example:
```
Memory Size: 192 MB
Max Memory Used: 192 MB
Status: timeout
```
From an operational perspective, this looks like memory exhaustion leading to timeout, but the OOM metric does not emit.
[...]

PR #1241, opened 2026-05-29
[SVLS-9175] feat: emit OOM metric on memory equality with per-request dedup

lym953:
## Background
From our knowledge (before this PR), here's the behavior when each runtime OOMs:
- emits runtime-specific error message. This can happen on **Java**, **Node** (case 1 in the table below) and **.NET**
- In `PlatformRuntimeDone` event, `error_type` is `Runtime.OutOfMemory`. This can happen on **Python** and **Ruby**.
- In `PlatformReport` event, `max_memory_used == memory_size`. This can happen on **Python**, **Ruby**, **Node** and **Go**.

To capture OOM for all these scenarios (except Node case 2, which was just called out in #1237) without double counting, right now the extension emits `aws.lambda.enhanced.out_of_memory` metric in these scenarios:
- when we see runtime-specific error messages for Java, Node and .NET
- when we see `Runtime.OutOfMemory`
- when we see `max_memory_used == memory_size` for Go, i.e. only when runtime is `provided.al2`. We don't do this for other runtimes (Python, Ruby, Node) to avoid double counting.

image

## Problem
In issue #1237, a customer called out a new scenario: "Node (case 2)" in the table. The only evidence of OOM is `max_memory_used == memory_size`, and there is no runtime-specific log message. As a result, OOMs like this are not captured by the OOM enhanced metric.

## This PR
- Regardless of runtime, use all the three ways to capture OOM.
- In addition, dedup by request_id to avoid double counting.
- Add one integration test per runtime (except for Node, which has 2 tests)

## Test plan

Passed the added unit tests and integration tests.

## To reviewers
Most of the code changes are for integration tests.

## Details (generated by Claude Code)

Closes the gap surfaced in #1237: a Node.js Lambda that hit its memory limit (`Memory Size 192 MB / Max Memory Used 192 MB`, `Status: timeout`) did not emit `aws.lambda.enhanced.out_of_memory` because none of the three existing detection paths matched.

- **Why the existing paths missed it.** V8 spent its budget in GC rather than declaring `JavaScript heap out of memory`, so the runtime log-line match never fired. The runtime crashed on a wall-clock timeout, so `PlatformRuntimeDone` reported no `error_type`. And the `max_memory_used_mb == memory_size_mb` check in `PlatformReport` was gated on `runtime.starts_with("provided.al")` to avoid double-counting against the log path, so Node was excluded.

- **What changes.** Drop the `provided.al*` restriction so the equality check applies to every runtime. To avoid double-counting against the two pre-existing paths (some invocations satisfy both equality and `Runtime.OutOfMemory` simultaneously), add a per-`Context` `oom_emitted` flag. All three detection paths funnel through a new `Processor::try_increment_oom_metric`, which checks/sets the flag and is a no-op on subsequent calls for the same `request_id`.

- **Plumbing.** `Event::OutOfMemory` now carries an `Option<String> request_id`. The log-path detector reads it from `LambdaProcessor::invocation_context.request_id` (set on `PlatformStart`, cleared on `PlatformRuntimeDone`/`PlatformReport`). `None` is only realistic in Managed Instance mode (extensions can't subscribe to INVOKE there); the helper falls back to a best-effort emit without dedup in that case.

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -1429,7 +1429,7 @@ impl Processor {
             }
         } else {
             debug!(
-                "Invocation Processor | Emitting OOM metric without dedup: no request_id available (likely LMI mode)"
+                "Invocation Processor | Emitting OOM metric without dedup: no request_id available (OOM log processed before PlatformStart or after PlatformRuntimeDone)"
             );
         }
         self.enhanced_metrics.increment_oom_metric(timestamp);
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -920,9 +920,9 @@ impl Processor {
         // If the invocation hit the memory limit, increment the OOM metric. This catches
         // OOM-induced failures that don't surface through a runtime-specific log line or a
         // `Runtime.OutOfMemory` error_type — most notably the suppressed-init / timeout-at-cap
-        // pattern reported in datadog-lambda-extension#1237 (Node). Dedup against the other
-        // two detection paths is handled by `Context::oom_emitted`, which
-        // `try_increment_oom_metric` checks and sets.
+        // pattern reported in datadog-lambda-extension#1237 (Node). Best-effort dedup
+        // against the other two detection paths is handled by `try_increment_oom_metric`
+        // (it can still double-count in edge cases — see that function's doc).
         if metrics.max_memory_used_mb == metrics.memory_size_mb {
             debug!(
                 "Invocation Processor | PlatformReport | Last invocation hit memory limit. Incrementing OOM metric."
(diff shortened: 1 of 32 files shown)


bottlecap/src/logs/lambda/processor.rs
```diff
@@ -163,6 +163,23 @@ impl LambdaProcessor {
...
+    /// Returns the `request_id` of the currently-active invocation, if known.
+    /// Used by the OOM log-line detector to tag `Event::OutOfMemory` so that
+    /// `Processor::try_increment_oom_metric` can dedup against the other two
+    /// detection paths (`Runtime.OutOfMemory` and `PlatformReport` equality).
+    ///
+    /// `invocation_context.request_id` is set on `PlatformStart` and cleared on
+    /// `PlatformRuntimeDone` / `PlatformReport`. Returns `None` in LMI mode,
+    /// where extensions cannot subscribe to the `INVOKE` event so
+    /// `platform.start` is never delivered.
+    fn current_request_id(&self) -> Option<String> {
```
lym953, line 177: To reviewers: is it anti-pattern to get request_id in this way?

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -1395,7 +1395,34 @@ impl Processor {
...
+        {
+            if ctx.oom_emitted {
+                debug!(
+                    "Invocation Processor | OOM metric already emitted for request_id {}, skipping",
+                    rid
+                );
+                return;
+            }
+            ctx.oom_emitted = true;
+        }
```
litianningdatadog, line 1443: I’m curious: in what cases would the request ID be empty or unavailable? Are either of those cases valid? Maybe we can add a debug log for this.

litianningdatadog (commented):
Left a minior comment

bottlecap/src/lifecycle/invocation/processor.rs
lym953: 1. Added a debug log
2. Tested on LMI. OOM log can arrive before the extension receives request_id from PlatformStart event. In this case, request_id is empty. Updated comment to explain this.

bottlecap/src/logs/lambda/processor.rs
lym953: ~~Actually, for all three cases, we can get request_id from either log payload or telemetry event payload, so we don't need this current_request_id() function. Let me delete it.~~

lym953: For the error
> FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory

No request id is available from the log itself, so we still need the function `current_request_id()`.

bottlecap/src/logs/lambda/processor.rs
```diff
@@ -194,7 +211,10 @@ impl LambdaProcessor {
                 if let Some(message) = message {
                     if is_oom_error(&message) {
                         debug!("LOGS | Got a runtime-specific OOM error. Incrementing OOM metric.");
-                        if let Err(e) = self.event_bus.send(Event::OutOfMemory(event.time.timestamp())).await {
+                        if let Err(e) = self.event_bus.send(Event::OutOfMemory {
+                            request_id: self.current_request_id(),
+                            timestamp: event.time.timestamp(),
+                        }).await {
```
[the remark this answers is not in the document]
lym953: Addressed. If request_id is available from log record or event payload, use it. Otherwise, rely on `current_request_id()` to get request_id.

integration-tests/tests/oom.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    );
+
+    await sleep(INITIAL_WAIT_MS);
+
+    const deadline = windowStart + TOTAL_BUDGET_MS;
+    let counts: Record<string, number> = {};
+    let attempt = 0;
+    while (Date.now() < deadline) {
+      attempt++;
+      counts = await fetchCounts(windowStart, Date.now());
```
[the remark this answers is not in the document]
lym953: Addressed.

integration-tests/lambda/oom-go/main.go
```diff
@@ -0,0 +1,30 @@
+// OOM reproducer for Go.
+// Allocates a 500 MB byte slice in a single shot, then writes to every page
+// to force physical commit. On a 192 MB Lambda this immediately exceeds the
+// cgroup memory limit and the kernel SIGKILLs the process, producing a
+// PlatformReport with max_memory_used_mb == memory_size_mb. The Go runtime
```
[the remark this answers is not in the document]
lym953: Fixed.

litianningdatadog (approved):

lym953:
The failing e2e test `aws.lambda.enhanced.init_duration` is irrelevant. Ignoring it and merging this PR.

PR #1244, opened 2026-06-03
chore(deps): bump `libdatadog` to `48da0d8` for client-computed header fix

lucaspimentel:
## Overview

Bump all `libdd-*` dependencies from `db05e1f` to `48da0d8` to pick up [libdatadog PR #2071](https://github.com/DataDog/libdatadog/pull/2071), which fixes how the `datadog-client-computed-stats` and `datadog-client-computed-top-level` headers are parsed so the behavior matches the Go trace-agent.

Bottlecap consumes this exact path in `bottlecap/src/traces/trace_agent.rs` via `(&parts.headers).into()`, so it directly benefits from the fix.

### Adapting to breaking API changes between the two revisions

- `HttpClientTrait` was renamed to `HttpClientCapability`
- `SendData::send` now requires `C: HttpClientCapability + SleepCapability`, so implemented `SleepCapability` for `HttpClient`

### Testing

Added Tier 0 coverage in `trace_agent.rs` pinning the new header-parsing behavior at the `(&headers).into()` boundary `handle_traces` uses:
- `true`, `yes`, `t`, `1` -> flag set
- `false`, `0`, `f`, `F`, `FALSE`, `False` -> flag clear
- absent/empty -> clear
- `client_computed_top_level` no longer set by presence alone (pre-bump regression)

These test would fail on the pre-bump rev (`db05e1f`), so they lock in the fix.

> *"Turns out `yes` means yes and `false` means false. Took a whole PR to teach the parser what every toddler already knows."* — Claude 🤖

[APMSVLS-487](https://datadoghq.atlassian.net/browse/APMSVLS-487)

[APMSVLS-487]: https://datadoghq.atlassian.net/browse/APMSVLS-487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/traces/http_client.rs
+++ b/bottlecap/src/traces/http_client.rs
@@ -11,7 +11,8 @@ use hyper_http_proxy;
 use hyper_rustls::HttpsConnectorBuilder;
 use libdd_capabilities::{
     MaybeSend,
-    http::{HttpClientTrait, HttpError},
+    http::{HttpClientCapability, HttpError},
+    sleep::SleepCapability,
 };
 use libdd_common::http_common::{self, Body, GenericHttpClient};
 use rustls::RootCertStore;
@@ -21,6 +22,7 @@ use std::fs::File;
 use std::future::Future;
 use std::io::BufReader;
 use std::sync::{Arc, LazyLock};
+use std::time::Duration;
 use tracing::debug;
 
 type InnerClient =
@@ -29,8 +31,8 @@ type InnerClient =
 /// HTTP client used by trace and stats flushers.
 ///
 /// Wraps a hyper client preconfigured with optional proxy and TLS settings, and
-/// implements [`HttpClientTrait`] so it can be passed to `libdd_trace_utils`
-/// senders such as `SendData::send`.
+/// implements [`HttpClientCapability`] and [`SleepCapability`] so it can be
+/// passed to `libdd_trace_utils` senders such as `SendData::send`.
 #[derive(Clone, Debug)]
 pub struct HttpClient {
     inner: InnerClient,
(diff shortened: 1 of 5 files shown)


bottlecap/src/traces/http_client.rs
```diff
@@ -76,6 +78,20 @@ impl HttpClientTrait for HttpClient {
     }
 }
 
+impl SleepCapability for HttpClient {
```
duncanista, line 81: Shouldn't this live then either in test or with the test feature so its not included in prod code?

lucaspimentel: The libdatadog version bump made `SendData::send` require `HttpClientCapability` + `SleepCapability`, and bottlecap calls `send`, so the binary won't compile without it.

The `// never invoked on bottlecap's code paths` comment only applies to the `new()` constructor. It's trait-required factory we never call, same as `new_client()` on `HttpClientCapability` above. The `sleep()` method itself is used in production: libdatadog's sender calls it for retry/backoff between send attempts.

duncanista: mmm, and so we're getting that new capability which might not be something we want for lambda lifecycle no?

lucaspimentel: According to 🤖 

The retry/backoff isn't new with this bump. `send_with_retry` already existed at the previous revision and bottlecap already goes through it. Bottlecap configures its own strategy in `trace_flusher.rs` (`trace_retry_strategy()`: 3 retries, 0ms delay, constant backoff), so we're already opted in explicitly.

`SleepCapability` only changed *how* libdatadog reaches the sleep primitive: the request timeout went from `tokio::time::timeout(...)` to a `select!` race against `capabilities.sleep(timeout)`, which is the same `tokio::time::sleep` underneath on our target. Same runtime behavior, no new waiting in the Lambda lifecycle.

duncanista (approved):

PR #1249, opened 2026-06-10
feat(config): introduce extension config on top of shared datadog-agent-config

duncanista:
## Overview

First step of migrating bottlecap's in-tree configuration module onto the shared `datadog-agent-config` crate (lives in [DataDog/serverless-components](https://github.com/DataDog/serverless-components/tree/main/crates/datadog-agent-config)).

This PR adds the foundation only — it introduces a `LambdaConfig` extension struct (in `bottlecap/src/config/mod.rs`, alongside the existing legacy code) that implements the upstream `ConfigExtension` trait, plus a `LambdaConfigSource` deserialization shape that figment uses for both env-var and YAML loading (dual extraction).

Nothing in bottlecap consumes the extension yet — the existing `bottlecap::config::Config` struct is untouched. The follow-up #1251 (stacked on this) replaces it with `Config<LambdaConfig>`, deletes the duplicated env.rs / yaml.rs / deserializer modules, and removes the legacy `#[macro_export]` `merge_*` macros at the top of `mod.rs` (`LambdaConfig::merge_from` deliberately uses fully-qualified upstream macro paths today to coexist with them).

### Why this shape

The upstream crate is purpose-built for this migration — see PR DataDog/serverless-components#111 ("ConfigExtension trait") which landed exactly the extensibility hook we need. Each consumer (bottlecap here, future serverless agents elsewhere) supplies its own extension type for fields that don't generalize, while sharing one canonical implementation for all the core agent fields (site, api_key, logs/APM/OTLP/proxy/trace propagation, etc.).

### Fields in the extension

The 19 Lambda-specific fields that have no upstream equivalent:

- `api_key_secret_arn`, `kms_api_key`, `api_key_ssm_arn`, `api_key_secret_reload_interval`
- `serverless_logs_enabled` (OR-merged with the `DD_LOGS_ENABLED` alias)
- `serverless_flush_strategy` (custom `FlushStrategy` deserializer for `"end" | "end,N" | "periodically,N" | "continuously,N"`)
- `enhanced_metrics`, `lambda_proc_enhanced_metrics`
- `capture_lambda_payload`, `capture_lambda_payload_max_depth`
- `compute_trace_stats_on_extension`, `span_dedup_timeout`
- `dd_org_uuid` (sourced from `DD_ORG_UUID`, source field `org_uuid` → config field `dd_org_uuid`)
- `serverless_appsec_enabled`, `appsec_rules`, `appsec_waf_timeout`
- `api_security_enabled`, `api_security_sample_delay`
- `custom_metrics_exclude_tags` (sourced from `DD_LAMBDA_CUSTOMER_METRICS_EXCLUDE_TAGS` / `lambda_customer_metrics_exclude_tags:` in YAML)

Kept in the extension rather than upstreamed because they're Lambda-runtime concerns. `custom_metrics_exclude_tags` is arguably generalizable to other serverless targets but stays here for now; we can migrate it upstream when another consumer needs it.

### Pre-requisites that already merged

- DataDog/serverless-components#135 — bumps libdatadog rev across all serverless-components crates from `0a3304c` to `48da0d8` to match bottlecap.
- DataDog/serverless-components#136 — switches `datadog-agent-config`'s libdd transitive deps to `default-features = false` and exposes `https` / `fips` opt-in features. Without this, the FIPS dep tree was poisoned by an implicit `https` (ring) path.

Bottlecap pins all three serverless-components crates (`dogstatsd`, `datadog-fips`, `datadog-agent-config`) to the merge SHA `bb4dedee` so cargo deduplicates `dogstatsd`.

## Testing

37 new tests in `bottlecap/src/config/mod.rs::lambda_config_tests` cover each extension field from both env vars and YAML where applicable, plus:

- Per-field env round-trip (`api_key_secret_arn`, `kms_api_key`, `api_key_ssm_arn`, …)
- YAML coverage for the two source-to-config renames: `org_uuid` → `dd_org_uuid` and `lambda_customer_metrics_exclude_tags` → `custom_metrics_exclude_tags`
- `DD_LOGS_ENABLED` ↔ `DD_SERVERLESS_LOGS_ENABLED` OR-merge semantics
- `FlushStrategy` — `"end"`, `"end,N"` (`EndPeriodically`), `"periodically,N"`, `"continuously,N"`, invalid → `Default`
- Duration parsing — seconds, microseconds, `_ignore_zero` variant
- env precedence over YAML
- Forgiving fallback when a single field is malformed (default preserved instead of failing the whole extraction)

```
$ cargo test --lib config::lambda_config_tests
test result: ok. 37 passed; 0 failed; 0 ignored
```

Existing bottlecap config behavior is unchanged (this PR adds; it doesn't replace yet).

## Follow-up

- #1251 — stacked on this branch — does the actual wire-in: replaces `bottlecap::config::Config` with `Config<LambdaConfig>`, deletes the duplicated config files, and removes the legacy `#[macro_export]` macros at the top of `mod.rs`.


duncanista (commented):
Took a look in isolation against `main`. A few real things — none are merge blockers, but a couple are worth a response. Grouped by severity.

## Question (worth a reply)

**1. Cargo.lock has two `dogstatsd` packages now.** The PR description says "pins three serverless-components crates to the same SHA", but the diff actually pins them to two different SHAs:

```
dogstatsd            -> aaac1a5
datadog-fips         -> aaac1a5
datadog-agent-config -> bb4dede   <-- different
```

Because `datadog-agent-config` depends on `dogstatsd = { path = "../dogstatsd" }`, cargo resolves the lockfile with two distinct `dogstatsd` source entries (visible in the lock diff). Inspecting `bb4dede` upstream, the only diff vs. `aaac1a5` is the agent-config TLS-feature gating PR (#136) — the `dogstatsd` and `datadog-fips` source trees are byte-identical between the two SHAs.

Suggest bumping all three to `bb4dede` so cargo deduplicates `dogstatsd` and the description matches reality. No type-system risk today (upstream's public `Config` doesn't expose `dogstatsd::*` types — it just uses the crate internally), but it's bloat and the lockfile noise will recur on every dep bump.

**2. `capture_lambda_payload_max_depth: Option<u32>` doesn't have a graceful deserializer.** The upstream `ConfigExtension` doc (lib.rs:296-300) is explicit:

> The `Source` type **must** use `#[serde(default)]` on the struct **and graceful deserializers (e.g., `deserialize_optional_bool_from_anything`) on each field**. Without these, a missing or malformed value will cause the entire extension extraction to fail — the extension silently falls back to `E::default()` with a `tracing::warn!` log.

Every other field in `LambdaConfigSource` opts into a `deser_*` helper; this one doesn't. A user setting `DD_CAPTURE_LAMBDA_PAYLOAD_MAX_DEPTH=foo` would silently reset *all* extension fields to defaults (api_key_secret_arn, kms_api_key, …) and emit only a warn. This is also a subtle regression vs. the in-tree path, where a bad u32 only fails the legacy env extraction, not the whole config.

The in-tree code has the same shape, so this is not new behavior — but the upstream contract makes the blast radius bigger. Worth either adding a `deserialize_optional_u32_from_anything` (probably belongs upstream alongside the bool/duration helpers) or documenting why the field is allowed to violate the contract.

## Nits

**3. Test coverage doesn't match what the PR description claims.** The description says "Per-field env + YAML round-trip", but most fields have env-only tests:

- env-only, no YAML: `api_security_enabled`, `api_security_sample_delay`, `serverless_appsec_enabled`, `appsec_rules`, `enhanced_metrics`, `lambda_proc_enhanced_metrics`, `compute_trace_stats_on_extension`, `span_dedup_timeout`, `api_key_secret_reload_interval`, `dd_org_uuid`, `custom_metrics_exclude_tags`

The two source-field-to-config-field renames are exactly the cases where a YAML test matters most:
- `org_uuid` → `dd_org_uuid`: only tested via env (`DD_ORG_UUID`). If a user writes `org_uuid: ...` in `datadog.yaml`, is that the right key? `dd_org_uuid:`? Neither test exists.
- `lambda_customer_metrics_exclude_tags` → `custom_metrics_exclude_tags`: same — only env-tested.

Either tighten the description to "env round-trip per field, plus YAML for a representative subset" or add the missing YAML tests for the renames at minimum.

**4. `FlushStrategy` coverage is partial.** Tests cover `"end"`, `"periodically,N"`, and invalid. Missing: `"end,N"` → `EndPeriodically(N)`, `"continuously,N"` → `Continuously(N)`. The upstream's enum implements all four variants; bottlecap will rely on all four behaviorally. The legacy `bottlecap/src/config/flush_strategy.rs` tested these — leaning entirely on upstream's unit tests for them at the bottlecap boundary is fine, just call that out.

**5. PR description mentions `bottlecap/src/config/lambda_extension.rs`, but the code is appended to the bottom of `mod.rs`.** The 19-field type + 33-test module ends up roughly 570 lines tacked onto an already-2,243-line file. Splitting into a sibling `lambda_extension.rs` would (a) match the description, (b) let `pub mod lambda_extension;` host the `use datadog_agent_config::...` imports without sharing scope with the legacy code, and (c) make the eventual deletion-of-legacy step a much cleaner diff.

**6. Comment at line 1815-1818 mentions colliding with the in-tree `merge_*` macros declared `#[macro_export]`** at the top of the file. Worth noting in the description that the migration plan will delete those `#[macro_export]`s — anyone in bottlecap can still write `crate::merge_string!` today and get the in-tree version. Non-blocking for this PR but easy to forget come PR #1251.

litianningdatadog (approved):

PR #1250, opened 2026-06-10
[SLES-2877] fix(traces): try to pair universal instrumentation on request id for on-demand

shreyamalpani:
## Overview
When the tracer sends the `lambda-runtime-aws-request-id` header (in the start-invocation and end-invocation requests), use request ID based pairing instead of FIFO for both `UniversalInstrumentationStart` and `UniversalInstrumentationEnd` in On Demand mode in addition to Managed Instance mode. FIFO pairing is preserved as a fallback for tracers that do not send the header. 

Motivation: [SLES-2877](https://datadoghq.atlassian.net/browse/SLES-2877)

[SLES-2877]: https://datadoghq.atlassian.net/browse/SLES-2877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -219,34 +219,32 @@ impl Processor {
         self.enhanced_metrics.increment_invocation_metric(timestamp);
         self.enhanced_metrics.set_invoked_received();
 
-        // MANAGED INSTANCE MODE: Check for buffered UniversalInstrumentationStart with request_id
-        if self.aws_config.is_managed_instance_mode() {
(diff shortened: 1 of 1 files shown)


duncanista:
Is there a fallback since old revisions of the SDK doesn't send the `lambda-runtime-aws-request-id`?

litianningdatadog (commented):
LGTM. Could we also update the test coverage for this change?

litianningdatadog (approved):

PR #1253, opened 2026-06-12
fix(metrics): emit init duration metric on platform report as fallback

shreyamalpani:
## Overview
The init duration metric is sent on the `PlatformInitReport` event. In the case of this event not being received / arriving late, this PR adds a fallback to use the init duration from the `PlatformReport` event on cold start invocations in On Demand mode.

## Motivation
Saw failures on emitting the init duration metric in E2E tests.

## Testing
Added unit tests and E2E tests no longer fail on sending the init duration metric on this PR.

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -397,7 +397,11 @@ impl Processor {
         // If the init duration metric has already been emitted, it means the init report event was received late /
         // the platform report event with init duration was received first.
         // In this case, we skip emitting the init duration metric from the platform init report event, to prevent double counting.
-        if !self.init_duration_metric_emitted {
+        if self.init_duration_metric_emitted {
+            debug!(
+                "Skipping init duration metric on PlatformInitReport, metric was already emitted on PlatformReport"
+            );
+        } else {
             self.enhanced_metrics
                 .set_init_duration_metric(init_type, duration_ms, timestamp);
             self.init_duration_metric_emitted = true;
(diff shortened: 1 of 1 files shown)


bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -386,8 +395,15 @@ impl Processor {
...
-            .set_init_duration_metric(init_type, duration_ms, timestamp);
+        // `platform.initReport` is the primary source for the init duration metric and carries the exact `init_type`.
+        // If the init duration metric has already been emitted, it means the init report event was received late /
+        // the platform report event with init duration was received first.
+        // In this case, we skip emitting the init duration metric from the platform init report event, to prevent double counting.
+        if !self.init_duration_metric_emitted {
+            self.enhanced_metrics
+                .set_init_duration_metric(init_type, duration_ms, timestamp);
+            self.init_duration_metric_emitted = true;
+        }
```
purple4reina, line 408: I don't know this repo very well, but will this need locking when reading/setting the `init_duration_metric_emitted` value?

purple4reina (approved):

bottlecap/src/lifecycle/invocation/processor.rs
shreyamalpani: I think we only handle one event at a time, so it should not be needed here

purple4reina: Hmmm, worth looking into a bit I think. For example, in the off chance you get two events at once, what would happen? Good to know what the race condition would potentially produce, and then if we're okay with that, then there's no need for locking.

shreyamalpani: Yup, commands are processed serially by `InvocationProcessorService::run()`, so should be safe. We set the flag `self.awaiting_first_invocation` similarly without any locking

litianningdatadog, line 408: Maybe we should log when self.init_duration_metric_emitted == true, since that may indicate an error case, ie init event comes after the report event

shreyamalpani: I don't think this necessarily indicates an error, just that the events were received out of expected order

litianningdatadog (approved):

PR #1257, opened 2026-06-15
SLES 2889 | Fixing payload size

jchrostek-dd:
https://datadoghq.atlassian.net/browse/SLES-2889

## Overview

Large single-invocation traces could be rejected by the trace intake with a 413, because the extension undercounted how big a payload would be once it left the extension.

 The extension enriches every span after the tracer posts it (adding the tags-provider map), so the protobuf payload it sends is larger than the raw msgpack it received. The trace aggregator's batch cap therefore undercounted the true outbound bytes and packed too much into a single request, whose combined size exceeded the intake limit → 413.

The batch cap was below realistic enriched payload sizes. The cap was set to the old 3.2 MB documented limit. However, from [trace intake config](https://github.com/DataDog/logs-backend/blob/6b17c3834760efe43f0fc2ae09f8c03f893f32fb/domains/intake/apps/intake/src/main/resources/track-types.conf#L269), the limit is actually 15. 

This PR:

- Computes `body_size` from the enriched, protobuf-encoded payload **unconditionally** (not only when on-extension stats are enabled), so the aggregator's batch cap reflects exactly what will be sent.
- Raises the batch cap from 3.2 MB to 12 MB — kept below the trace intake's ~15 MB limit so a large enriched payload flushes in a single batch without a 413.
- Adds debug logs for the post-enrichment payload size and per-batch coalesced size.

## Testing

- **Unit:** new test asserting `body_size` equals the protobuf `encoded_len` of the payload actually sent, and exceeds the (smaller) ingress size — on the default config (no on-extension stats).
- **Integration:** new `payload-size` suite deploys a Node Lambda that emits a ~10 MB single-invocation trace (400 spans carrying large captured payloads) and verifies it: enriches to >10 MB, flushes in a single batch >10 MB, arrives with no 413, and is delivered intact (one trace, root span present, all 400 spans). Run against a locally published extension layer in the serverless sandbox.

### DIFF
--- a/bottlecap/src/traces/trace_aggregator.rs
+++ b/bottlecap/src/traces/trace_aggregator.rs
@@ -1,12 +1,11 @@
 use libdd_trace_utils::send_data::SendDataBuilder;
 use libdd_trace_utils::trace_utils::TracerHeaderTags;
 use std::collections::VecDeque;
+use tracing::debug;
 
-/// Maximum content size per payload uncompressed in bytes,
-/// that the Datadog Trace API accepts. The value is 3.2 MB.
-///
-/// <https://github.com/DataDog/datadog-agent/blob/9d57c10a9eeb3916e661d35dbd23c6e36395a99d/pkg/trace/writer/trace.go#L27-L31>
-pub const MAX_CONTENT_SIZE_BYTES: usize = 3_200_000;
+/// Maximum uncompressed bytes per outbound batch. Kept below the trace intake's
+/// ~15 MB limit so an enriched payload flushes in a single batch.
+pub const MAX_CONTENT_SIZE_BYTES: usize = 12_000_000;
 
 /// Owned version of `TracerHeaderTags<'a>` so it can be stored across async
 /// boundaries without lifetime issues.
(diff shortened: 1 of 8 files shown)


bottlecap/src/traces/trace_aggregator.rs
```diff
@@ -132,6 +131,14 @@ impl TraceAggregator {
...
         }
 
+        if !self.buffer.is_empty() {
+            debug!(
+                "TRACES | batched {} payload(s) totaling {batch_size} bytes (cap {} bytes)",
+                self.buffer.len(),
+                self.max_content_size_bytes
+            );
+        }
+
```
duncanista, line 141: Wonder if libdatadog has something similar?

jchrostek-dd: I don't think they do. If they did, would still prefer for it to be here as well since we use this log message for our integration tests to ensure we are actually testing against a large payload.

duncanista (approved):

lucaspimentel (approved):
LGTM aside from Copilot's comment about the hard-coded `us-east-1` which seems legit.

PR #1265, opened 2026-06-18
feat(dsm): Add support for automatic DSM context extraction inside the extension

jeastham1993:
**Please include Jira ticket in title.**

## Overview

Update the trace context propagation support in universal instrumentation to include the automatic extraction of Data Streams Monitoring context for SQS, SNS, Kinesis and EventBridge.

## Testing 

Added unit tests. Manually tested the functionality for Java, Go and .NET. Java and Go work correctly. [The .NET tracer needs updating to support the 2.x version of `Amazon.Lambda.RuntimeSupport`](https://github.com/DataDog/dd-trace-dotnet/pull/8797) so can't test that manually yet.


bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -330,6 +330,34 @@ async fn extension_loop_active(
...
+    let proxy_aggregator = Arc::new(TokioMutex::new(proxy_aggregator::Aggregator::default()));
+
+    // Extension-side Data Streams Monitoring (consume checkpoints), gated by
+    // DD_DSM_CONSUME_ENABLED.
+    let dsm_processor = if config.dsm_consume_enabled {
+        let service = config
+            .service
+            .clone()
+            .or_else(|| tags_provider.get_canonical_resource_name())
+            .unwrap_or_else(|| "aws.lambda".to_string())
```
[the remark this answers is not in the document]
jeastham1993: The Lambda function name is correct.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -330,6 +330,49 @@ async fn extension_loop_active(
     .await;
 
     let propagator = Arc::new(DatadogCompositePropagator::new(Arc::clone(config)));
+
+    // Shared proxy aggregator (used by the trace agent's proxy endpoints and,
+    // when enabled, the extension-side DSM processor).
+    let proxy_aggregator = Arc::new(TokioMutex::new(proxy_aggregator::Aggregator::default()));
+
+    let dsm_service = config
```
duncanista: cold-start nit: this `dsm_service`/`env`/`version`/`tags` setup (including `to_lowercase()` and the `dsm_tags` build + `sort()`) runs unconditionally, before the `dsm_consume_enabled` gate just below. so every cold start pays for it even with dsm off, which is the default for basically all functions. appsec/otlp/proxy all skip their setup entirely when disabled. could we move this block inside the `if config.ext.dsm_consume_enabled` so disabled functions stay at zero init cost?

bottlecap/src/config/mod.rs
```diff
@@ -81,6 +81,20 @@ pub struct LambdaConfig {
     /// without durable execution context enrichment. Defaults to 0 until the tracer-side
     /// durable execution support is released; set to 50 to re-enable enrichment.
     pub lambda_durable_function_log_buffer_size: usize,
+
+    // Data Streams Monitoring
+    /// Enable extension-side DSM consume checkpoints. Gated by the same
+    /// `DD_DATA_STREAMS_ENABLED` flag the tracer libraries use; the extension
+    /// and tracer never emit checkpoints for the same runtime, so sharing the
+    /// flag cannot double-count.
```
duncanista, line 88: can we spell out how this 'never double-counts' guarantee is actually enforced? the consume hook fires in `process_on_universal_instrumentation_start`, which also runs when a real tracer drives `/lambda/start-invocation`. for a runtime whose tracer already supports lambda dsm, would `DD_DATA_STREAMS_ENABLED=true` turn on both the tracer's dsm and the extension's, giving two consume checkpoints per message? a quick matrix of (runtime x tracer-version x proxy-mode) would make this easier to trust.

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -1088,6 +1103,53 @@ impl Processor {
         context.extracted_span_context =
             Self::extract_span_context(&headers, &payload_value, Arc::clone(&self.propagator));
 
+        // Extension-side DSM: record a consume (`direction:in`) checkpoint for
+        // DSM-eligible event sources, continuing any inbound pathway context.
+        if let Some(dsm) = self.dsm_processor.as_ref() {
```
duncanista, line 1223: does this hook actually fire on managed instances? the usual lmi ordering is `platform.start` creating the context before the tracer's `/lambda/start-invocation`, and `on_invoke_event` skips the enqueue in managed mode. so the later request-id start looks like it'd get buffered by `pair_universal_instrumentation_start_with_request_id` rather than processed, and this block never records a checkpoint. could you confirm the lmi path emits dsm, and that the new lmi test actually exercises it?

bottlecap/src/traces/data_streams/processor.rs
```diff
@@ -0,0 +1,237 @@
...
+
+/// Extract the inbound DSM pathway context from a carrier, preferring the
+/// base64 key. Fails closed (returns `None`) on malformed input.
+fn extract_pathway_context(
+    carrier: &std::collections::HashMap<String, String>,
+) -> Option<DsmContext> {
+    carrier
+        .get(DD_PATHWAY_CTX_BASE64_KEY)
+        .or_else(|| carrier.get(DD_PATHWAY_CTX_KEY))
+        .and_then(|v| DsmContext::from_base64(v))
```
duncanista, line 164: the legacy `dd-pathway-ctx` fallback is decoded with `from_base64`, but by dd-trace convention that key carries raw binary (it's the kafka/msk propagation form), and this pr adds msk consume checkpoints. for msk this fallback will fail to decode and start a fresh pathway instead of continuing the inbound one. should the legacy key go through `from_bytes`, or are msk headers guaranteed to arrive base64 via `headers_to_string_map`? if the latter, worth a quick comment saying so.

bottlecap/src/traces/data_streams/sketch.rs
```diff
@@ -0,0 +1,341 @@
...
+            let key = self.mapping.key(-value);
+            self.negative_store.add(key, 1.0);
+        } else {
+            self.zero_count += 1.0;
+        }
+    }
+
+    /// Serialize to the `DDSketch` protobuf wire format.
+    #[must_use]
+    pub fn to_proto_bytes(&self) -> Vec<u8> {
```
duncanista, line 251: we already depend on `libdd-ddsketch` (trace stats serialize ddsketch protos through libdatadog today). did you evaluate reusing it here before hand-rolling the sketch + the golden fixture? if it can emit the same `LogCollapsingLowestDenseDDSketch` bytes the pipeline_stats intake expects, we'd drop ~340 lines plus the compat-drift risk against `@datadog/sketches-js`. if it genuinely can't (different store/gamma), a one-line note here on why reuse was rejected would save the next reader the dig.

duncanista (commented):
a few things worth digging into before merge, mostly architecture and correctness plus one cold-start nit. nothing blocking that can't be sorted in-thread, inline below.

bottlecap/src/config/mod.rs
jeastham1993: Done, comment updated.

bottlecap/src/lifecycle/invocation/processor.rs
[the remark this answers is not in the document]
jeastham1993: Fixed.

bottlecap/src/traces/data_streams/sketch.rs
jeastham1993: Looking at the code (and asking Claude to do some analysis) it seems it doesn't expose the right configuration, so I've updated the comment. In the future I could chat with the DSM team though and maybe look at adding something to libddd

bottlecap/src/lifecycle/invocation/processor.rs
jeastham1993: Fixed in the latest revision. In Managed Instance mode, /lambda/start-invocation now processes immediately when the request-id context already exists, instead of relying only on the FIFO pairing queue. That matches the normal LMI ordering where the platform invoke event creates the context first.

I added a regression test covering this path with DSM enabled, and verified with `cd bottlecap && cargo test --lib`

jeastham1993:
@duncanista think that's everything addressed 👍

bottlecap/src/proxy/interceptor.rs
```diff
@@ -196,14 +221,37 @@ async fn invocation_next_proxy(
...
-            if aws_config.aws_lwa_proxy_lambda_runtime_api.is_some() {
+            // Drive full LWA universal instrumentation only for LWA and the
+            // experimental wrapper proxy. DSM-only proxy support must not reuse
+            // `lwa::process_invocation_next`, because that path also queues LWA
+            // reparenting with a synthetic invocation span id. In tracer runtimes
+            // that still call `/lambda/start-invocation`, that synthetic id can
+            // conflict with the tracer-provided invocation span id.
+            let experimental_proxy_enabled = std::env::var("DD_EXPERIMENTAL_ENABLE_PROXY")
+                .is_ok_and(|v| v.eq_ignore_ascii_case("true"));
+            let dsm_consume_enabled = std::env::var("DD_DATA_STREAMS_ENABLED")
```
duncanista: blocker. two things here. this reads the flag straight from env with a strict "true" check, but the config path uses the lenient bool parser (accepts 1, yes, TRUE). so DD_DATA_STREAMS_ENABLED=1 builds the processor and runs the flusher, but silently skips this proxy branch, so dsm goes missing on that path with no error. it also does an env::var read on every invocation on the hot path. can we thread config.ext.dsm_consume_enabled into the interceptor state and use that? that fixes the mismatch and the per-invocation read in one go.

bottlecap/src/proxy/interceptor.rs
```diff
@@ -112,6 +112,31 @@ async fn graceful_shutdown(tasks: Arc<Mutex<JoinSet<()>>>, shutdown_token: Cance
...
+        .get("Lambda-Runtime-Aws-Request-Id")
+        .and_then(|v| v.to_str().ok())
+        .map(std::string::ToString::to_string)
+    else {
+        debug!("PROXY | invocation_next_proxy | missing request id for DSM consume extraction");
+        return;
+    };
+
+    let payload_value = serde_json::from_slice(body).unwrap_or_else(|e| {
+        error!("PROXY | invocation_next_proxy | error parsing DSM payload as JSON: {e}");
```
duncanista: nit, not a blocker. non-json payloads are fairly common, and this logs at error for each one. since it falls back to {} and just skips, debug feels more appropriate here.

bottlecap/src/lifecycle/invocation/triggers/sns_event.rs
```diff
@@ -165,6 +166,23 @@ impl Trigger for SnsRecord {
     fn is_async(&self) -> bool {
         true
     }
+
+    fn get_dsm_edge_tags(&self) -> Option<Vec<String>> {
+        // SNS uses the full topic ARN as the topic tag (matches dd-trace-js).
+        Some(vec![
+            "direction:in".to_string(),
+            format!("topic:{}", self.sns.topic_arn),
+            "type:sns".to_string(),
```
duncanista, line 175: blocker, needs confirmation before merge. dd-trace-js emits sns only as a produce checkpoint (type:sns) and keys the consume on the receiving queue as type:sqs, so it does not emit a type:sns consume checkpoint. stitching still works via parent_hash, but a type:sns consume node here may not line up with the rest of the fleet. can we confirm type:sns consume nodes match the dsm backend convention and dd-trace-go/java before this ships?

bottlecap/src/lifecycle/invocation/triggers/event_bridge_event.rs
```diff
@@ -111,6 +111,47 @@ impl Trigger for EventBridgeEvent {
     fn is_async(&self) -> bool {
         true
     }
+
+    fn get_payload_size_bytes(&self) -> f64 {
+        // Measure the serialized JSON byte length of the event detail object.
+        serde_json::to_string(&self.detail).map_or(0.0, |s| s.len() as f64)
+    }
+
+    fn get_dsm_edge_tags(&self) -> Option<Vec<String>> {
```
duncanista, line 120: blocker, needs confirmation before merge. dd-trace-js emits no dsm checkpoint for eventbridge, so type:eventbridge is extension-specific here. also, if nothing injects dd-pathway-ctx into eventbridge events, these are always pathway entry points (parent=0) with nothing upstream to stitch to. can we confirm the type:eventbridge and exchange conventions against the dsm backend and other tracers before this ships?

bottlecap/src/config/mod.rs
```diff
@@ -81,6 +81,23 @@ pub struct LambdaConfig {
...
```
duncanista, line 88: blocker. the no-double-count claim holds inside the extension via request_id dedup, but nothing in code stops the extension emitting alongside a self-instrumenting python/js wrapper. it relies on those runtimes not enabling extension dsm. can we confirm that assumption holds, or enforce it in code? otherwise dsm stats get double-counted for those runtimes.

bottlecap/src/traces/data_streams/sketch.rs
```diff
@@ -0,0 +1,357 @@
...
+}
+
+impl LogarithmicMapping {
+    fn new(relative_accuracy: f64) -> Self {
+        let i = 2.0 * relative_accuracy / (1.0 - relative_accuracy);
+        let gamma = 1.0 + i;
+        // KeyMapping multiplier is 1/ln1p(i); LogarithmicMapping then * ln(2).
+        let multiplier = std::f64::consts::LN_2 / i.ln_1p();
+        // KeyMapping.minPossible = MIN_NORMAL * gamma (MIN_NORMAL = f64::MIN_POSITIVE).
+        let min_possible = f64::MIN_POSITIVE * gamma;
```
duncanista: nit, not a blocker. sketches-js uses MIN_SAFE_FLOAT (2^-1023) then * gamma for minPossible, this uses f64::MIN_POSITIVE (2^-1022), which is 2x larger. no impact for dsm values (all >= 1e-9 or exactly 0, and the goldens pass), but the MIN_NORMAL comment reads as if it matches the reference.

bottlecap/src/traces/data_streams/sketch.rs
```diff
@@ -0,0 +1,357 @@
+//! A byte-for-byte port of the tracer's `LogCollapsingLowestDenseDDSketch`
+//! (relative accuracy 0.01, bin limit 2048) and its protobuf serialization.
+//!
+//! DSM's `EdgeLatency` / `PathwayLatency` / `PayloadSize` fields are the raw
+//! `DDSketch` protobuf bytes produced by `@datadog/sketches-js`. To stay
+//! compatible we must reproduce both the binning *and* the dense-store layout
+//! exactly (including the bin-centering that pads each chunk with zeros).
+//!
+//! Verified against fixtures generated by the real tracer in
+//! `fixtures/sketch_golden.json` (see `tools/dsm/gen_sketch_golden.js`).
```
duncanista: nit, not a blocker. this points at tools/dsm/gen_sketch_golden.js but that file isn't in the repo, so regenerating the goldens later will be hard. could we commit the generator or drop the reference?

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -330,6 +330,54 @@ async fn extension_loop_active(
...
+
+    // Extension-side Data Streams Monitoring (consume checkpoints), gated by
+    // DD_DATA_STREAMS_ENABLED.
+    let dsm_processor = if config.ext.dsm_consume_enabled {
+        let dsm_service = config
+            .service
+            .clone()
+            .or_else(|| tags_provider.get_canonical_resource_name())
+            .unwrap_or_else(|| "aws.lambda".to_string())
+            .to_lowercase();
```
duncanista: nit, not a blocker. we lowercase the dsm service here. worth confirming this matches the tracer service normalization, otherwise the same service could show up twice in dsm (MyService vs myservice).

duncanista (commented):
nice work, this is a clean feature and the compatibility-sensitive parts are well isolated. i checked the core against dd-trace-js and the fixtures, and verified the utf8 lossy round-trip matches node over a 50k fuzz, so the pathway hashes line up. none of the notes below are blockers. mostly one flag-parsing consistency thing, a couple of confirmations on the sns/eventbridge consume tags, and small nits.

bottlecap/src/proxy/interceptor.rs
duncanista: quick context on why this is env and not config: the interceptor state only carries AwsConfig, and Config (which holds ext.dsm_consume_enabled) is never threaded into interceptor::start, so the handler has to fall back to env::var. the neighbouring DD_EXPERIMENTAL_ENABLE_PROXY read is env-only by design, but dsm already has a parsed config field, so re-reading env here parses it with a stricter rule, which is the =1 mismatch. start_api_runtime_proxy already has config in scope, so threading dsm_consume_enabled (or Arc<Config>) through into the state and reading it from there would drop the per-invocation env::var and make the gating consistent in one change.

bottlecap/src/lifecycle/invocation/triggers/sns_event.rs
jeastham1993: Checked against datadog-lambda-js - https://github.com/DataDog/datadog-lambda-js/blob/main/src/trace/context/extractors/sns.ts

bottlecap/src/lifecycle/invocation/triggers/event_bridge_event.rs
jeastham1993: @duncanista I've got open PR's against dd-trace-js and dd-lambda-js (as we ll as all the other tracers, some merged some not) to standardise this across the board. It's been confirmed with the DSM team as the correct shape. LMK if you want that in a code comment.

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -1025,7 +1053,15 @@ impl Processor {
...
-            if self
+            if self.aws_config.is_managed_instance_mode()
+                && self.context_buffer.get(&req_id).is_some()
+            {
+                // Managed Instance mode creates the invocation context from the
+                // platform invoke event without enqueueing the request id in the
+                // FIFO pairing queue. If the request-id-addressed context already
+                // exists, process immediately instead of buffering forever.
+                self.process_on_universal_instrumentation_start(req_id, headers, payload_value);
+            } else if self
```
litianningdatadog, line 1125: LMI runs concurrent invocations, but `self.inferrer` is shared `Processor` state. A later invocation's `infer_span` call can overwrite an earlier invocation's data before it's read, so the earlier invocation ends up completing its span with the wrong trigger/inference data. Move the inferred span into each invocation's own `Context` instead of storing it on `Processor`.

bottlecap/src/traces/data_streams/processor.rs
```diff
@@ -0,0 +1,276 @@
...
+        proxy_aggregator: Arc<TokioMutex<ProxyAggregator>>,
+    ) -> Self {
+        let aggregator =
+            Aggregator::new(service.clone(), env.clone(), tracer_version, version, tags);
+        Self {
+            service,
+            env,
+            aggregator: Mutex::new(aggregator),
+            proxy_aggregator,
+            target_url: format!("https://trace.agent.{site}/api/v0.1/pipeline_stats"),
```
litianningdatadog: This target URL no longer honors custom endpoint defined via DD_APM_DD_URL (see [here](https://github.com/DataDog/datadog-lambda-extension/blob/088fd479f8900143df641d8fb3595b74ea5df852/bottlecap/src/traces/trace_processor.rs#L418)). Is this expected?

bottlecap/src/bin/bottlecap/main.rs
litianningdatadog, line 1498: The `DsmOnly` branch in `interceptor.rs` only executes if the runtime API proxy is running. But this function doesn't seem to consider DSM enablement. Is this by design or I missed something?

litianningdatadog (changes requested):

litianningdatadog (approved):

PR #1266, opened 2026-06-18
refactor(config): rename compute_trace_stats_on_extension -> lambda_extension_compute_stats

duncanista:
## Summary

Renames the public env var \`DD_COMPUTE_TRACE_STATS_ON_EXTENSION\` to \`DD_LAMBDA_EXTENSION_COMPUTE_STATS\`, plus the matching Rust field, so the name reflects the actual scope: this toggle only governs the Lambda extension's APM trace-stats pipeline and is not a generic Datadog Agent concept.

## Why

Auditing the \`LambdaConfig\` extension fields surfaced this one as the one with an ambiguous env var name. \"Compute trace stats on extension\" reads as if there's a generic \"extension\" concept; in practice it's specifically the Lambda extension. Making that explicit avoids confusion if/when the same idea surfaces in other embedders (which would then need a differently-named knob).

## Changes

- \`LambdaConfigSource\` field renamed: \`compute_trace_stats_on_extension\` → \`lambda_extension_compute_stats\`. Figment's \`DD_\` prefix maps to the new env var automatically.
- \`LambdaConfig\` field renamed in lockstep. All call sites in \`trace_processor.rs\`, \`otlp/agent.rs\`, \`lifecycle/invocation/processor.rs\`, \`tags/lambda/tags.rs\`, and the integration test now read \`config.ext.lambda_extension_compute_stats\`.
- \`merge_from\` keeps the field in the existing \`value:\` group of \`merge_fields!\` — no source-to-config rename needed since field names match.
- **No back-compat alias** for the old env var. The \`.ext\` migration this depends on has not shipped, so the legacy name was never released externally.
- Test renamed; default-false test added.

## Test plan

- [x] \`cargo test\` — \`config::lambda_config_tests::lambda_extension_compute_stats_from_env\` and \`lambda_extension_compute_stats_defaults_false\` pass; 38/38 in the module.
- [x] \`cargo clippy --workspace --all-targets --features default\` — clean.
- [x] \`cargo fmt -- --check\` — clean.

## Notes for review

Stacked on top of \`feat/migrate-bottlecap-to-upstream-config\` (the .ext migration PR). Merges into that branch.

### DIFF
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -65,7 +65,7 @@ pub struct LambdaConfig {
     pub lambda_proc_enhanced_metrics: bool,
     pub capture_lambda_payload: bool,
     pub capture_lambda_payload_max_depth: u32,
-    pub compute_trace_stats_on_extension: bool,
+    pub lambda_extension_compute_stats: bool,
     pub span_dedup_timeout: Option<Duration>,
     pub api_key_secret_reload_interval: Option<Duration>,
     pub dd_org_uuid: String,
@@ -95,7 +95,7 @@ impl Default for LambdaConfig {
             lambda_proc_enhanced_metrics: true,
             capture_lambda_payload: false,
             capture_lambda_payload_max_depth: 10,
-            compute_trace_stats_on_extension: false,
+            lambda_extension_compute_stats: false,
             span_dedup_timeout: None,
             api_key_secret_reload_interval: None,
             dd_org_uuid: String::new(),
@@ -144,8 +144,10 @@ pub struct LambdaConfigSource {
     pub capture_lambda_payload: Option<bool>,
     #[serde(deserialize_with = "deser_opt_lossless")]
     pub capture_lambda_payload_max_depth: Option<u32>,
+    /// `DD_LAMBDA_EXTENSION_COMPUTE_STATS` — when true, the extension computes
+    /// APM trace stats locally instead of letting the backend do it.
     #[serde(deserialize_with = "deser_opt_bool")]
-    pub compute_trace_stats_on_extension: Option<bool>,
+    pub lambda_extension_compute_stats: Option<bool>,
 
     #[serde(deserialize_with = "deser_dur_secs_ignore_zero")]
     pub span_dedup_timeout: Option<Duration>,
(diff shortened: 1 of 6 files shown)


bottlecap/src/config/mod.rs
```diff
@@ -144,8 +144,12 @@ pub struct LambdaConfigSource {
     pub capture_lambda_payload: Option<bool>,
     #[serde(deserialize_with = "deser_opt_lossless")]
     pub capture_lambda_payload_max_depth: Option<u32>,
+    /// `DD_LAMBDA_EXTENSION_COMPUTE_STATS` — when true, the extension computes
+    /// APM trace stats locally instead of letting the backend do it. Renamed
+    /// from the legacy `DD_COMPUTE_TRACE_STATS_ON_EXTENSION` to reflect that
+    /// this is a Lambda-extension-specific concern.
     #[serde(deserialize_with = "deser_opt_bool")]
-    pub compute_trace_stats_on_extension: Option<bool>,
+    pub lambda_extension_compute_stats: Option<bool>,
```
lucaspimentel, line 150: Is there's anyone already using `DD_COMPUTE_TRACE_STATS_ON_EXTENSION` we should keep it as a fallback to avoid breaking changes.

lucaspimentel (approved):
LGTM aside from the question about keeping `DD_COMPUTE_TRACE_STATS_ON_EXTENSION` as a fallback for backwards compatibility.

bottlecap/src/config/mod.rs
duncanista: this has never been released, so safe to rename before releasing

lym953:
I mentioned this env var in https://github.com/DataDog/datadog-lambda-extension/issues/688, so you may need to communicate there.

bottlecap/src/config/mod.rs
lucaspimentel: Did we not include the fallback? This setting was available for several extension releases (at least 9 months).

lucaspimentel:
> I mentioned this env var in https://github.com/DataDog/datadog-lambda-extension/issues/688, so you may need to communicate there.

I posted a comment there, thanks @lym953.

PR #1280, opened 2026-06-24
refactor(build): feature-gate OTLP and AppSec subsystems (default-on)

duncanista:
> [!NOTE]
> **DRAFT — stacked on #1271** (`jordan.gonzalez/cold-start-instrumentation/feature`). Review/merge that PR first; this diff is scoped to the build/feature-gating change on top of it.
>
> **Jira:** none yet — add before marking ready.

## Overview

Puts the `otlp` and `appsec` subsystems behind Cargo features that are **ON by default**, so the shipped binary is byte-for-byte unchanged, but can be turned off to drop their heavy, *exclusive* dependencies. This is part of the cold-start dependency/binary-size reduction work (Confluence H13).

**Deps gated**

- `otlp` feature → `opentelemetry-proto`, `opentelemetry-semantic-conventions`, `tonic`, `tonic-types` (verified otlp-only).
- `appsec` feature → `libddwaf` (and transitively `libddwaf-sys` + its C build toolchain — `bindgen`/`clang-sys`/`cmake`/`nom`/`tar`/… — plus `aws-lc-rs`/`aws-lc-sys`).

**Deliberately NOT gated**

- `prost` — shared: used by `proxy/interceptor.rs`, `traces/trace_processor.rs`, and `datadog-protos`. Left as a normal dependency (verified by grep before gating).

**How the gating works**

- `src/lib.rs`: `#[cfg(feature = "otlp")] pub mod otlp;` and `#[cfg(feature = "appsec")] pub mod appsec;`.
- When `appsec` is off, an uninhabited `AppSecProcessorStub` enum stands in for the processor type, so the `Option<Arc<Mutex<…>>>` fields (`SendingTraceProcessor`, `TraceAgent`), the proxy `InterceptorState` tuple, and every `appsec: None` construction site keep the same shape across both builds. Only the code paths that actually call into the WAF (`process_span` / `process_invocation_next` / `process_invocation_result` / hold-trace) are `#[cfg(feature = "appsec")]`-gated. The stub is never constructed (the `Option` is always `None`).
- `start_otlp_agent` gets a no-op `#[cfg(not(feature = "otlp"))]` fallback that returns `None`, so its call site and shutdown wiring are unchanged.
- `fips` now implies `appsec` (FIPS-mode WAF requires `libddwaf` compiled in); `fips` feature resolution verified to still pull in `libddwaf`.

**Dependency-graph delta** (proxy for binary size — a precise stripped/LTO/cross-target binary measurement was not run, it is expensive): turning off `otlp`+`appsec` drops **26 crates** from the graph (396 → 370), including the native-compiled `libddwaf-sys` + `aws-lc-sys` and the `bindgen`/`clang`/`cmake` build chain.

## Testing

Both builds are clippy-clean (pedantic + `unwrap_used` denied; only the pre-existing accepted `buf_redux`/`multipart` future-incompat note remains):

- **Default build (otlp + appsec on):**
  `cargo clippy --bin bottlecap --no-deps` → exit 0
  `cargo clippy --all-targets --no-deps` (compiles tests) → exit 0
- **Feature-OFF build (everything in `default` EXCEPT otlp/appsec) — verified:**
  ```
  cargo clippy --bin bottlecap --no-deps --no-default-features \
    --features="reqwest/rustls-tls-native-roots,datadog-fips/default,datadog-agent-config/https,libdd-common/https,libdd-trace-utils/https,libdd-trace-obfuscation/https,libdd-trace-stats/https"
  ```
  → exit 0 (`libddwaf`/`libddwaf-sys` confirmed absent from the dependency graph; direct `opentelemetry-proto`/`tonic-types` dropped).

### DIFF
--- a/bottlecap/Cargo.toml
+++ b/bottlecap/Cargo.toml
@@ -48,14 +48,17 @@ rustls-pki-types = { version = "1.0", default-features = false }
 hyper-rustls = { version = "0.27.7", default-features = false }
 rand = { version = "0.8", default-features = false }
 prost = { version = "0.14", default-features = false }
-tonic = { version = "0.14", features = ["transport", "codegen", "server", "channel", "router"], default-features = false }
-tonic-types = { version = "0.14", default-features = false }
+# otlp-only (also pulled transitively via datadog-protos, but the OTLP gRPC server
+# needs them as direct deps). Gated behind the `otlp` feature.
+tonic = { version = "0.14", features = ["transport", "codegen", "server", "channel", "router"], default-features = false, optional = true }
+tonic-types = { version = "0.14", default-features = false, optional = true }
 zstd = { version = "0.13.3", default-features = false }
 futures = { version = "0.3.31", default-features = false }
 serde-aux = { version = "4.7", default-features = false }
 serde_html_form = { version = "0.2", default-features = false }
-opentelemetry-proto = { version = "0.31.0", features = ["trace", "with-serde", "gen-tonic"] }
-opentelemetry-semantic-conventions = { version = "0.30", features = ["semconv_experimental"] }
+# otlp-only. Gated behind the `otlp` feature.
+opentelemetry-proto = { version = "0.31.0", features = ["trace", "with-serde", "gen-tonic"], optional = true }
+opentelemetry-semantic-conventions = { version = "0.30", features = ["semconv_experimental"], optional = true }
 # Pinned to <0.8.3: version 0.8.3 upgraded to openssl-probe 0.2.x which scans all cert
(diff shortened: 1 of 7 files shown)


duncanista:
Closing: the extension always ships as a single binary with all subsystems compiled in, and OTLP/AppSec are enabled at runtime via env vars — so compile-time feature-gating provides no value for the shipped product. The disabled subsystems also do no eager work at init, so there's no cold-start win either (only a build-time/size effect that we never actually take advantage of given the single-artifact shipping model). Reopen if the shipping model ever changes to slim/per-feature binaries.

PR #1284, opened 2026-06-24
Cold-start improvements — integration (combined testing)

duncanista:
**Jira:** none yet — add before marking ready.

> **DRAFT / DO NOT MERGE AS-IS.** This is an *umbrella* PR that merges every individual cold-start PR onto one branch so the set can be built and benchmarked together in CI as a single unit. Each change lands (and is reviewed) in its own PR; this branch exists only for combined testing. It is **not** intended to merge — once the constituent PRs land on `main` this branch is discarded.

## Overview

Integration branch (`jordan.gonzalez/cold-start-integration/feature`) built on `origin/main`, merging the full cold-start improvement stack. Conflicts were resolved by taking the **union** of every PR's intent (no change was dropped). Merge order was chosen to minimize conflict pain (build PRs first, then deps, then the `main.rs` init-path PRs last):

Included PRs:

- [x] H0 #1271 — init-phase timing instrumentation (`log_init_checkpoint` + checkpoints in `main.rs`); the base all others were stacked on
- [x] H16 #1272 — jemalloc `narenas:1` (`ENV JEMALLOC_SYS_WITH_MALLOC_CONF`) in both compile Dockerfiles
- [x] H7 #1273 — eager symbol binding (`-Clink-arg=-Wl,-z,now -Clink-arg=-Wl,-z,relro`)
- [x] H8 #1274 — `-Ctarget-cpu` per platform (`neoverse-n1` on arm64, `x86-64-v2` on x86_64)
- [x] H10 #1275 — build hygiene (`--default-toolchain none`, drop `+stable` / `--component rust-src`, remove dead UPX install in `build_layer`)
- [x] H1 #1276 — share TLS roots: default feature `reqwest/rustls-tls-webpki-roots` (FIPS still uses native-roots)
- [x] H12 #1279 — unify duplicate crate versions (nix 0.29, thiserror 2, opentelemetry-semantic-conventions 0.31, rand 0.9 + rand-0.9 API fix in `lifecycle/invocation/mod.rs`)
- [x] H15 #1277 — right-size Tokio runtime: explicit `Builder` from the Lambda memory tier; body moved into `async fn run()`; `tokio_worker_threads()` helper
- [x] H4 #1282 — parallelize init HTTP client construction: register/`/next` client build + `register()` moved into a background task returning `(client, RegisterResponse)`
- [x] H2 #1281 — issue telemetry `subscribe` earlier, overlapping service construction via `tokio::join!`; `TelemetryListener::start()` is now `async` with a synchronous bind
[...]

PR #1285, opened 2026-06-24
perf(init): issue telemetry subscribe concurrently with first /next (H18 experiment)

duncanista:
**Jira:** none yet — add before marking ready.

> [!WARNING]
> **EXPERIMENT.** This is an intentionally experimental cold-start change (hypothesis **H18**) to be benchmarked later. It is stacked on the cold-start integration umbrella PR #1284 — base branch is `jordan.gonzalez/cold-start-integration/feature`, **not** `main`. Do not merge to `main` directly; review/merge into the integration branch.

## Overview

Take the Telemetry API `subscribe` round-trip **off the cold-start critical path** by issuing it concurrently with the first `/next`, instead of awaiting its `200 OK` before the extension reaches `extension::next_event` (`/next`).

**Mechanism.** The Lambda Extensions API ends the extension's INIT phase at the first `/next` call. That call then long-polls until the first INVOKE arrives (≫ the ~25 ms subscribe round-trip, which our per-phase init checkpoints measured as the single dominant init phase). Previously the subscribe was awaited to its `200 OK` *before* `/next`, serializing it on the critical path. This change issues `telemetry::subscribe(...)` in a detached `tokio::spawn` and lets init proceed straight to `/next`, so the subscribe HTTP round-trip overlaps the `/next` long-poll wait rather than running ahead of it. Target: shave the dominant ~25 ms init phase off `Init Duration` / ext-ready-in.

**What changed** (`bottlecap/src/bin/bottlecap/main.rs`, `setup_telemetry_client`):
- **Kept** binding the telemetry listener socket synchronously — `listener.start().await` still awaits the bind before we proceed, so the socket is already accepting connections (no early telemetry lost to a *closed port*).
- **Kept** the listener cancel token wired exactly as before: still obtained via `listener.cancel_token()` and returned unchanged, so graceful shutdown of the spawned serve task and both shutdown paths are untouched.
- **Moved** `telemetry::subscribe(...)` into a detached `tokio::spawn` (owned `client`/`runtime_api`/`extension_id`), with its `Result` handled inside the task via `error!`. It is no longer awaited before `/next`.
- The function signature is unchanged, so the existing `tokio::join!(telemetry_setup, build_services)` (from H2) and all downstream wiring are identical. **Both On-Demand and Managed-Instance paths are preserved** — they share the same `telemetry_listener_cancel_token`.
- Updated the call-site doc comment to describe the reorder and reference the open risk.

This is a behavior change limited to **when** the subscribe completes relative to `/next`. Everything else (logs channel, all services, cancel-token shutdown wiring) is unchanged.

### Relationship to H2
H2 (already on the integration branch) overlaps the subscribe with service *construction* via `tokio::join!`. H18 goes further: it removes the subscribe from the critical path entirely by not awaiting it before `/next`. The two compose cleanly — the bind still runs concurrently with construction; only the subscribe await is removed.

### OPEN RISK
We deliberately do **not** solve the correctness question here. When the subscription registers slightly *after* `/next`, the platform may already have emitted `platform.initStart` / `platform.initReport` and early function logs the moment INIT ended — those could be delivered before the subscription exists and be **lost**. The listener socket is bound before we return, so nothing is dropped for a closed-port reason; the only exposure is the platform-side registration race. Closing (or proving harmless) that race is exactly what the benchmark below is for.

## Testing

Validation plan (to run before this is marked ready) — on a minimal `provided.al2023` / Go function:
- Confirm `aws.lambda.enhanced.init_duration` is **still emitted** (i.e. `platform.initReport` is still received after the reordered subscribe).
- Confirm **no** `platform.logsDropped` telemetry event is observed (no early/init logs lost to the registration race).
- Measure the cold-start delta: compare reported **Init Duration** and the "Datadog Next-Gen Extension ready in …ms" / `INIT | phase=...` checkpoints against the integration branch baseline to quantify the savings from overlapping the subscribe with the `/next` long-poll.

If init telemetry is lost or `platform.logsDropped` appears, this hypothesis is rejected (or needs a guard) and should not advance.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -429,20 +429,23 @@ async fn extension_loop_active(
     let propagator = Arc::new(DatadogCompositePropagator::new(Arc::clone(config)));
 
     // The Lambda Extensions API ends the INIT phase at the first `/next` call, so the
-    // serialized critical path before that call directly inflates cold-start time. The
-    // telemetry subscription only depends on `logs_agent_channel` (already available),
-    // and its `subscribe` HTTP round-trip is independent of constructing the trace agent,
-    // AppSec, API-runtime proxy, and lifecycle listener below. We therefore issue the
-    // subscribe and run that remaining (synchronous) service construction concurrently
-    // via `tokio::join!`: the subscribe future is polled first so its network round-trip
-    // is in flight while construction proceeds, instead of being serialized behind it.
+    // serialized critical path before that call directly inflates cold-start time.
+    // `setup_telemetry_client` binds the telemetry listener socket synchronously (it awaits
+    // the bind — see `TelemetryListener::start`) so the socket is already accepting
+    // connections, then — per H18 — issues the `subscribe` HTTP round-trip in a detached
+    // task and returns the cancel token WITHOUT awaiting the subscribe. We run that bind
+    // concurrently with the remaining (synchronous) service construction (trace agent,
+    // AppSec, API-runtime proxy, lifecycle listener) via `tokio::join!`. Because the
+    // subscribe is no longer awaited here, neither it nor service construction blocks the
+    // path to the first `/next`: the subscribe round-trip now overlaps the `/next`
+    // long-poll wait instead of being serialized ahead of it.
     //
-    // Correctness: `setup_telemetry_client` binds the telemetry listener socket
-    // (synchronously, before `subscribe` returns — see `TelemetryListener::start`) prior
-    // to subscribing, so the listener is already accepting connections when the Telemetry
-    // API begins delivering events. No early `platform.initStart`/`initReport` or logs are
-    // dropped. The construction branch only spawns background tasks and returns handles;
-    // it performs no I/O that the subscribe depends on, so the two are order-independent.
+    // Correctness: the listener socket is bound before this returns, so the listener is
+    // serving when the Telemetry API begins delivering events — nothing is dropped for a
+    // closed-port reason, and the cancel token still drives graceful shutdown. The OPEN
+    // RISK that the subscription may register slightly after `/next` (potentially missing
+    // early `platform.initStart`/`initReport`/logs) is documented in `setup_telemetry_client`
+    // and is what the H18 experiment will measure.
     let telemetry_setup = setup_telemetry_client(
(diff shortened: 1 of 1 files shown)


duncanista:
## Tested — recommend not pursuing H18

Built this branch into an arm64 layer and benchmarked it against the integration baseline (#1284) and prod **v98**, across **python**, **go**, and **rust** (`provided.al2023`) runtimes, forcing cold starts via the env-var bump trick.

**Finding: H18 makes the extension's own init faster, but does *not* reduce total `Init Duration` on any runtime — including the most minimal one (rust), where the extension genuinely is the cold-start long pole.**

Rust, n=30 (the decisive case):
| layer | Init Duration p50 (ms) | ext `ready_in` p50 (ms) |
|---|--:|--:|
| integration (#1284) | 100.4 | 19 |
| **+H18** | **97.8** (−2.6, within noise) | **12** |

Same pattern on python (184 → 187, flat) and go (flat/noisy at n=17). So H18 *does* do what it was designed to — `ready_in` drops as the subscribe moves off the path to `/next` — but it buys ~nothing in the **customer-visible** cold start.

**Why:** the `Init Duration` floor (min ~80 ms on rust) is **sandbox provisioning + loading/linking the ~15 MB extension binary**, not the extension's async init and not the runtime (rust runtime is ~16 ms). Once the integration changes already drop `ready_in` to ~19 ms, the extension sits at/below that floor — so taking the subscribe off the path has nothing left to give.

**Plus an unresolved correctness risk:** this change fires `telemetry::subscribe` detached (not awaited before `/next`), so the subscription can register *after* the platform emits `platform.initStart`/`initReport`/early logs → potential loss of the `aws.lambda.enhanced.init_duration` metric / `platform.logsDropped`. I never verified this, because the perf result already makes it not worth shipping.

**Recommendation: close / don't pursue.** No Init-Duration payoff on any runtime tested, plus a telemetry-correctness risk. The remaining cold-start lever for minimal runtimes is the **extension binary's load time** (build/link: eager symbol binding, opt-level/size, etc.) — not further init-sequence reordering.

_(Benchmark detail: identical functions per runtime, arm64 / 1024 MB / `DD_LOG_LEVEL=debug`, cold starts forced by bumping a dummy env var. Companion results on the integration PR #1284.)_

PR #1293, opened 2026-07-02
feat(config): add DD_SERVERLESS_APM_ONLY traces-only mode

zarirhamza:
## Summary

Adds a single master switch — **`DD_SERVERLESS_APM_ONLY`** — that puts the extension in **APM-only ("traces only") mode**. When enabled, every billable metrics and logs egress path is suppressed so the customer incurs no infrastructure-monitoring or log-ingestion charges, while **traces and APM trace stats are unaffected**.

This closes the gap that made a true "traces only" configuration impossible before: there was no toggle to disable **custom DogStatsD metrics**, so even with logs and enhanced metrics turned off, custom metrics could still egress. A single, defensive flag now guarantees the behavior.

- Default is `false` — existing behavior is completely preserved.
- Readable from both environment (`DD_SERVERLESS_APM_ONLY`) and YAML.

## Behavior

| Signal | APM-only mode |
|---|---|
| Traces / APM trace stats | **sent** (unchanged) |
| Enhanced metrics | suppressed |
| Process (proc) enhanced metrics | suppressed |
| Custom DogStatsD metrics | suppressed |
| OTLP metrics | suppressed |
| Logs (incl. OTLP logs) | suppressed |

## Implementation (on the Config<LambdaConfig> model)

- `LambdaConfig` / `LambdaConfigSource` gain a `serverless_apm_only` field, merged from both env (`DD_SERVERLESS_APM_ONLY`) and YAML via `merge_fields!`. Defaults to `false`.
- `get_config()` applies a post-merge override (`apply_serverless_apm_only`) that forces `serverless_logs_enabled`, `enhanced_metrics`, `lambda_proc_enhanced_metrics` (on `.ext`) and the core `otlp_config_metrics_enabled` / `otlp_config_logs_enabled` toggles off — overriding any individually set values so the guarantee holds regardless of other configuration.

## Enforcement (defense in depth)

1. **Config** (`config/mod.rs::apply_serverless_apm_only`) — forces the logs/metrics/OTLP toggles off after the shared env/yaml merge.
2. **Metrics wiring** (`main.rs::start_metrics_flushers`) — returns no flushers, so anything drained from the DogStatsD aggregator (custom / enhanced / process) is discarded rather than sent.
3. **Logs flusher** (`logs/flusher.rs::flush`) — short-circuits to an empty request set, guaranteeing no log egress even if something is queued or redriven.

## Testing

- `cargo fmt --all -- --check` — passes.
- `cargo clippy --workspace --all-targets --features default` — passes (no new warnings).
- `cargo test` — passes, including two new tests:
  - `serverless_apm_only_defaults_off` — flag defaults off; other toggles unchanged.
  - `serverless_apm_only_forces_metrics_and_logs_off` — flag overrides explicitly-enabled metrics/logs/OTLP toggles.

### Live validation on AWS Lambda

Built the layer (arm64) and deployed two fully-distinct Node.js 22 functions (Datadog Node layer + this extension), one with the flag off and one on, then invoked each and queried the Datadog API:

| Signal | flag OFF (`serverless-apm-full`) | flag ON (`serverless-apm-standalone`) |
|---|---|---|
| Spans / traces | present | **present** |
| Logs | present | **0** |
| Custom metric | present | **absent (never ingested)** |

The extension's Lambda Telemetry API subscription also drops from `[Platform, Extension, Function]` to `[Platform]` when the flag is on, confirming logs are cut off at the source.

## Risk / rollback

Low. Behavior is gated entirely behind a new flag that defaults to `false`; when unset the code paths are unchanged. Rollback is leaving `DD_SERVERLESS_APM_ONLY` unset.

## Notes for reviewers

- The FIPS-feature clippy variant could not be run locally (macOS is missing the aws-lc FIPS crypto dylib); it exercises the same feature-agnostic config/logs code and runs on the Ubuntu CI runners.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -1300,14 +1300,14 @@ fn start_metrics_flushers(
 ) -> Vec<MetricsFlusher> {
     let mut flushers = Vec::new();
 
-    // APM-only ("traces only") mode: do not create any metrics flushers, so that
-    // no metrics (custom DogStatsD, enhanced, or process) ever reach intake. The
-    // DogStatsD server still runs and the aggregator is still drained on flush, but
-    // the drained data is discarded because there is no flusher to send it. This is
-    // the authoritative guarantee that no infrastructure-monitoring charges are
-    // incurred (see DD_SERVERLESS_APM_ONLY).
-    if config.ext.serverless_apm_only {
-        debug!("DD_SERVERLESS_APM_ONLY is enabled: not starting any metrics flushers");
+    // APM standalone ("traces only") mode: skip wiring up metrics flushers
+    // entirely, so no metrics (custom DogStatsD, enhanced, or process) reach
+    // intake. The DogStatsD server still runs and its aggregator is still drained
+    // on flush, but the drained data is discarded since there is no flusher to
+    // send it. Logs and OTLP signals are suppressed separately (see
+    // `apply_apm_standalone` and the logs flusher guard). See
+    // DD_APM_STANDALONE_ENABLED.
+    if config.ext.apm_standalone_enabled {
         return flushers;
     }
 
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -1300,6 +1300,17 @@ fn start_metrics_flushers(
 ) -> Vec<MetricsFlusher> {
     let mut flushers = Vec::new();
 
+    // APM-only ("traces only") mode: do not create any metrics flushers, so that
+    // no metrics (custom DogStatsD, enhanced, or process) ever reach intake. The
+    // DogStatsD server still runs and the aggregator is still drained on flush, but
+    // the drained data is discarded because there is no flusher to send it. This is
+    // the authoritative guarantee that no infrastructure-monitoring charges are
+    // incurred (see DD_SERVERLESS_APM_ONLY).
+    if config.ext.serverless_apm_only {
+        debug!("DD_SERVERLESS_APM_ONLY is enabled: not starting any metrics flushers");
+        return flushers;
+    }
+
     let metrics_intake_url = if !config.dd_url.is_empty() {
         let dd_dd_url = DdDdUrl::new(config.dd_url.clone()).expect("can't parse DD_DD_URL");
 
--- a/bottlecap/src/config/mod.rs
+++ b/bottlecap/src/config/mod.rs
@@ -28,33 +28,26 @@ pub type Config = datadog_agent_config::Config<LambdaConfig>;
 #[must_use]
 pub fn get_config(config_directory: &Path) -> Config {
     let mut config = get_config_with_extension::<LambdaConfig>(config_directory);
-    apply_serverless_apm_only(&mut config);
+    apply_apm_standalone(&mut config);
     config
 }
 
-/// APM-only ("traces only") mode: suppress every billable metrics and logs
-/// egress path so the customer incurs no infrastructure-monitoring or
+/// APM standalone ("traces only") mode: suppress every billable metrics and
+/// logs egress path so the customer incurs no infrastructure-monitoring or
 /// log-ingestion charges. This intentionally overrides any individually
 /// configured metrics/logs toggles, since the guarantee must hold even if a
 /// user also set, e.g., `DD_ENHANCED_METRICS=true`. Traces and APM trace stats
 /// are unaffected. The custom `DogStatsD` egress is additionally disabled in
 /// the metrics flusher wiring (see `start_metrics_flushers` in `main.rs`), and
 /// log egress is guarded in the logs flusher.
-fn apply_serverless_apm_only(config: &mut Config) {
-    if !config.ext.serverless_apm_only {
+fn apply_apm_standalone(config: &mut Config) {
+    if !config.ext.apm_standalone_enabled {
         return;
     }
 
-    if config.ext.serverless_logs_enabled
-        || config.ext.enhanced_metrics
-        || config.ext.lambda_proc_enhanced_metrics
-        || config.otlp_config_metrics_enabled
-        || config.otlp_config_logs_enabled
-    {
-        tracing::debug!(
-            "DD_SERVERLESS_APM_ONLY is enabled: forcing logs and all metrics off (traces-only mode)"
-        );
-    }
+    tracing::debug!(
+        "DD_APM_STANDALONE_ENABLED is set: forcing logs and all metrics off (traces-only mode)"
+    );
 
     config.ext.serverless_logs_enabled = false;
     config.ext.enhanced_metrics = false;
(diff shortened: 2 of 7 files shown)


bottlecap/src/logs/flusher.rs
```diff
@@ -259,6 +259,14 @@ impl LogsFlusher {
         &self,
         retry_request: Option<reqwest::RequestBuilder>,
     ) -> Vec<reqwest::RequestBuilder> {
+        // APM-only ("traces only") mode: never send logs to intake. Logs are also
+        // dropped upstream (serverless_logs_enabled is forced off, so the processor
+        // never queues them), but this guard guarantees no log egress regardless of
+        // aggregator or redrive state. See DD_SERVERLESS_APM_ONLY.
+        if self.config.ext.serverless_apm_only {
+            return Vec::new();
+        }
```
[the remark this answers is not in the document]
zarirhamza: Good call — done in latest commit. The logs flusher now drains the aggregator and discards the batches in APM-only mode instead of returning early, so the aggregator stays bounded and the logs path mirrors the metrics-side wiring (which also drains + discards). Note that logs are still never actually queued in this mode since `serverless_logs_enabled` is forced off upstream, so this is purely defensive.

bottlecap/src/config/mod.rs
```diff
@@ -260,6 +307,43 @@ mod lambda_config_tests {
...
+
+            assert!(config.ext.serverless_apm_only);
+            assert!(!config.ext.serverless_logs_enabled);
+            assert!(!config.ext.enhanced_metrics);
+            assert!(!config.ext.lambda_proc_enhanced_metrics);
+            assert!(!config.otlp_config_metrics_enabled);
+            assert!(!config.otlp_config_logs_enabled);
+            Ok(())
+        });
+    }
```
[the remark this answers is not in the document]
zarirhamza: Added `serverless_apm_only_from_yaml` in latest commit, which guards the YAML key mapping (`serverless_apm_only:`) via the `load` helper. The override-application path (forcing logs/metrics/OTLP off) is already covered by `serverless_apm_only_forces_metrics_and_logs_off` through `get_config`.

bottlecap/src/bin/bottlecap/main.rs
```diff
@@ -1300,6 +1300,17 @@ fn start_metrics_flushers(
...
     let mut flushers = Vec::new();
 
+    // APM-only ("traces only") mode: do not create any metrics flushers, so that
+    // no metrics (custom DogStatsD, enhanced, or process) ever reach intake. The
+    // DogStatsD server still runs and the aggregator is still drained on flush, but
+    // the drained data is discarded because there is no flusher to send it. This is
+    // the authoritative guarantee that no infrastructure-monitoring charges are
+    // incurred (see DD_SERVERLESS_APM_ONLY).
+    if config.ext.serverless_apm_only {
+        debug!("DD_SERVERLESS_APM_ONLY is enabled: not starting any metrics flushers");
```
duncanista: misleading, you're disabling other things

bottlecap/src/config/mod.rs
```diff
@@ -25,10 +25,42 @@ use serde::Deserialize;
...
+
+    if config.ext.serverless_logs_enabled
+        || config.ext.enhanced_metrics
+        || config.ext.lambda_proc_enhanced_metrics
+        || config.otlp_config_metrics_enabled
+        || config.otlp_config_logs_enabled
+    {
+        tracing::debug!(
+            "DD_SERVERLESS_APM_ONLY is enabled: forcing logs and all metrics off (traces-only mode)"
```
duncanista: double logging?

duncanista (commented):
some integration testing would be good here, overall, not sure about the naming of the env var, is this going to be applied to every serverless product? does this type of env var exist outside of serverless?

bottlecap/src/bin/bottlecap/main.rs
zarirhamza: Reworded so this comment only describes the metrics-egress path, and removed the "authoritative guarantee" claim. Logs/OTLP suppression happens separately in `apply_apm_standalone` (config override) and the logs flusher guard — noted that in the comment. Done in a1b1b897.

bottlecap/src/config/mod.rs
zarirhamza: Good catch — consolidated to a single debug line in `apply_apm_standalone` and dropped the redundant one in `start_metrics_flushers`. Also removed the `if any-toggle-was-set` wrapper so it just logs once when the mode is enabled. Done in a1b1b897.

zarirhamza:
@duncanista thanks for the review — addressed your points in a1b1b897:

**Naming.** Renamed `DD_SERVERLESS_APM_ONLY` → **`DD_APM_STANDALONE_ENABLED`** (field `apm_standalone_enabled`, fn `apply_apm_standalone`) to line up with the "APM Standalone" product terminology (the brief is "APM Standalone on Lambda"). The closest prior art is the tracers' `DD_APM_TRACING_ENABLED` (Standalone ASM/DSM), which is the inverse direction; `DD_APM_STANDALONE_ENABLED` reads as an explicit opt-in here.

**Applied to every serverless product?** That's the intent — the extension is serverless-only, so I dropped the `SERVERLESS_` prefix but kept it product-agnostic so Cloud Run / Azure Container Apps can adopt the same flag. Only the Lambda egress paths (metrics flushers + logs flusher) are wired today; other products can hook into the same `apm_standalone_enabled` override as they land.

**Outside serverless?** Not currently — this flag lives in the Lambda `ConfigExtension`, so it's scoped to the extension. If we ever want a core-agent equivalent we can revisit aligning names.

**Integration testing.** Added an `apm-standalone` suite (`integration-tests/tests/apm-standalone.test.ts` + stack + `test-suites.yaml`). It deploys a baseline function and a standalone (`DD_APM_STANDALONE_ENABLED=true`) function that both emit a trace, a custom DogStatsD metric, and logs, then asserts:
- traces (incl. the `aws.lambda` root span) survive for **both**
- `aws.lambda.enhanced.invocations` and the custom metric are emitted for baseline, **suppressed** for standalone
- logs are forwarded for baseline, **not** for standalone — even though `DD_SERVERLESS_LOGS_ENABLED=true` is inherited from the default env, which the mode overrides.

bottlecap/src/logs/flusher.rs
```diff
@@ -259,6 +259,19 @@ impl LogsFlusher {
         &self,
         retry_request: Option<reqwest::RequestBuilder>,
     ) -> Vec<reqwest::RequestBuilder> {
+        // APM standalone ("traces only") mode: never send logs to intake. Logs are
+        // also dropped upstream (serverless_logs_enabled is forced off, so the
+        // processor never queues them), but this guard guarantees no log egress
+        // regardless of aggregator or redrive state. See DD_APM_STANDALONE_ENABLED.
+        if self.config.ext.apm_standalone_enabled {
```
duncanista, line 266: isn't there like an "enabled" section where we'd just add this?

duncanista (approved):
LGTM – i'd cut down the amount of comments, they're too long, can be done with fewer words, overall, left another comment, would test it on actual resources/integration tests before merging!

PR #1301, opened 2026-07-11
[SVLS-8800] fix: hold back cold-start invocation metric until durable tag is known

lym953:
## Problem
For durable functions, the `aws.lambda.enhanced.invocations` metric for the first invocation (after cold start) doesn't have the tag `durable_function:true` as expected.

This is because the metric is emitted in `on_invoke_event()`, which, from what we observed, always happens before the extension receives `platform.initStart` event, which carries the information of whether the function is a durable function. As a result, the `durable_function:true` tag is only set on the metric for subsequent invocations but not for the first one.

## Fix
Delay the flushing of `aws.lambda.enhanced.invocations` metric for the first invocation, if the extension doesn't yet know whether the function is a durable function.

Use an enum to track the state of the first invocation:
```
enum FirstInvocationMetricStatus {
    Pending,
    InitStartSeen,
    HeldBack(i64),
    Resolved,
}
```

This covers both cases about the order of `on_invoke_event()` event and `platform.initStart` event:
1. Common case: `on_invoke_event()` comes first, then `platform.initStart`:
    - To start with: `Pending`
    - `on_invoke_event()` comes: hold metric and move to `HeldBack`
    - `platform.initStart` comes: flush metric and move to `Resolved`
2. Rare case: `platform.initStart` comes before `on_invoke_event()`
    - To start with: `Pending`
    - `platform.initStart` comes: `InitStartSeen`
    - `on_invoke_event()` comes: flush metric immediately and move to `HeldBack`

Also, flush the metric at shutdown, in case `platform.initStart` is never received for whatever reason.

### Manual testing
For a durable function, all data points for the `aws.lambda.enhanced.invocations` metric have the tag `durable_function:true` set.
image

### Automated testing
Passed the added integration test.

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor.rs
+++ b/bottlecap/src/lifecycle/invocation/processor.rs
@@ -63,7 +63,6 @@ pub const DATADOG_INVOCATION_ERROR_KEY: &str = "x-datadog-invocation-error";
 const DATADOG_SPAN_ID_KEY: &str = "x-datadog-span-id";
 const TAG_SAMPLING_PRIORITY: &str = "_sampling_priority_v1";
 
-#[allow(clippy::struct_excessive_bools)]
 pub struct Processor {
     /// Buffer containing context of the previous 5 invocations
     context_buffer: ContextBuffer,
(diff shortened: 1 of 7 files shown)


jchrostek-dd:
Do you think we should add a comment to agents.md about the ordering of lambda events so agents are aware and proactively catch issues before they are bugs?

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -111,6 +111,12 @@ pub struct Processor {
     /// on `platform.report`. This flag ensures whichever event arrives first wins and the other is skipped,
     /// preventing double counting.
     init_duration_metric_emitted: bool,
+    /// Timestamp of the cold-start invocation metric in On-Demand mode, held back until
+    /// `platform.initStart` sets the durable tag so it's not missed on invocation #1.
+    ///
+    /// This assumes `platform.initStart` always arrives after the Invoke event, which is what
+    /// we've observed in production.
```
jchrostek-dd: Is this actually a guarantee? That platform.initStart must appear after the first invoke event? Or is is just likely to appear after the first invoke event?

bottlecap/src/lifecycle/invocation/processor.rs
```diff
@@ -154,12 +160,17 @@ impl Processor {
             durable_context_tx,
             restore_time: None,
             init_duration_metric_emitted: false,
+            pending_first_invocation_metric_timestamp: None,
         }
```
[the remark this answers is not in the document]
lym953: Not sure what this means. There are no `init_start_seen` / `first_invocation_seen` flags.

bottlecap/src/lifecycle/invocation/processor.rs
lym953: It's not a guarantee. This is just the order I consistently observed from my tests, and I didn't want to over-engineer.
Since you raised this, let me make code robust by handling both cases.

lym953: Addressed

lym953:
Oh no, I see duplicate logs in Datadog. Let me fix it.

lym953:
> Oh no, I see duplicate logs in Datadog. Let me fix it.

That's unrelated to this PR. Marking the PR as Ready for Review again.

jchrostek-dd (approved):

Issue #144, opened 2023-06-23
Misleading information when DD_API_KEY_SECRET_ARN points to non-existing secret


t0mmili:
Let's say I made a mistake when typing ARN of SecretsManager secret into DD_API_KEY_SECRET_ARN.

The chain of information in DEBUG log is a little bit misleading:
1. Retrieving - ok
2. Found - that's not possible
3. Secrets Manager read error: AccessDeniedException - I assume this is from AWS API...?
4. No API key configured - actually it is, but not found :)

```
2023-06-23T10:42:03.581+02:00	2023-06-23 08:42:03 UTC | DD_EXTENSION | DEBUG | Retrieving DD_API_KEY_SECRET_ARN=arn:aws:secretsmanager:us-east-2:012345678901:secret:secret-datacat from secrets manager
2023-06-23T10:42:03.581+02:00	2023-06-23 08:42:03 UTC | DD_EXTENSION | DEBUG | Found arn:aws:secretsmanager:us-east-2:012345678901:secret:secret-datacat value, trying to use it.
2023-06-23T10:42:03.701+02:00	2023-06-23 08:42:03 UTC | DD_EXTENSION | DEBUG | Couldn't read API key from Secrets Manager: Secrets Manager read error: AccessDeniedException: User: arn:aws:sts::012345678901:assumed-role/role-my-lambda-use2/my-lambda-use2 is not authorized to perform: secretsmanager:GetSecretValue on resource: arn:aws:secretsmanager:us-east-2:012345678901:secret:secret-datacat because no identity-based policy allows the secretsmanager:GetSecretValue action
2023-06-23T10:42:03.701+02:00	status code: 400, request id: b507ebca-3b2d-4f6d-bf68-cb5fe57e42ad
2023-06-23T10:42:03.701+02:00	2023-06-23 08:42:03 UTC | DD_EXTENSION | ERROR | No API key configured
```

astuyve:
Hey @t0mmili! Thanks for reaching out.

To prevent information leaking (enumeration of secrets, etc), AWS doesn't tell us if the key doesn't exist or if we don't have permissions to read it. Github does the same thing for repositories but in reverse - if you're logged out and try to access a private repo, they throw a 404 page (even though in reality the repo exists but you just aren't logged in).
This is relatively standard practice.

[...]

PR #1303, opened 2026-07-13
[SLES-2907] chore(deps): align libdatadog rev, dedupe libdd-* crates

litianningdatadog:
## Overview

Follows up on the `dd-trace-rs` v0.5.0 bump ([#1302](https://github.com/DataDog/datadog-lambda-extension/pull/1302)): bumps the shared `libdatadog` git rev (`libdd-capabilities`, `libdd-common`, `libdd-trace-protobuf`, `libdd-trace-utils`, `libdd-trace-normalization`, `libdd-trace-obfuscation`, `libdd-trace-stats`) from `a820699` to `85ce322a`, matching the exact versions `datadog-opentelemetry` v0.5.0 pulls from crates.io. Adds a `[patch.crates-io]` block redirecting those (and the other transitively-pulled `libdd-*` crates) to the same rev, collapsing what would otherwise be two compiled copies of each.

This addresses the duplication [Copilot flagged on serverless-components#144](https://github.com/DataDog/serverless-components/pull/144#discussion_r3572767510) where broke type identity for `ReplaceRule`/`Endpoint`/`Span`/etc. across the git-vs-registry boundary throughout the trace pipeline) or leaving the duplication as-is.

### Why not just add this to #1302?

#1302 skipped this on purpose. The duplication only cost +256 bytes on the release binary (LTO strips the unreachable copy), while actually fixing it means an untested major-version bump across six libdatadog crates, including `libdd-trace-obfuscation` — the code that redacts sensitive data from spans. Not something to risk on the same PR as an urgent customer fix (SLES-2907, B3 propagation errors). This PR does that bump on its own, with its own tests.

**Depends on [DataDog/serverless-components#145](https://github.com/DataDog/serverless-components/pull/145)** (which itself builds on [#144](https://github.com/DataDog/serverless-components/pull/144)) — the `dogstatsd`/`datadog-fips`/`datadog-agent-config` rev pin here must stay in sync with whatever commit that PR lands at. Opened as draft for that reason.

JIRA: [SLES-2907](https://datadoghq.atlassian.net/browse/SLES-2907)

### Code changes at the new rev

Most of the API changes in `SpanConcentrator` are just things the compiler forces on you, with no behavior change:

- `new()` gained an obfuscation-config param. We pass `None` — bottlecap has never done client-side stats obfuscation.
- `flush()` now returns `FlushResult { obfuscated_buckets, unobfuscated_buckets }` instead of a flat `Vec`. We concatenate both — obfuscation is off, so everything ends up in `unobfuscated_buckets` anyway.
- `pb::TracerPayload` gained `container_debug`, `pb::ClientGroupedStats` gained `additional_metric_tags`. Added the missing fields to 3 struct literals so things compile; both are left at their defaults.

`SpanConcentrator::new` also gained a cardinality-override param — the new rev defaults to capping trace-stats aggregation at 7,000 distinct keys per 10-second bucket, added in [DataDog/libdatadog#2158](https://github.com/DataDog/libdatadog/pull/2158) to bound `SpanConcentrator`'s own memory use ([#2158](https://github.com/DataDog/libdatadog/pull/2158) has the math: ~45.6 MB typical, ~700 MB worst case across 3 buckets with the limit on). This is an APM/trace-stats decision, unrelated to Datadog's custom-metrics cardinality limits.

The scope of this PR is dependency consolidation, not adopting a new safety feature, so we pass `Some(usize::MAX)` to keep this rev bump behavior-neutral: no Lambda workload will ever produce `usize::MAX` distinct keys in a 10-second window, so stats collapsing never triggers, matching pre-bump behavior exactly. Passing `None` instead would opt into new behavior (collapsing above 7,000 keys into a generic `tracer_blocked_value` bucket for workloads that never hit that before) — worth considering on its own merits given Lambda's memory constraints, but that's a deliberate product/safety decision that deserves its own PR and review, not something to fold into a dependency bump. `test_no_cardinality_limit_applied` covers the unbounded behavior we're keeping.

## Testing

- `cargo build --all-targets` — clean
- `cargo test --all-targets` — 548/548 pass across all test binaries
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean
- `cargo tree --duplicates | grep ^libdd-` — empty (zero `libdd-*` duplicates)
- `dd-rust-license-tool check` — passes
- Verified against the real `serverless-components#145` commit (not a local `file://` override used during earlier iteration)

[SLES-2907]: https://datadoghq.atlassian.net/browse/SLES-2907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/traces/stats_concentrator_service.rs
+++ b/bottlecap/src/traces/stats_concentrator_service.rs
@@ -222,9 +222,11 @@ impl StatsConcentratorService {
     ) {
         let flush_result = self.concentrator.flush(SystemTime::now(), force_flush);
         // Obfuscation is disabled (see `SpanConcentrator::new` above), so every bucket ends up
-        // in `unobfuscated_buckets`; combine both to stay correct if that ever changes.
-        let mut stats_buckets = flush_result.obfuscated_buckets;
-        stats_buckets.extend(flush_result.unobfuscated_buckets);
+        // in `unobfuscated_buckets`; combine both to stay correct if that ever changes. Start
+        // from `unobfuscated_buckets` since it's normally the only non-empty one, avoiding a
+        // reallocation to grow the (usually empty) `obfuscated_buckets` vec.
+        let mut stats_buckets = flush_result.unobfuscated_buckets;
+        stats_buckets.extend(flush_result.obfuscated_buckets);
         let stats = if stats_buckets.is_empty() {
             None
         } else {
--- a/bottlecap/src/traces/stats_concentrator_service.rs
+++ b/bottlecap/src/traces/stats_concentrator_service.rs
@@ -186,8 +186,8 @@ impl StatsConcentratorService {
                 .iter()
                 .map(ToString::to_string)
                 .collect(),
-            // Use libdd-trace-stats' default cardinality limit.
-            None,
+            // Disable the cardinality limit to match pre-existing (unbounded) behavior.
+            Some(usize::MAX),
             // Bottlecap does not perform client-side stats obfuscation.
             None,
         );
@@ -280,6 +280,16 @@ mod tests {
     /// The span is non-root (`parent_id=1`) and not measured, so it will only be
     /// eligible for stats if `span_kinds_stats_computed` includes its `span.kind`.
     fn create_span_kind_span(span_kind: &str, meta: Vec<(&str, &str)>) -> pb::Span {
+        create_span_kind_span_with_resource(span_kind, "test-resource", meta)
+    }
+
+    /// Same as `create_span_kind_span`, but with a caller-provided resource name so tests can
+    /// generate many distinct aggregation keys.
+    fn create_span_kind_span_with_resource(
+        span_kind: &str,
+        resource: &str,
+        meta: Vec<(&str, &str)>,
+    ) -> pb::Span {
         let now_ns = i64::try_from(
             SystemTime::now()
                 .duration_since(SystemTime::UNIX_EPOCH)
@@ -295,7 +305,7 @@ mod tests {
         pb::Span {
             service: "test-service".to_string(),
             name: "test-op".to_string(),
-            resource: "test-resource".to_string(),
+            resource: resource.to_string(),
             trace_id: 1,
             span_id: 2,
             parent_id: 1, // non-root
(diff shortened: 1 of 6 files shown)


purple4reina (changes requested):
A few things I'd like to understand better before I can approve.

First, on the PR description: I'd love a full rewrite. Right now it reads like deep implementation notes — accurate, but pitched at someone already inside the code, not at a reviewer trying to understand intent and risk. For each change, could you write a couple of plain-language sentences on what it does, why it's needed, and what the impact/risk is? That's the level I need to actually evaluate this, and it'll help whoever reads this PR's history down the line too.

What I understand this PR to be doing: bumping dependencies and then aligning dep versions between here and datadog-opentelemetry, so these packages are only compiled into the binary once. That part makes sense to me.

What I don't understand, and would like explained:

The other changes in the PR don't seem related to the dependency change — or if they are, that hasn't been made clear to me yet.

The first of those changes disables the cardinality limit on metrics. This seems like a huge and impactful change. Why are we doing this? What impact does it have? How do we know this is safe? I know Datadog just announced some level of unlimited cardinality on metrics for our users — is this related to that? Has it been approved by the metrics team? I'm not comfortable approving this specific change without answers to these questions. As is, the release notes won't say anything about turning off cardinality.

The second change has something to do with obfuscation. I don't understand what this is doing or why. Is it related to the dependency change? If not, it needs a clearer explanation.

litianningdatadog:
> A few things I'd like to understand better before I can approve.
> 
> First, on the PR description: I'd love a full rewrite. Right now it reads like deep implementation notes — accurate, but pitched at someone already inside the code, not at a reviewer trying to understand intent and risk. For each change, could you write a couple of plain-language sentences on what it does, why it's needed, and what the impact/risk is? That's the level I need to actually evaluate this, and it'll help whoever reads this PR's history down the line too.
> 
> What I understand this PR to be doing: bumping dependencies and then aligning dep versions between here and datadog-opentelemetry, so these packages are only compiled into the binary once. That part makes sense to me.
> 
> What I don't understand, and would like explained:
> 
> The other changes in the PR don't seem related to the dependency change — or if they are, that hasn't been made clear to me yet.
> 
> The first of those changes disables the cardinality limit on metrics. This seems like a huge and impactful change. Why are we doing this? What impact does it have? How do we know this is safe? I know Datadog just announced some level of unlimited cardinality on metrics for our users — is this related to that? Has it been approved by the metrics team? I'm not comfortable approving this specific change without answers to these questions. As is, the release notes won't say anything about turning off cardinality.
> 
> The second change has something to do with obfuscation. I don't understand what this is doing or why. Is it related to the dependency change? If not, it needs a clearer explanation.

@purple4reina The PR message has been revised to address your concerns. Aside from the dependency and required API updates, the remaining changes were triggered by Codex/Copilot code review.

bottlecap/src/traces/stats_concentrator_service.rs
```diff
@@ -186,6 +186,10 @@ impl StatsConcentratorService {
                 .iter()
                 .map(ToString::to_string)
                 .collect(),
+            // Disable the cardinality limit to match pre-existing (unbounded) behavior.
+            Some(usize::MAX),
+            // Bottlecap does not perform client-side stats obfuscation.
```
purple4reina:
```suggestion
            // Bottlecap does not perform agent-side stats obfuscation.
```

purple4reina (approved):
Thanks for the explanation.

bottlecap/src/traces/stats_concentrator_service.rs
litianningdatadog: Acked

PR #1312, opened 2026-07-27
feat(logging): surface flush data-type and duration at INFO level

purple4reina:
⚠️ DO NOT MERGE ⚠️ 

Promote existing DEBUG logs (traces, stats, proxy) to INFO and add minimal new INFO logs (logs, metrics) so customer can see which telemetry type is flushing and how long each flush took without needing DD_LOG_LEVEL=debug. All lines share a FLUSH_TIMING tag so they're easy to search for in Datadog.

https://datadoghq.atlassian.net/browse/SLES-2903

These changes are built on v97 and will be shipped to a single customer as a custom extension build.

### DIFF
--- a/bottlecap/src/flushing/service.rs
+++ b/bottlecap/src/flushing/service.rs
@@ -1,8 +1,9 @@
 //! `FlushingService` for coordinating flush operations across multiple flusher types.
 
 use std::sync::Arc;
+use std::time::Instant;
 
-use tracing::{debug, error};
+use tracing::{debug, error, info};
 
 use dogstatsd::{
     aggregator::AggregatorHandle as MetricsAggregatorHandle, flusher::Flusher as MetricsFlusher,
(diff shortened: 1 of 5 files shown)


bottlecap/src/logs/flusher.rs
```diff
@@ -299,6 +300,15 @@ impl LogsFlusher {
             for failed in results {
                 failed_requests.extend(failed);
             }
+
```
litianningdatadog, line 311: This may cause self-sustaining log. A safer approach would be emitting metric if possible.

bottlecap/src/flushing/service.rs
```diff
@@ -317,10 +325,19 @@ impl FlushingService {
...
-                    flush_response.distributions.clone(),
-                )
+                let series = flush_response.series.clone();
+                let sketches = flush_response.distributions.clone();
+                let series_count = series.len();
+                let sketches_count = sketches.len();
+                async move {
+                    let start = Instant::now();
+                    let result = f.flush_metrics(series, sketches).await;
+                    info!(
```
litianningdatadog, line 341: How about logging idx to identify flusher?

litianningdatadog (commented):
Added some comments

bottlecap/src/logs/flusher.rs
purple4reina: Good call out. I don't want to emit a metric because that's making a much larger change when I need to keep things as close to the original v97 as possible.

I'll test this code out in a live lambda function and see if we get the infinite loop and if it causes any sort of issues.

litianningdatadog: ok, as long as it is temporary.

purple4reina: As stated in the PR description, there is no intention of merging this PR. It is just to provide a custom build to a single customer.

purple4reina:
Custom build has been provided to the customer

PR #1314, opened 2026-07-29
[SLES-2961] fix(traces): extract Binary _datadog attribute from SQS records

litianningdatadog:
https://datadoghq.atlassian.net/browse/SLES-2961

## Overview

SNS→SQS→Lambda trace propagation breaks when the `_datadog` context attribute reaches SQS as a Binary message attribute (e.g. via SNS raw message delivery). The consumer Lambda starts a new trace instead of continuing the publisher's.

`SqsRecord::get_carrier()` only checked `string_value` on the `_datadog` attribute. When the attribute was Binary-typed, that check silently failed and the code fell through to parsing the SQS body as an SNS envelope — which doesn't exist under raw delivery — so it returned an empty carrier. The SNS-native path (`SnsRecord::get_carrier()`) already branches on the attribute's type and base64-decodes Binary values; the SQS-native path never got the same treatment.

Fix: branch on `data_type` in `SqsRecord::get_carrier()`, base64-decoding the Binary case before parsing the carrier JSON, matching `sns_event.rs`.

## Testing

- Added `sqs_event_binary.json` fixture and `test_get_carrier_binary`, reproducing the bug (fails without the fix, passes with it).
- `cargo test --lib` — 145 trigger tests pass.
- `cargo clippy --lib -- -D warnings` — clean.
- Integration tests using ([publisher](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/tianning-li-20260802-2118-publisher) and [consumer](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/tianning-li-20260802-2119-consumer?tab=code))
[Before the fix](https://ddserverless.datadoghq.com/apm/trace/8657375107204609059?graphType=waterfall&shouldShowLegend=true&spanID=12291667060860692126&traceQuery=), the spans were disconnected
image

[After the fix](https://ddserverless.datadoghq.com/apm/trace/10216024880978878436?graphType=waterfall&shouldShowLegend=true&spanID=16966308994805779089&timeHint=1785722075413&trace=AwAAAZ_FVHkVYFTqSwAAABhBWl9GVko4VUFBRDIwZTI0N2tna3pjYU4AAAAkZjE5ZmM1NTQtOTM0YS00OGJjLThmMmEtM2ZhYzA2ZjBjOWE5AAAACA&traceQuery=), the spans are no longer disconnected
image


bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
```diff
@@ -177,10 +178,24 @@ impl Trigger for SqsRecord {
...
+                    if let Some(binary_value) = &ma.binary_value
+                        && let Ok(carrier) = base64_to_string(binary_value)
+                    {
+                        return serde_json::from_str(&carrier).unwrap_or_default();
+                    }
+                }
+                _ => {
+                    debug!("Unsupported dataType in SQS message attribute");
+                }
+            }
```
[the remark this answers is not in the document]
litianningdatadog: Fixed in cd97dde5 — matching `data_type` via `starts_with("String")`/`starts_with("Binary")` instead of exact equality, so custom-labeled types (`String.foo`, `Binary.foo`) fall through correctly. Added `test_get_carrier_binary_custom_label` to cover it.

bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
```diff
@@ -177,10 +178,24 @@ impl Trigger for SqsRecord {
     fn get_carrier(&self) -> HashMap<String, String> {
         let carrier = HashMap::new();
 
-        if let Some(ma) = self.message_attributes.get(DATADOG_CARRIER_KEY)
-            && let Some(string_value) = &ma.string_value
-        {
-            return serde_json::from_str(string_value).unwrap_or_default();
+        if let Some(ma) = self.message_attributes.get(DATADOG_CARRIER_KEY) {
+            match ma.data_type.as_str() {
```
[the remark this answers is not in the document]
litianningdatadog: Same finding as the Copilot comment above — fixed in cd97dde5 by matching `data_type` by prefix instead of exact string, preserving support for custom-labeled types.

bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
lym953, line 417: Should we rename this test and this file by adding "string" to the name?

bottlecap/tests/payloads/sqs_event_binary_custom_label.json
```diff
@@ -0,0 +1,27 @@
+{
```
lym953, line 1: nit: Should we add "binary" to file name?

lym953 (approved):

bottlecap/src/lifecycle/invocation/triggers/sqs_event.rs
litianningdatadog: Renaming it will trigger much bigger scope of changes as it is referenced in multiple places.

bottlecap/tests/payloads/sqs_event_binary_custom_label.json
litianningdatadog: Acked

PR #1333, opened 2026-08-19
Advertise bare /evp_proxy/v2/ prefix in /info endpoints list

purple4reina:
## Overview

See also https://chonk.us1.staging.dog/v2/get/chonk-uploads/rey.abolofia/5b3d2091-88e4-438b-853e-baccff15a8a3#rootcause1

dd-trace-java's `DDAgentFeaturesDiscovery` and dd-trace-js's `setAgentStrategy` decide whether the local agent supports EVP proxying by checking `/info`'s `endpoints` array for the exact bare string `/evp_proxy/v2/`. This extension only ever advertised fully-qualified concrete routes (`/evp_proxy/v2/api/v2/llmobs`, etc.) — never the bare prefix the real Datadog Agent advertises (`pkg/trace/api/endpoints.go`). Both tracers read that as "proxy not supported" and silently fall back to agentless LLM Obs submission, which drops spans whenever Lambda freezes the sandbox before that fallback path's independent background flush timer fires (it isn't part of the extension's normal flush-before-freeze gate).

This is the confirmed root cause of frequent `ml_obs.trace` metric failures for java and node Lambda runtimes in serverless-e2e-tests. (dd-trace-py isn't affected — its equivalent check does a substring match, which happens to still find `/evp_proxy/v2` inside the concrete routes.)

Fix: advertise the bare `/evp_proxy/v2/` prefix in `/info` alongside the existing concrete routes, matching the real Agent's behavior. No routing changes — the concrete `/evp_proxy/v2/...` handlers were already correct and reachable.

## Testing

- Added `info_advertises_the_bare_evp_proxy_v2_prefix`, which calls `TraceAgent::info()`
  directly and asserts the bare prefix is present in the JSON `endpoints` array.
- `cargo test --lib traces::trace_agent::tests` — 5 passed.
- `cargo clippy --lib -- -D warnings` and `cargo fmt --check` — clean.

### DIFF
--- a/bottlecap/src/bin/bottlecap/main.rs
+++ b/bottlecap/src/bin/bottlecap/main.rs
@@ -1193,8 +1193,16 @@ fn start_trace_agent(
     let trace_agent_channel = trace_agent.get_sender_copy();
     let shutdown_token = trace_agent.shutdown_token();
 
+    // Bind the trace agent's socket synchronously, before spawning its async task and before
+    // this function returns to the caller (which proceeds to call the Lambda Extensions API's
+    // `/next`, triggering the runtime to invoke the handler). This closes the race where a
+    // tracer's very early `/info` probe on cold start could hit a not-yet-listening socket and
+    // permanently cache a "no EVP proxy support" decision. See TraceAgent::bind() for details.
+    let trace_agent_listener =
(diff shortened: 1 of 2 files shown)


lym953:
Thanks for the fix!

lym953 (approved):

purple4reina:
This was not the fix to the issue.

PR #1344, opened 2026-08-27
[APMSVLS-501] refactor(bottlecap): preparatory work for a new "test-mode" binary

lucaspimentel:
TL;DR: refactoring here first to prepare for #1216 and keep that PR slightly smaller.

> NOTE: The original version of this PR was #1201. I've made several changes after feedback, so starting fresh to reduce baggage.

Part of a PR stack:
1. #1344 **👈🏽 this PR**
2. #1216

## Overview

Preparatory refactors for the upcoming `bottlecap-testmode` binary ([APMSVLS-501](https://datadoghq.atlassian.net/browse/APMSVLS-501)), which reuses `TraceAgent` / `handle_traces` but has no Lambda lifecycle to drive. Two logical changes:

The `bottlecap::startup` extraction (`build_trace_agent` and the public `startup` module) lands in #1216 alongside the `bottlecap-test-mode` binary that consumes it, where it has a caller; this PR keeps its scope to the two seams below and leaves `main.rs` and `lib.rs` untouched.

### 1. `InvocationProcessorHandle::noop()`

Constructor backed by a background task that acknowledges every `ProcessorCommand` with a sensible default so callers never block on their response oneshots:
- Request-response commands (`GetReparentingInfo`, `UpdateReparenting`, `SetColdStartSpanTraceId`, `PlatformRuntimeDone`, `PlatformReport`) reply with empty/default values.
- Fire-and-forget commands are dropped silently.

The `match` is exhaustive: adding a new `ProcessorCommand` variant will cause a compile error here, forcing test-mode behavior to be decided explicitly. A response-carrying variant placed in the fire-and-forget arm would silently drop its sender, causing the caller to receive `ProcessorError::ChannelReceive` instead of the intended default.

The noop channel capacity matches the real `InvocationProcessorService` to avoid backpressure surprises.

### 2. `RouterExtension` seam on `TraceAgent`

Adds a generic extension point that lets a caller merge additional axum routes into the trace-agent's HTTP router without `TraceAgent` knowing what those routes do. When no caller attaches an extension (the Lambda binary case), the HTTP surface on port 8126 is unchanged.

- New `pub trait RouterExtension: Send + Sync { fn extend(&self, router: Router) -> Result<Router, Box<dyn Error>>; }` defined in `traces::trace_agent`. Returning `Err` aborts agent startup and surfaces the error in the existing log path, preventing silent failures from a panicking or misconfigured extension.
- `TraceAgent` holds `router_extension: Option<Arc<dyn RouterExtension>>` with a consuming builder method `with_router_extension(self, ext: Arc<dyn RouterExtension>) -> Self`.
- `make_router` calls `extension.extend(router)?` when the field is set, before applying the outer fallback and body-limit layers.
- A `#[cfg(test)]`-only `SpyExtension` validates the seam end-to-end: the trait shape compiles, state is carried via `Arc`, and merged routes are reachable through the composed `Router` returned by `make_router`.

Any concrete flush/drain/diagnostic endpoint lives behind an impl of this trait in the consumer crate, not in `trace_agent.rs`.

## Testing

- [x] `cargo check --bin bottlecap`
- [x] `cargo check --lib`
- [x] `cargo test --lib`
- [x] `cargo clippy --workspace --all-targets --features default -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] `cargo nextest run --workspace` (532/532 passed)
- New unit tests on `InvocationProcessorHandle::noop()`: `noop_request_response_methods_return_defaults`, `noop_fire_and_forget_commands_do_not_panic`, `noop_platform_runtime_done_and_report_respond_without_blocking`, `noop_request_response_variants_complete_within_timeout`
- New unit tests on the `RouterExtension` seam: `with_router_extension_adds_reachable_route_to_make_router`, `make_router_returns_404_for_extension_route_when_none_attached`

[APMSVLS-501]: https://datadoghq.atlassian.net/browse/APMSVLS-501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bottlecap/src/lifecycle/invocation/processor_service.rs
+++ b/bottlecap/src/lifecycle/invocation/processor_service.rs
@@ -141,6 +141,71 @@ pub struct InvocationProcessorHandle {
 }
 
 impl InvocationProcessorHandle {
+    /// Returns a handle backed by a background task that acknowledges every
+    /// command with a sensible default. Use this for callers that need an
+    /// `InvocationProcessorHandle` (for example, to reuse `TraceAgent` or
+    /// `handle_traces`) but have no Lambda lifecycle state to drive.
+    ///
+    /// The match below is exhaustive: adding a new `ProcessorCommand` variant
+    /// will fail to compile here, so the noop behavior for it must be
+    /// decided explicitly. A response-carrying variant placed in the
+    /// fire-and-forget arm would silently drop its sender, causing the
+    /// caller to receive `ProcessorError::ChannelReceive` instead of the
+    /// intended default. Note that the compiler only catches *new* variants:
+    /// adding a `response` field to an existing fire-and-forget variant still
+    /// matches its `{ .. }` pattern, so that case must be caught by review.
+    ///
+    /// Compiled only under unit tests and when the `test-mode` feature is
+    /// enabled, so this constructor does not exist in production builds.
+    ///
+    /// Must be called from within an active Tokio runtime: the constructor
+    /// uses `tokio::spawn` to start the draining task, which panics if no
+    /// runtime is registered for the current thread.
+    #[cfg(any(test, feature = "test-mode"))]
+    #[must_use]
+    pub fn noop() -> Self {
+        let (sender, mut receiver) = mpsc::channel::<ProcessorCommand>(1000);
+        tokio::spawn(async move {
+            while let Some(command) = receiver.recv().await {
+                match command {
+                    // Request-response commands: reply with a default so the
+                    // caller doesn't block on the oneshot forever.
+                    ProcessorCommand::GetReparentingInfo { response } => {
+                        let _ = response.send(Ok(std::collections::VecDeque::new()));
+                    }
+                    ProcessorCommand::UpdateReparenting { response, .. } => {
+                        let _ = response.send(Ok(Vec::new()));
+                    }
+                    ProcessorCommand::SetColdStartSpanTraceId { response, .. } => {
+                        let _ = response.send(Ok(None));
+                    }
+                    ProcessorCommand::PlatformRuntimeDone { response, .. }
+                    | ProcessorCommand::PlatformReport { response, .. } => {
+                        let _ = response.send(());
+                    }
+                    // Fire-and-forget commands: drop silently.
+                    ProcessorCommand::InvokeEvent { .. }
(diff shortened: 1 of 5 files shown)


bottlecap/src/lifecycle/invocation/processor_service.rs
```diff
@@ -137,6 +137,70 @@ pub struct InvocationProcessorHandle {
...
+                    }
+                    ProcessorCommand::SetColdStartSpanTraceId { response, .. } => {
+                        let _ = response.send(Ok(None));
+                    }
+                    ProcessorCommand::PlatformRuntimeDone { response, .. }
+                    | ProcessorCommand::PlatformReport { response, .. } => {
+                        let _ = response.send(());
+                    }
+                    // Fire-and-forget commands: drop silently.
+                    ProcessorCommand::InvokeEvent { .. }
```
litianningdatadog, line 187: Do we want to consider to add `RecordDsmConsumeFromPayload` as well?

litianningdatadog (changes requested):

bottlecap/src/lifecycle/invocation/processor_service.rs
lucaspimentel: Good catch, thanks! Looks like `RecordDsmConsumeFromPayload` was added to `main` in #1265 (DSM context extraction), which was merged ~45 minutes after the last commit on _this_ branch (at the time of this writing), so it's not visible on this branch yet. I'll update this branch by rebasing on `main` now and fix the merge conflicts.

lucaspimentel: To confirm, rebasing on `main` caused a failure on `cargo nextest run --workspace` until I added `RecordDsmConsumeFromPayload`, so this is working as expected 😅 

```
error[E0004]: non-exhaustive patterns: `ProcessorCommand::RecordDsmConsumeFromPayload { .. }` not covered
    --> src/lifecycle/invocation/processor_service.rs:170:23
```

lucaspimentel:
@codex review

litianningdatadog (approved):

PR #1357, opened 2026-09-03
[SLES-2997] fix: stitch split telemetry payloads instead of dropping the batch

shreyamalpani:
## Overview

When a record exceeds the Telemetry API subscription's `maxBytes`, AWS cuts it mid-value and sends the remainder in the next POST, which repeats the cut record's envelope ahead of the resumed bytes. Neither half is valid JSON on its own, so `serde_json` failed and the listener discarded the entire batch — sometimes including the `platform.runtimeDone` that the on-demand loop waits on before calling `/next`. With that event gone the extension held the invocation open until Lambda killed the sandbox.

## Changes
- **`extension/telemetry/stitch.rs` (new)** — buffers the leading fragment and joins the continuation onto it, pairing the two on the repeated envelope (the API gives no sequence number). Joining strips the head's framing and the continuation's repeated envelope.
  - Only payloads cut mid-value are held (`serde_json`'s `Eof` category); anything else
    takes the existing drop path. One fragment at a time, 4 MiB cap (a full POST can be `2 * maxBytes + metadataBytes`), 1s TTL.
- **`maxBytes` 256 KiB → 1 MiB** (the AWS maximum) — prevention, not the fix. The ~328 KB
  record now arrives in a single POST; stitching covers anything above 1 MiB.
- **Listener body limit 2 MiB → 4 MiB** — axum's default would reject a full-size POST before the handler sees it, since the API can send up to 2 * maxBytes + metadataBytes.

## Testing 
Manually tested with a repro function emitting a ~330 KB log record, left at the original 256 KiB `maxBytes` so every invocation splits: every batch reassembled, 0 parse failures, 0 dropped fragments, invocations 65–102 ms against ~62,000 ms beforehand. This fixes the problem at the root, and does not depend on the buffer increase.

Saw debug logs confirming that events were recovered from re-joining the split payload:
```
{
    "status": "DEBUG",
    "message": "DD_EXTENSION | DEBUG | TELEMETRY API | Reassembled a split payload, recovered 4 events"
}
```

### DIFF
--- a/bottlecap/src/extension/telemetry/stitch.rs
+++ b/bottlecap/src/extension/telemetry/stitch.rs
@@ -41,9 +41,6 @@ const FRAGMENT_TTL: Duration = Duration::from_secs(1);
 /// Precedes a record's value, so everything up to and including it is the envelope.
 const RECORD_KEY: &[u8] = b"\"record\":";
 
-/// Opens a record.
-const RECORD_START: &[u8] = b"{\"time\":";
-
 /// The API writes the record's closing `}` and the array's `]` even after cutting the
 /// record's value short, so a fragment ends with framing that belongs to neither half.
 const FRAMING: &[u8] = b"}]";
@@ -136,8 +133,7 @@ impl Fragment {
 
         // Rebuild the cut record's envelope as the continuation will send it: `[` then the
         // record's keys, up to the value that got cut. The cut record is the last one here.
-        let value_start = rfind(&body, RECORD_KEY)? + RECORD_KEY.len();
-        let record_start = rfind(body.get(..value_start)?, RECORD_START)?;
+        let (record_start, value_start) = envelope_bounds(&body)?;
 
         let mut repeated_envelope = vec![b'['];
         repeated_envelope.extend_from_slice(body.get(record_start..value_start)?);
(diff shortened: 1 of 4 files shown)


bottlecap/src/extension/telemetry/stitch.rs
```diff
@@ -0,0 +1,287 @@
...
+        if error.classify() != Category::Eof
+            || body.len() > MAX_FRAGMENT_BYTES
+            || body.first() != Some(&b'[')
+        {
+            return None;
+        }
+
+        // Rebuild the cut record's envelope as the continuation will send it: `[` then the
+        // record's keys, up to the value that got cut. The cut record is the last one here.
+        let value_start = rfind(&body, RECORD_KEY)? + RECORD_KEY.len();
```
litianningdatadog: Not sure whether it is over concern, but this still locates the envelope with `rfind("\"record\":")`. Structured function logs can contain nested `record` fields. If the split occurs after one, the stitcher builds the wrong repeated envelope, such as "[{"record":{"message":{"record":"XYZ}]"
How about replacing the raw `rfind` logic with a bounded JSON-aware scanner.

litianningdatadog (commented):
Commented.

bottlecap/src/extension/telemetry/stitch.rs
shreyamalpani: Real bug — rfind searches the head backwards, which starts inside customer data. Fixed by reading the envelope off the continuation instead, where it sits ahead of any customer bytes (time is a timestamp, type is a fixed enum), so a forward search always lands on the right one and no depth tracking is needed.

litianningdatadog:
Discovered by AI: looks like `next_event` invocation incorrectly twisted the position of runtime_id and extension_id in [this line](https://github.com/DataDog/datadog-lambda-extension/blob/f5171728e4b8f5097e629c49f57eefbe58892c87/bottlecap/src/bin/bottlecap/main.rs#L292). Since this PR starts return Err on the rejected subscriptions, that error bubbles out of `extension_loop_active`, and the caller explicitly falls back to `extension_loop_idle`. I think we should fix that in this PR.

litianningdatadog (approved):

PR #1367, opened 2026-09-10
fix: make race-safe layer publishing opt-in for the e2e sandbox job

lucaspimentel:
## Overview

Automatic layer publishing (`AUTOMATICALLY_BUMP_VERSION=1`) predicted the next layer version by listing the current latest and adding 1, then ran a strict publish loop that fails the job when AWS assigns a higher version than predicted. Because layer versions are assigned atomically by AWS at publish time, a concurrent pipeline publishing to the same layer between the list call and the publish call makes the prediction stale, and the job fails with `Published version N is greater than the desired version N-1` even though the published version holds this job's own binary.

This PR adds an explicit `USE_AWS_ASSIGNED_VERSION` flag that publishes once and adopts the version AWS returns, and enables it **only** for the `publish layer e2e sandbox` job, which is the job that blocks serverless e2e tests when it fails. All other auto-publishing callers (sandbox, serverless_testing, self-monitoring, govcloud us1-staging-fed) keep the previous latest-plus-one prediction and strict behavior. Fixed-version publishing (prod tags, explicit `VERSION`) is unchanged and remains strict.

For the e2e job, the exported `EXTENSION_LAYER_ARN` dotenv artifact now always refers to the version this job actually published, so the downstream e2e-test trigger jobs test the binary from this pipeline.

## Testing

- `bash -n .gitlab/scripts/publish_layers.sh` passes.
- Regression tests with a stubbed `aws` CLI covering five scenarios, all passing:
  - auto mode + `USE_AWS_ASSIGNED_VERSION=1` (latest 698, AWS assigns 700): exits 0, one `publish-layer-version` call, no `list-layer-versions` calls, dotenv exports the ARN of the AWS-assigned version.
  - auto mode without the flag, AWS assigns above the prediction: exits 1 with `greater than the desired version`, no dotenv (previous behavior preserved).
  - auto mode without the flag, happy path (latest 698, AWS assigns 699): exits 0, exports `:699` ARN.
  - fixed mode `VERSION=699`, AWS assigns 700: still fails, no dotenv.
  - fixed mode `VERSION=699`, AWS assigns 699: succeeds, exports `:699` ARN.
- `shellcheck` findings limited to the pre-existing SC2086/SC2059/SC2016/SC2004 classes already present on `main`.


lucaspimentel:
@codex review

lucaspimentel:
@datadog review

lucaspimentel:
@datadog review

